PR #51, opened 2014-07-03
CB-7067 run jasmine tests individually

jsoref:
CB-7067 run jasmine tests individually

`superspawn.spawn` `opts` now take `onstdio` and `onstderr` to
incrementally handle output from child processes.

`npm run jasmine` now runs `jasmine-serial` instead of `jasmine-node`

`jasmine-serial` runs each test that `jasmine-node` would have run
in the same order `jasmine-node` would have run them using
`jasmine-node` with any command-line arguments you pass it (that
`jasmine-serial` supports).

`serve.spec` now skips tests for platforms for which it's missing
the required binaries.

### DIFF
--- a/cordova-lib/jasmine-isolated.js
+++ b/cordova-lib/jasmine-isolated.js
@@ -8,7 +8,8 @@ var path = require('path')
   , superspawn = require('./src/cordova/superspawn')
   , jasmine = require('jasmine-node')
   , specs = require('jasmine-node/lib/jasmine-node/spec-collection')
-  , d = Q.defer()
+  , scheduled = Q.defer()
+  , jobs = [scheduled.promise]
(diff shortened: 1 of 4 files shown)


jsoref:
@agrieve: I'm open to suggestions. I don't know what to do with this. I'm happy to push it in, or have someone else figure out how to host it somewhere.

kamrik:
At the time I thought the problem of interfering tests would be biting us much more often but it didn't manifest itself in a while. So I don't think the extra code and complexity of this change is worth it as part of cordova. It would be nice, though, to see it as something external.

stevengill:
I am going to close this. We have stepped away from jasmine-node. Now using jasmine ^2.x

PR #127, opened 2014-12-03
indented scriptContent.

roadmanfong:
Indented plugin JS files.


agrieve:
Ah, okay. I think if you're doing this, then you need to manually delete the function(){} wrapper that cordova adds to the file, no? In which case you wouldn't want it indented anymore?

purplecabbage:
Yeah, I agree with Andrew. -1 on this from me.
It is okay if the top level is missing one level of indentation after being wrapped by plugman install.

PR #140, opened 2015-01-03
CB-8226 'cordova platform add' : Look up version in config.xml if no version specified

omefire:
CB-8226 
'cordova platform add' : Look up version in config.xml if no version specified

### DIFF
--- a/cordova-lib/src/cordova/platform.js
+++ b/cordova-lib/src/cordova/platform.js
@@ -50,7 +50,7 @@ function add(hooksRunner, projectRoot, targets, opts) {
     var msg;
     if ( !targets || !targets.length ) {
         msg = 'No platform specified. Please specify a platform to add. ' +
-            'See `' + cordova_util.binname + ' platform list`.';
+              'See `' + cordova_util.binname + ' platform list`.';
         return Q.reject(new CordovaError(msg));
     }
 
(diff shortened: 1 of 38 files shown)


omefire:
These failures seem unrelated to my changes. Ran the tests again on an Ubuntu box and all tests pass. I think these failures are intermittent and sometimes they show up, sometimes they don't.

omefire:
Hey @agrieve, I've just addressed your review comments. could you please take a look again ? Thanks.

cordova-lib/src/cordova/platform.js
```diff
@@ -78,62 +76,119 @@ function add(hooksRunner, projectRoot, targets, opts) {
...
+	            if (!(platform in platforms)) {
+	                return getPlatformDetailsFromDir(target);
+	            }
+	            else {
+	                if (!version) {
+	                    events.emit('verbose', 'No version supplied. Retrieving version from config.xml...');
+	                }
+	                version = version || getVersionFromConfigFile(platform, cfg);
+	                if (version) {
+	                    version = version.replace(/^file:\/\//, ''); // If version is a folder, remove the 'file://' prefix from it.
```
agrieve: Couldn't figure out how to reply to old comment, so creating a new one again :S.

If it starts with "file:", then it's a URL (with file: being the scheme). If it has any `:`s, or spaces, then it's a malformed URL that most parses allow. However, we really should support properly formatted URLs by using the url.parse method.

agrieve:
Looks good except for the URL thing :)

omefire:
Using url.parse introduces additional complexity. Especially, Windows paths are not going to be handled properly: the drive letter gets omitted after parsing. This leads to more confusion for the user.
Linux/Unix absolute paths would be handled appropriately, but Windows's wouldn't.

I suggest we remove the file:// prefix and not make config.xml contain urls for folder paths.
Take the following examples:
1-  url.parse('file://D:/Cordova/cordova-lib') => 
{ 
  protocol: 'file:',
  slashes: true,
  auth: null,
  host: 'd',
  port: null,
  hostname: 'd',
  hash: null,
  search: null,
  query: null,
  pathname: '/Cordova/cordova-lib',
  path: '/Cordova/cordova-lib',
  href: 'file://d/Cordova/cordova-lib' 
}

2- url.parse('file://D:\Cordova\cordova-lib') => 
{ 
  protocol: 'file:',
  slashes: true,
  auth: null,
  host: 'd',
  port: null,
  hostname: 'd',
  hash: null,
  search: null,
  query: null,
  pathname: '/:Cordovacordova-lib',
  path: '/:Cordovacordova-lib',
  href: 'file://d/:Cordovacordova-lib' 
}

3- url.parse('file:///Cordova/cordova-lib') =>
{ 
  protocol: 'file:',
  slashes: true,
  auth: null,
  host: '',
  port: null,
  hostname: '',
  hash: null,
  search: null,
  query: null,
  pathname: '/Cordova/cordova-lib',
  path: '/Cordova/cordova-lib',
  href: 'file:///Cordova/cordova-lib' 
}

So, my suggestion is to only support plain filesystem paths for folders in config.xml.
What do you think ?

agrieve:
I agree that's some unnecessary complexity. Let's just drop supporting of file: prefix.

omefire:
Done! 'file://' prefix support dropped. :)

agrieve:
Looks good I think. Just tried to pull it in and there are many merge conflicts unfortunately. Would you be able to rebase?

omefire:
Working on it ...

omefire:
travis-ci failure not related to my changes.

omefire:
@agrieve, have you had a chance to take another look at this ?   --Thanks.

agrieve:
Merged! However - after a closer look at the unit tests, I highly encourage you to try and refactor things so that the tests do not mock out private methods within modules (e.g. like `resolvePath()`). Tests that do that make it very hard to change the code going forward, as it tests implementation details rather than function pre/post-conditions.

omefire:
@agrieve , I will keep that in mind. Thanks for taking the time to review this !

PR #143, opened 2015-01-08
Auto save restore for plugins

gorkem:
uses .cordova/config.json to automatically save or restore plugins to/from config.xml on a project

### DIFF
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -32,12 +32,13 @@ var cordova_util  = require('./util'),
     shell         = require('shelljs'),
     PluginInfo    = require('../PluginInfo'),
     plugman       = require('../plugman/plugman'),
-    events        = require('../events');
+    events        = require('../events'),
+    save          = require('./save');
 
 // Returns a promise.
 module.exports = function plugin(command, targets, opts) {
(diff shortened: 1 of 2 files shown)


cordova-lib/src/cordova/plugin.js
```diff
@@ -206,7 +207,7 @@ module.exports = function plugin(command, targets, opts) {
                         });
                     }, Q())
                     .then(function() {
-                        return plugman.raw.uninstall.uninstallPlugin(target, path.join(projectRoot, 'plugins'));
+                        return plugman.raw.uninstall.uninstallPlugin(target, path.join(projectRoot, 'plugins')).then(savePluginsToConfig());
```
agrieve: I think here and below you actually want: `.then(savePluginsToConfig)` rather than `.then(savePluginsToConfig())`. Often with promises, it's best to just always use `.then(function() {})`, even if all it does is delegate to a helper.

gorkem:
@omefire Correct did that because plugman has its own hooks, I think, it may be useful to keep the config.xml updated for them

gorkem:
@agrieve Good catch. It is dumb luck this was working correctly. Updated the PR

gorkem:
closing as this is obsolete

PR #145, opened 2015-01-12
CB-8168 --list support for cordova-lib

muratsu:
Add list support back after the discussion on the dev mail list

### DIFF
--- a/cordova-lib/src/cordova/targets.js
+++ b/cordova-lib/src/cordova/targets.js
@@ -1,0 +1,68 @@
+/**
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


cordova-lib/src/cordova/targets.js
```diff
@@ -0,0 +1,65 @@
...
+*/
+
+var cordova_util = require('./util'),
+    Q = require('q'),
+    superspawn = require('./superspawn'),
+    path = require('path');
+
+function handleError(error) {
+    if (error.code === 'ENOENT') {
+        console.log('Platform not supported');
```
agrieve: You should use the events.emit function rather than console.log'ing things.

agrieve:
LGTM save the one comment!

cordova-lib/src/cordova/targets.js
muratsu: fixed

PR #147, opened 2015-01-13
Generate apk name

fabiomontefuscolo:
After update the phonegap for my app, the build always generate an APK file named CordovaApp.apk. Is this a good way to solve this?


agrieve:
Certainly is annoying, but I think changing the name of the activity will cause it to not find the associated `.java` file.

PR #148, opened 2015-01-13
CB-8239 Add support for git urls to 'cordova platform add'

omefire:
CB-8239 Add support for git urls to 'cordova platform add'

These changes allow the following scenarios:
     *users can issue 'cordova platform add https://github.com/apache/cordova-android.git' and the git repo will be cloned and used.
     *users can issue 'cordova platform add android@https://github.com/apache/cordova-android.git' and the git repo will be cloned and used.
     *users can issue 'cordova platform add android' and if their config.xml file contains: '<engine id='android'      version='https://github.com/apache/cordova-android.git' />, then the git repo pointed to by config.xml will be cloned and used.

### DIFF
--- a/cordova-lib/spec-cordova/lazy_load.spec.js
+++ b/cordova-lib/spec-cordova/lazy_load.spec.js
@@ -26,7 +26,8 @@ var lazy_load = require('../src/cordova/lazy_load'),
     request = require('request'),
     fs = require('fs'),
     Q = require('q'),
-    platforms = require('../src/cordova/platforms');
+    platforms = require('../src/cordova/platforms'),
+    child_process = require('child_process');
 
 describe('lazy_load module', function() {
     var custom_path;
@@ -91,7 +92,7 @@ describe('lazy_load module', function() {
                 }
             };
             lazy_load.custom(mock_platforms, 'platform X').then(function() {
-                expect(fire).not.toHaveBeenCalled()
+                expect(fire).not.toHaveBeenCalled();
             }, function(err) {
                 expect(err).not.toBeDefined();
             }).fin(done);
@@ -106,7 +107,7 @@ describe('lazy_load module', function() {
                 }
             };
             lazy_load.custom(mock_platforms, 'platform X').then(function() {
-                expect(fire).not.toHaveBeenCalled()
+                expect(fire).not.toHaveBeenCalled();
             }, function(err) {
                 expect(err).not.toBeDefined();
             }).fin(done);
@@ -136,7 +137,7 @@ describe('lazy_load module', function() {
                     }, 10);
                     return fakeRequest;
                 });
-                load_spy = spyOn(npmconf, 'load').andCallFake(function(cb) { cb(null, { get: function() { return npmConfProxy }}); });
+                load_spy = spyOn(npmconf, 'load').andCallFake(function(cb) { cb(null, { get: function() { return npmConfProxy; }}); });
             });
 
             it('should call request with appropriate url params', function(done) {
@@ -268,5 +269,60 @@ describe('lazy_load module', function() {
                 expect(err).toBeUndefined();
             }).fin(done);
         });
+    });
+});
+
+describe('git_clone method', function() {
+    var cordova, custom, tmp_dir, fire;
+    beforeEach(function() {
+        cordova = spyOn(lazy_load, 'cordova').andReturn(Q());
+        custom = spyOn(lazy_load, 'custom').andReturn(Q());
+        fire = spyOn(HooksRunner, 'fire').andCallFake(function(method,args){
+            tmp_dir = args.location;
+            fs.writeFileSync(args.location + '/package.json', '{"name":"cordova-ios"}');
+            return Q();
+        });
+        spyOn(child_process, 'exec').andCallFake(function(cmd, config, callback){
+            callback();
+        });
+    });
+    afterEach(function() {
+        shell.rm('-rf', tmp_dir);
+    });
+    it('should call before_library_clone hook', function(done){
+        lazy_load.git_clone('some repo').then(function() {
+            expect(fire).toHaveBeenCalledWith('before_library_clone', {repository:'some repo',location:tmp_dir});
+        }, function(err) {
+            expect(err).toBeUndefined();
+        }).fin(done);
+    });
+    it('should clone git repo', function(done){
+        lazy_load.git_clone('some repo').then(function() {
+            expect(child_process.exec).toHaveBeenCalledWith('git clone "some repo" "' + tmp_dir + '"', {}, jasmine.any(Function));
+        }, function(err) {
+            expect(err).toBeUndefined();
+        }).fin(done);
+    });
+    it('should checkout ref', function(done){
+        lazy_load.git_clone('some repo').then(function() {
+            expect(child_process.exec).toHaveBeenCalledWith('git checkout "master"', {cwd:tmp_dir}, jasmine.any(Function));
+        }, function(err) {
+            expect(err).toBeUndefined();
+        }).fin(done);
+    });
+    it('should call after_library_clone hook', function(done){
+        lazy_load.git_clone('some repo').then(function() {
+            expect(fire).toHaveBeenCalledWith('after_library_clone', {repository:'some repo',location:tmp_dir,platform:'ios'});
+        }, function(err) {
+            expect(err).toBeUndefined();
+        }).fin(done);
+    });
+    it('should return clone with platform and libDir', function(done){
+        lazy_load.git_clone('some repo').then(function(clone) {
+            expect(clone.platform).toBe('ios');
+            expect(clone.libDir).toBe(tmp_dir);
+        }, function(err) {
+            expect(err).toBeUndefined();
+        }).fin(done);
     });
 });
--- a/cordova-lib/spec-cordova/platform.spec.js
+++ b/cordova-lib/spec-cordova/platform.spec.js
@@ -28,7 +28,9 @@ var helpers = require('./helpers'),
     events = require('../src/events'),
     cordova = require('../src/cordova/cordova'),
     rewire = require('rewire'),
-    platform = rewire('../src/cordova/platform.js');
+    util = rewire('../src/cordova/util.js'),
+    platform = rewire('../src/cordova/platform.js'),
+    CordovaError = require('../src/CordovaError');
 
 var projectRoot = 'C:\\Projects\\cordova-projects\\move-tracker';
 
@@ -338,8 +340,11 @@ describe('add function', function () {
             platform: 'android',
             libDir: targets[0]
         }));
-        platform.__set__('getPlatformDetailsFromDir', getPlatformDetailsFromDirMock);
-
+	
+	platform.__set__('util', {
+	    getPlatformDetailsFromDir: getPlatformDetailsFromDirMock
+	});
+	
         var downloadPlatformMock = jasmine.createSpy();
         platform.__set__('downloadPlatform', downloadPlatformMock);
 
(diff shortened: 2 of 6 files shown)


jsoref:
please ensure there's a PR to update:
https://github.com/apache/cordova-cli/blob/38dd67242841dd2ec254f7f9f402f495f7a28f5b/doc/platform.txt#L20

cordova-lib/spec-cordova/platform.spec.js
```diff
@@ -354,22 +359,94 @@ describe('add function', function () {
...
-        var targets = ['C:\\Projects\\cordova-projects\\cordova-android'];
+        var targets = ['https://github.com/apache/cordova-android'],
+	    targetPlatform = 'android',
+	    libDir = 'C:\\Projects\\cordova-projects\\cordova-android-clone-directory';
+
+	var cloneGitRepoMock = jasmine.createSpy().andReturn(Q({
+	    platform: targetPlatform,
+	    libDir: libDir
+	}));
+	platform.__set__('cloneGitRepo', cloneGitRepoMock);
```
jsoref: you're inserting some tabs.

in general, tabs are frowned upon.

cordova-lib/src/cordova/platform.js
```diff
@@ -76,97 +77,94 @@ function add(hooksRunner, projectRoot, targets, opts) {
...
-            }).then(function(platDetails) {
-                var template = config_json && config_json.lib && config_json.lib[platform] && config_json.lib[platform].template || null;
-                return call_into_create(platDetails.platform, projectRoot, cfg, platDetails.libDir, template, opts);
-            });
-        });
-    })
-    .then(function() {
-        return hooksRunner.fire('after_platform_add', opts);
-    });
+	.then(function () {
```
jsoref: please don't insert tabs. and please try to avoid rewriting large chunks of files when adding features.

cordova-lib/spec-cordova/platform.spec.js
omefire: @jsoref , why do I still need a PR to update https://github.com/apache/cordova-cli/blob/38dd67242841dd2ec254f7f9f402f495f7a28f5b/doc/platform.txt#L20 ?
It seems like the file already refers to <url>s

omefire:
Task for docs update : https://issues.apache.org/jira/browse/CB-8308
I'll probably send another PR for that by the end of the week.

cordova-lib/src/cordova/lazy_load.js
```diff
@@ -284,4 +286,47 @@ function custom(platforms, platform) {
...
 
+// Returns a promise
+function git_clone(repository) {
+    // Create a tmp dir. Using /tmp is a problem because it's often on a different partition and sehll.mv()
+    // fails in this case with "EXDEV, cross-device link not permitted".
+    var tmp_subidr = 'tmp_cordova_git_' + process.pid + '_' + (new Date()).valueOf();
+    var tmp_dir = path.join(util.libDirectory, 'tmp', tmp_subidr);
+    shell.rm('-rf', tmp_dir);
+    shell.mkdir('-p', tmp_dir);
+    if (!shell.which('git')) {
```
agrieve: might as well do this check before mkdir. that way you don't need to rmdir.

cordova-lib/src/cordova/lazy_load.js
```diff
@@ -284,4 +286,47 @@ function custom(platforms, platform) {
     });
 }
 
+// Returns a promise
+function git_clone(repository) {
```
agrieve: cordova-lib already has code to checkout from git:
https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/plugman/util/plugins.js#L49

Would be better to make it so that both paths use the same helper for checking out a git repo.

cordova-lib/spec-cordova/util.spec.js
```diff
@@ -183,4 +184,138 @@ describe('util module', function() {
             expect(res.indexOf('CVS')).toEqual(-1);
         });
     });
+    describe('getPlatformDetailsFromDir function', function () {
+
+        var dir = 'C:\\Projects\\cordova-projects\\cordova-android';
```
agrieve: I hate to be too harsh, but I really don't think these tests add any value (suspect they are just maintenance burden).

The signs to me are:
1. There is more code here to mock out things than there is to actually test.
2. Windows-style paths, and they work perfectly fine on OS X. They really shouldn't but you're mocking out so much that nothing is really being exercised in a way that resembles reality.
3. Mocking out of private functions 

In an ideal world, tests will still work when implementations change so long as APIs remain the same. I this is too far from the truth, then the presence of the tests will hurt the project more than help it, because it will hurt the ability to make changes to the code.

So.. I'd be fine with this landing without new tests, but I don't think I'd land it with these ones.

If you'd like to have another stab at it, I think the biggest problem is that the code itself is difficult to test. This usually means a refactor will improve both the testability and quality of the code (generally, testable code is well-factored code). Likely introducing dependency injection will help as well (e.g. create a "Downloader" class). We can't change the platform.js API, but it's perfectly fine to create a new API, test that, and then write some glue to make the old API call the new one.

If you're not up for refactoring / rewriting the tests, totally understandable.

cordova-lib/spec-cordova/util.spec.js
```diff
@@ -183,4 +184,138 @@ describe('util module', function() {
...
+            getPlatformDetailsFromDir(dir).then(function (platformDetails) {
+                expect(platformDetails.platform).toBe('android');
+                expect(platformDetails.libDir).toBe(dir);
+                done();
+            });
+        });
+
+        it('throws if the directory supplied does not contain a package.json file', function (done) {
+
+            util.__set__('getPackageJsonContent', function (p) {
```
agrieve: Better to use Jasmine's spyOn(util, 'getPackageJsonContent'), since jasmine will take care of restoring the original function after each spec

cordova-lib/src/cordova/lazy_load.js
omefire: refactored plugins.js and lazy_load.js to use the same code for git cloning and checkout : https://github.com/MSOpenTech/cordova-lib/commit/f7ce4fb55aa55168023bbc62592c6323c81e6b8c

cordova-lib/spec-cordova/util.spec.js
omefire: Unit Tests removed: https://github.com/MSOpenTech/cordova-lib/commit/f7ce4fb55aa55168023bbc62592c6323c81e6b8c

cordova-lib/src/cordova/util.js
```diff
@@ -55,6 +61,8 @@ exports.preProcessOptions = preProcessOptions;
 exports.addModuleProperty = addModuleProperty;
 exports.getOrigWorkingDirectory = getOrigWorkingDirectory;
 exports.fixRelativePath = fixRelativePath;
+exports.getPlatformDetailsFromDir = getPlatformDetailsFromDir;
```
agrieve: "util" is generally a place where code factoring is low because people dump into it. I think this case applies - you should create a new module for the git cloning. Don't care where you put it though. Maybe call it "gitclone.js"

cordova-lib/spec-cordova/lazy_load.spec.js
```diff
@@ -26,7 +26,8 @@ var lazy_load = require('../src/cordova/lazy_load'),
     request = require('request'),
     fs = require('fs'),
     Q = require('q'),
-    platforms = require('../src/cordova/platforms');
+    platforms = require('../src/cordova/platforms'),
```
agrieve: The first several screens of this diff are no-op changes. Please have a look through the diff yourself and undo any highlighted lines that maybe your editor is changing on you? Might be easier to start fresh and re-apply your change rather than try and fix up this one.

cordova-lib/src/cordova/platform.js
```diff
@@ -71,102 +72,98 @@ function add(hooksRunner, projectRoot, targets, opts) {
...
-            return Q.when().then(function() {
+            return Q.when().then(function () {
                 if (!(platform in platforms)) {
-                    return getPlatformDetailsFromDir(target);
-                } else {
+                    // First, try handling 'platform' as a directory, if it fails, try handling it as a git repository
+                    return Q.fcall(function () {
+                        return util.getPlatformDetailsFromDir(target);
+                    }).fail(function (err) {
+                        // Maybe it's a git repo
```
agrieve: Will this cause mistyped directory names to give an error message about git cloning not working? Maybe first at least check if it is a valid URL before trying to clone it?

cordova-lib/src/cordova/platform.js
```diff
@@ -71,102 +72,98 @@ function add(hooksRunner, projectRoot, targets, opts) {
...
-    return path.resolve(pPath);
+function cloneGitRepo(gitRepo)
+{
+    // Check all exit points and make sure their return types match to expected one.
+    return lazy_load.git_clone(gitRepo)
+        .then(function(clone) {
+	    // After cloning, retrieve the platform details from dir
+	    return util.getPlatformDetailsFromDir(clone.libDir);
+        })
+        .then(function(platDetails){
```
agrieve: this block is a no-op

cordova-lib/src/cordova/platform.js
```diff
@@ -423,7 +420,7 @@ function platform(command, targets, opts) {
             // Neither path, nor platform name - throw.
             var msg;
             if (/[~:/\\.]/.test(t)) {
-                msg = 'Platform path "' + t + '" not found.';
+                return;
```
agrieve: This needs a code comment saying what's going on in this case.

cordova-lib/src/cordova/util.js
```diff
@@ -262,3 +270,81 @@ function addModuleProperty(module, symbol, modulePath, opt_wrap, opt_obj) {
         });
     }
 }
+
+function resolvePath(pPath) {
```
agrieve: Extracting this as a helper just makes the code harder to follow.

cordova-lib/src/cordova/platform.js
omefire: No. The error message would be: 'Unable to add platform android. Make sure to provide a valid version, an existing folder or an accessible git repository: '. so, we don't need the check.

cordova-lib/src/cordova/util.js
omefire: Fixed in: https://github.com/MSOpenTech/cordova-lib/commit/9d267df1b389ef1f70dcbca4292e1ffd6641e895

cordova-lib/src/cordova/util.js
omefire: moved cloning code to new module: https://github.com/MSOpenTech/cordova-lib/commit/9d267df1b389ef1f70dcbca4292e1ffd6641e895

cordova-lib/spec-cordova/lazy_load.spec.js
omefire: It seems the no-op changes you're referring to are fixes to make jshint stop complaining.

omefire:
Tests are passing locally, so is jshint. I don't know why travis isn't picking up my changes.

purplecabbage:
it's a merge issue.

agrieve:
Merged this in after cleaning up merge / rebase issues (and some minor fixups from some testing)

omefire:
Hey @agrieve, It seems like the refactoring you did before merging this in introduced a bug : 

Suppose the latest android version is: 3.7.0 and my project contains android version 3.5.0.
When I run 'cordova platform update android' with my config.xml containing: <cdv:engine id="android" version="3.6.0" />, my project's android platform is going to be upgraded to 3.6.0 instead of being upgraded to 3.7.0.
What happens is that it grabs the version from the config.xml if any and only resorts to the latest pinned version (3.7.0) if the config.xml file is empty.

omefire:
Here's a pull request for the fix of the afore-mentionned bug : https://github.com/apache/cordova-lib/pull/159

PR #153, opened 2015-01-23
CB-6973 Improving JSHint for src folder

muratsu:
Enforcing better JSHint coverage for src\* :
- Move base rules to src\.jshintrc
- Fix previously un-jshinted files
- Remove base rules from individual files to contain exceptions only


kamrik:
LGTM, merging.

PR #154, opened 2015-01-24
CB-6973 enable JSHint for spec-cordova

muratsu:
This change enables jshint for spec-cordova meaning that `npm run jshint` will not only check `src` folder but `spec-cordova` too. I've also fixed all the JSHint issues for the spec-cordova.


kamrik:
Looks good.
There is a conflict now with the last commit in spec-cordova/platform.spec.js
Could you please update the PR, thanks.

muratsu:
Should be good to to go once tests are all green

PR #155, opened 2015-01-28
CB-8123 Plugin references can target specific windows platforms.

TimBarham:
Adds support for `target`, `versions` and `arch` attributes on `<lib-file>` and `<framework>` elements in the windows platform of plugin.xml. This allows plugin authors to target different references to different target platforms.

Also adds support for `src` attribute as an alias for the `Include` attribute on the `<lib-file>` element (since `src` is documented, but `Include` is used by existing plugins).

Adds some tests to cover the new attributes. Updates existing plugin tests for windows8 platform to also test windows platform (left in windows8 tests to help verify backward compatibility with old windows8 platform).

As part of this change, refactored `jsproj` to `jsprojManager` to reflect the fact that, with the windows platform, this class now manages multiple jsproj files.


purplecabbage:
Very cool Tim!

cordova-lib/spec-plugman/platforms/windows.spec.js
```diff
@@ -0,0 +1,415 @@
...
+                        windows['lib-file'].install(invalidLibFiles[1], faultyplugin, cordovaProjectWindowsPlatformDir, faulty_id, null, proj_files);
+                    }).toThrow('Invalid lib-file versions attribute (must be a valid a valid node semantic version range): 8.0a');
+
+                    expect(function () {
+                        windows['lib-file'].install(invalidLibFiles[2], faultyplugin, cordovaProjectWindowsPlatformDir, faulty_id, null, proj_files);
+                    }).toThrow('Invalid lib-file target attribute (must be "all", "phone", "windows" or "win"): daphne');
+                });
+            });
+
+            describe('of <framework> elements', function () {
```
purplecabbage: First failing test in Travis

cordova-lib/spec-plugman/platforms/windows.spec.js
```diff
@@ -0,0 +1,415 @@
...
+                    xpath = 'Reference[@Include="dummy4"][@Condition="\'$(Platform)\'==\'ARM\'"]/HintPath';
+                    validateInstalledProjects('framework', frameworks[3], xpath, ['phone']);
+                });
+            });
+
+            describe('of <framework> elements of type \'projectReference\'', function () {
+                var frameworks = copyArray(valid_frameworks);
+
+                it('should write to correct project files when conditions are specified', function () {
+                    var xpath = 'ProjectReference[@Include="' + dummyplugin + '\\src\\windows\\dummy1.vcxproj"][@Condition="\'$(Platform)\'==\'x64\'"]';
```
purplecabbage: First failing test in Travis.

purplecabbage:
You should use path.join( ) in your tests when creating paths so the tests will still pass on MacOS, or check the environment and skip running your tests altogether.

TimBarham:
The tests use path.join(), but jsproj uses hard-coded backslashes (which is reasonable, because normally it would only run on Windows). This problem is showing up now because previously this code wasn’t exercised by tests.

I’m looking to see if it is feasible/reasonable to update jsproj to be platform agnostic, to the extent it needs to be for tests to pass. I’d prefer that to skipping the tests, so we get consistent coverage (the tests don’t build projects – just verify that files are modified as expected).

purplecabbage:
yeah, whichever approach is easier. Ultimately the test in this case is invalid, but we don't want it appearing to break the build.

TimBarham:
I'll close this PR, and open a new, cleaned up PR.

PR #161, opened 2015-02-05
[ADD] GET attribute 'version' config.xml Restore Platform

ramonornela:
- Add attribute 'version' restore platform


gorkem:
There is a new implementation of restore cooking, which auto restores engines and also does handle the version. As part of that work restore command will vanish completely. You can see the work in progress at https://github.com/gorkem/cordova-lib/tree/save_restore_final

PR #162, opened 2015-02-09
CB-8420 'cordova plugin add' should retrieve version, url or local folder from config.xml if none is specified

omefire:
CB-8420 'cordova plugin add' should retrieve version, url or local folder from config.xml if none is specified

### DIFF
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -91,6 +91,8 @@ module.exports = function plugin(command, targets, opts) {
                 return Q.reject(new CordovaError('No plugin specified. Please specify a plugin to add. See `'+cordova_util.binname+' plugin search`.'));
             }
 
+            var xml = cordova_util.projectConfig(projectRoot);
+            var cfg = new ConfigParser(xml);
             var config_json = config.read(projectRoot);
             var searchPath = config_json.plugin_search_path || [];
             if (typeof opts.searchpath == 'string') {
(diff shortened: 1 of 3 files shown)


omefire:
@vladimir-kotikov, @agrieve and @gorkem , please review.

gorkem:
-1 since we agreed to have auto restore for plugins I do not think this feature will be used at all.

omefire:
Well, if you review the document we all agreed on, it mentions this. Plus, this feature is already in for platforms as well. I'm trying to mirror it : https://docs.google.com/document/d/1qKjhzSf48ybGg2GFZPtjXP8dkF4Z5Jnc7FU41V3-jFc/edit#heading=h.tsqn5em3043l

omefire:
@gorkem Also, this feature does not prevent the other one you mentionned.

vladimir-kotikov:
I'm not aware of overall design of save/restore features, but code LGTM.

cordova-lib/src/cordova/plugin.js
```diff
@@ -114,6 +116,22 @@ module.exports = function plugin(command, targets, opts) {
...
+                        var parts = target.split('@');
+                        var id = parts[0];
+                        var version = parts[1];
+
+                        // If no version is specified, retrieve the version from config.xml
+                        if(!version && !cordova_util.isUrl(id) && !cordova_util.isDirectory(id)){
+                            events.emit('verbose', 'no version specified, retrieving version from config.xml');
+                            var ver = getVersionFromConfigFile(id, cfg);
+
+                            if( cordova_util.isUrl(ver) || cordova_util.isDirectory(ver) ){
```
agrieve: isDirectory don't check for null/undefined. Should either add that to isDirectory, or add check here.

agrieve:
lgtm save one nit. I think this behaviour is fine. Normal case will be to use add --save anyways.

cordova-lib/src/cordova/plugin.js
omefire: @agrieve, isDirectory does handle the null/undefined cases. I just tested it. Am I missing something ?

PR #163, opened 2015-02-09
CB-8448 Add support for "activities" field in FirefoxOS manifest.webapp

AxelNennker:
See https://issues.apache.org/jira/browse/CB-8448


zalun:
Isn't it already sorted?

AxelNennker:
This PR was replaced by #172

PR #164, opened 2015-02-11
CB-8123 Plugin references can target specific windows platforms.

TimBarham:
Adds support for `target`, `versions` and `arch` attributes on `<lib-file>` and `<framework>` elements in the windows platform of plugin.xml. This allows plugin authors to target different references to different target platforms.

Also adds support for `src` attribute as an alias for the `Include` attribute on the `
<lib-file>` element (since `src` is documented, but `Include` is used by existing plugins).

Adds some tests to cover the new attributes. Updates existing plugin tests for windows8 platform to also test windows platform (left in windows8 tests to help verify backward compatibility with old windows8 platform).

As part of this change, refactored `jsproj` to `jsprojManager` to reflect the fact that, with the windows platform, this class now manages multiple jsproj files.

Broken into three commits to separate file renames from changes, so that git can successfully track the changes.


sgrebnov:
Reviewed and tested, LGTM :+1:

purplecabbage:
Thanks Tim! This is now merged, the mirror is just slow to update. This should close in a few minutes.

TimBarham:
Thanks Jesse!

stefangordon:
Great stuff! Any pointers on how users can easily install this via NPM now? I've only had success manually cloning the lib and cli repos and copying correct binaries.

PR #166, opened 2015-02-12
Save/Restore for plugins and platforms

gorkem:
I think this is a good point to merge these features.. It provides most of the functionality and frankly I am tired of rebasing/merging.

Several features implemented. 
1. Implements auto-restore for platforms and plugins on prepare. Platforms are only restore when a platforms is specifically requested.
2. Implements --save flag for cordova plugins add/rm
3. removes the old save and restore commands
4. removes the old tests for save and restore 

What is missing:
1. Switch to plugin tag on config.xml. ( Another PR is coming for that shortly)
2. tests for --save flag
3. tests for autorestore


omefire:
Cool feature list. reviewing this, :)

omefire:
if no version is specified for (e.g: cordova plugin add org.apache.cordova.device --save), we end up saving the current edge version (org.apache.cordova.device@0.2.13) into config.xml. I think no version specified means always get the 'latest' edge version, not pin (and get) the 'current' edge version.

Also, this behavior is different from how 'cordova platform add android --save' currently behaves.

I think we should not save the current edge version in this case, and upon restoring we restore the edge version instead.

omefire:
I think the 'saving to config.xml' should be moved into a separate function and called from here. this function is getting way too long to easily reason about it.

omefire:
'saving to config.xml' should happen after the plugin has been installed on the platforms. 
That way, if any error is encountered during the install, the config.xml does not get updated.
What do you think ?

omefire:
Hey @gorkem, I notice that --save covers add and remove. Are you going to send a pull request to handle the update case ? : 'cordova plugin update org.apache.cordova.device@3.5.0 --save'

omefire:
Hey @gorkem, I notice that --save covers add and remove. Are you going to send a pull request to handle the update case ? : 'cordova plugin update org.apache.cordova.device@3.5.0 --save'.

[IGNORE THIS COMMENT AS IT IS NOT APPLICABLE, there is no update for plugins]

omefire:
When the flag --save/autosave on, we should not be retrieving the version from config.xml.
For example,
    - config.xml contains org.apache.cordova.device@0.2.9
    - running 'cordova plugin add org.apache.cordova.device --save' results in org.apache.cordova.device@0.2.9 being saved into config.xml instead of org.apache.cordova.device. (with no version).

This means that the check on this line should be updated: https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/cordova/plugin.js#L124

gorkem:
Save logic is a bit disorganized because I was expecting @omefire to implement it for plugins as well, so I have whipped up something. I will probably make another pass with the syntax change for plugins on config.xml. 
Here are some notes
* **no version on save:** The original save command implementation had a flag called -shrinkwrap that was used to indicate that you want to persist the version for the plugin to config.xml, without such a flag users need to know the plugin version beforehand to be able to pin. I am a bit conflicted about what to do without such a flag.

* ** retrieving the version from config.xml:** I have already added a check for this but I am failing to see how this feature is used. If an entry for a plugin already exists on config.xml, it will be restored (provided that this PR is applied) the moment that you try to do something useful with CLI (run,build, prepare, add another plugin) so the only way this feature will be executed is if you clone a project and try to add the plugin immediately before issuing any other commands.

omefire:
Also, we should probably save the plugin variables along with the version into config.xml, so that on prepare/restore, we don't fail.

gorkem:
sure... as long as there plugin install logic uses variables from config.xml, which does not.

omefire:
hmmm, it seems to me that during prepare, while restoring from config.xml, we do use variables from config.xml. so, we should save it during --save/autosave. Please, let me know if I'm missing something :
 
https://github.com/gorkem/cordova-lib/blob/a70ae2d0c3927cbb6bdb1dc214db469ea74086dd/cordova-lib/src/cordova/restore-util.js#L108

omefire:
What if a plugin install fails ? don't we want to continue with other plugins' install like we do for platforms ?
That way, we also handle platform-dependent plugins that are going to fail for certain platforms.

WDYT ?

cordova-lib/src/cordova/plugin.js
```diff
@@ -136,6 +135,35 @@ module.exports = function plugin(command, targets, opts) {
                         events.emit('verbose', 'Calling plugman.fetch on plugin "' + target + '"');
                         return plugman.raw.fetch(target, pluginsDir, { searchpath: searchPath, noregistry: opts.noregistry, link: opts.link, pluginInfoProvider: pluginInfoProvider});
                     })
+                    .then(function(dir){
+                      // save to config.xml
```
agrieve: please stick with a 4-space indent (used everywhere else)

cordova-lib/src/cordova/plugin.js
```diff
@@ -136,6 +135,35 @@ module.exports = function plugin(command, targets, opts) {
                         events.emit('verbose', 'Calling plugman.fetch on plugin "' + target + '"');
                         return plugman.raw.fetch(target, pluginsDir, { searchpath: searchPath, noregistry: opts.noregistry, link: opts.link, pluginInfoProvider: pluginInfoProvider});
                     })
+                    .then(function(dir){
+                      // save to config.xml 
+
+                      if(autosave || opts.save){
+                        var pluginInfo =  pluginInfoProvider.get(dir);
```
agrieve: Good chunk of code here. Would be appropriate for a helper method.

gorkem:
I think I have addressed the concerns with the latest. If everyone is cool with it, we should go ahead and merge them in.

omefire:
Reviewing latest changes ...

omefire:
you forgot to negate opts.save as well : if(!autosave && !opts.save && ...)

omefire:
when --save/autosave is on, we should be overriding what's in config.xml, like we do on the platform side : https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/cordova/platform.js#L161

omefire:
what about the other options ? (opts).
shouldn't we pass them along to 'prepare' ?

danielcompton:
What is the command to run the restore command? I'm using cordova 4.3.0 and I get this message running restore:
```sh
$cordova restore
Cordova does not know restore; try `cordova help` for a list of all the available commands.
```

Looking at https://github.com/apache/cordova-cli/pull/209, the restore docs were removed. I can see code has been added in the Cordova lib repository but it doesn't seem to be working. Can we add documentation for running the restore? I assume it does work but I'm just calling it wrong.

Running `npm ls -g`

```sh
...
├─┬ cordova@4.3.0
│ ├─┬ cordova-lib@4.3.0
│ │ ├── bplist-parser@0.0.6
│ │ ├─┬ cordova-js@3.8.0
│ │ │ ├─┬ browserify@7.1.0
│ │ │ │ ├── assert@1.1.2
...
```

TimBarham:
I believe the decision was made not to implement `restore`, since it happens automatically on `prepare`.

gorkem:
correct. `restore` command is removed and restoration happens during `prepare`. You can call `cordova prepare` explicitly to get the similar results to `restore`.

PR #167, opened 2015-02-16
CB-8472 Can't find config.xml error installing platform after plugin.

TimBarham:
Recent changes to the process of adding a platform meant we were trying to trying to install plugins for the platform before we had called prepare() for the platform. Also, we were calling prepare() outside the loop that processes each platform, which meant the plaform local variable was not in scope, and we were passing a reference to a function called platform() to prepare() rather than a platform name.

Fixed this by moving the prepare() call back where it used to be - just before we install plugins.


TimBarham:
@agrieve could you take a look at this? It impacts on refactoring work you've recently done here.

PR #168, opened 2015-02-17
CB-8482: Update engine syntax within config.xml

omefire:
CB-8482: Update engine syntax within config.xml

- from : `<cdv:engine id="android" version="3.7.0-dev" />`
- to :      `<engine name="android" version="3.7.0-dev" />`
- updated and tested all modules that depended on old syntax

### DIFF
--- a/cordova-lib/src/configparser/ConfigParser.js
+++ b/cordova-lib/src/configparser/ConfigParser.js
@@ -385,24 +385,24 @@ ConfigParser.prototype = {
 
     /**
      * Adds an engine. Does not check for duplicates.
-     * @param  {String} id the engine id
+     * @param  {String} name the engine name
      * @param  {String} version engine version (optional)
      */
-    addEngine: function(id, version){
-        if(!id) return;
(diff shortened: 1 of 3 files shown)


omefire:
CB-8482: Update engine syntax within config.xml

- from : `<cdv:engine id="android" version="3.7.0-dev" />`
- to :      `<engine name="android" version="3.7.0-dev" />`

gorkem:
I wonder if it is worth to do an auto convert from old syntax when we encounter it?

omefire:
There will need to be another pass to update dependent modules to handle the possibility that some new code depending on old syntax might have been inserted in the interim. this is just a reminder to myself and the person who will merge this.

gorkem:
@omefire this looks OK and would like to get it in. The dependencies have changed though save.js and restore.js are gone. I could not quickly spot anything on platform.js other then the current one. Can you do a rebase and double check as well.

omefire:
@gorkem, cool. working on a rebase + update to new dependent modules ...

omefire:
@gorkem, I did find some things that needed to be updated in restore-util.js. those have been taken care of.

PR #170, opened 2015-02-20
CB-8499 `cordova platform save` should save all currently installed platforms and their versions/git-url/folder-location into config.xml.

omefire:
CB-8499 `cordova platform save` should save all currently installed platforms and their versions/git-url/folder-location into config.xml.

### DIFF
--- a/cordova-lib/src/cordova/platform.js
+++ b/cordova-lib/src/cordova/platform.js
@@ -211,6 +211,19 @@ function savePlatformVersion(platformsDir, platform, version) {
     fs.writeFileSync(jsonPath, JSON.stringify(data, null, 4), 'utf-8');
 }
 
+function removePlatformVersion(platformsDir, platform){
+    var jsonPath = path.join(platformsDir, 'platforms.json');
+    if(!fs.existsSync(jsonPath)){
+        return;
+    }
+    var data = getJson(jsonPath);
+
+    // test: what if version is null ? non-null ?
+    delete data[platform]; //test: what if data[platform] is null? non-null?
+    // how does JSON.stringify() work ?
+    fs.writeFileSync(jsonPath, JSON.stringify(data, null, 4), 'utf-8');
+}
+
 function getJson(jsonPath) {  // jsonPath -> jsonFile  
     return JSON.parse(fs.readFileSync(jsonPath, 'utf-8'));
 }
@@ -311,8 +324,14 @@ function remove(hooksRunner, projectRoot, targets, opts) {
 		cfg.write();
 	    });
 	}
-    })
-    .then(function() {
+    }).then(function() {
+        // Remove targets from platforms.json
+        var platformsDir = path.join(projectRoot, 'platforms');
+        targets.forEach(function(target) {
+            events.emit('verbose', 'Removing ' + target + ' from platforms.json file ...');
+            removePlatformVersion(platformsDir, target);
+        });
+    }).then(function() {
         return hooksRunner.fire('after_platform_rm', opts);
     });
 }
--- a/cordova-lib/src/cordova/platform.js
+++ b/cordova-lib/src/cordova/platform.js
@@ -75,7 +75,8 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
 
     // The "platforms" dir is safe to delete, it's almost equivalent to
     // cordova platform rm <list of all platforms>
-    shell.mkdir('-p', path.join(projectRoot, 'platforms'));
+    var platformsDir = path.join(projectRoot, 'platforms');
+    shell.mkdir('-p', platformsDir);
 
     return hooksRunner.fire('before_platform_' + cmd, opts)
     .then(function() {
@@ -162,6 +163,8 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
                         return installPluginsForNewPlatform(platform, projectRoot, cfg, opts);
                     }
                 }).then(function() {
+                    savePlatformVersion(platformsDir, platform, version);
+                }).then(function() {
                     if(opts.save || autosave){
                         // Save target into config.xml, overriding already existing settings
                         events.emit('log', '--save flag or autosave detected');
@@ -176,6 +179,40 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
     }).then(function() {
         return hooksRunner.fire('after_platform_' + cmd, opts);
     });
+}
+
+// save the platform's installed or updated version
+// tests: add, update
+// 
+function savePlatformVersion(platformsDir, platform, version) {
+    debugger;
+    // if no version set, save current edge version
+    if(!version){
+        version = platforms[platform].version;
+    }
+
+    events.emit('verbose', 'saving ' + platform + '@' + version + ' into platforms.json');
+    
+    // test: what if platforms.json already contains this platform and version => override it
+    // test: what if platforms.json is empty ?
+    // test: what if platforms.json hasn't been created yet ? => create it and write into it
+    var jsonPath = path.join(platformsDir, 'platforms.json');
+    if(!fs.existsSync(jsonPath)){
+        // ugly. create in a better way
+        var fd = fs.openSync(jsonPath, 'w'); // test: what if there is an error while creating ? 
+        fs.writeFileSync(jsonPath, JSON.stringify({}, null, 4), 'utf-8');
+        fs.closeSync(fd);
+    }
+    var data = getJson(jsonPath);
+
+    // test: what if version is null ? non-null ?
+    data[platform] = version; //test: what if data[platform] is null? non-null?
+    // how does JSON.stringify() work ?
+    fs.writeFileSync(jsonPath, JSON.stringify(data, null, 4), 'utf-8');
+}
+
+function getJson(jsonPath) {  // jsonPath -> jsonFile  
+    return JSON.parse(fs.readFileSync(jsonPath, 'utf-8'));
 }
 
 // Downloads via npm or via git clone (tries both)
@@ -498,6 +535,8 @@ function platform(command, targets, opts) {
             return update(hooksRunner, projectRoot, targets, opts);
         case 'check':
             return check(hooksRunner, projectRoot);
+        //case 'save':
+            //return save(hooksRunner, projectRoot);
         default:
             return list(hooksRunner, projectRoot);
     }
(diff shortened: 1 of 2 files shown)


omefire:
@agrieve , @vladimir-kotikov  and @gorkem, please review.

cordova-lib/src/cordova/platform.js
```diff
@@ -158,6 +160,11 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
                         return installPluginsForNewPlatform(platform, projectRoot, cfg, opts);
                     }
                 }).then(function() {
+                    // Save platform@version into platforms.json. i.e: 'android@https://github.com/apache/cordova-android.git'
+                    // If no version was specified, save the edge version
+                    var versionToSave = version ? version : platforms[platform].version;
```
vladimir-kotikov: just a nitpick - there is no need for ternary operator, `version || platforms[platform].version` will work the same way

cordova-lib/src/cordova/platform.js
```diff
@@ -174,6 +181,24 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
     });
 }
 
+function save(hooksRunner, projectRoot, opts) {
+    var xml = cordova_util.projectConfig(projectRoot);
+    var cfg = new ConfigParser(xml);
+
+    // First, remove all platforms that are already in config.xml
+    cfg.getEngines().map(function(engine){
+        cfg.removeEngine();
```
vladimir-kotikov: Something strange here. `removeEngine` requires an engine Id param (https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/configparser/ConfigParser.js#L404). Probably should be `cfg.removeEngine(engine.id);` instead.
Also `forEach` IMHO is more appropriate here than `map` here

cordova-lib/src/cordova/platform.js
```diff
@@ -174,6 +181,24 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
...
+    });
+
+    // Save installed platforms into config.xml
+    return platformMetadata.getPlatformVersions(projectRoot).then(function(platformVersions){
+        platformVersions.map(function(platVer){
```
vladimir-kotikov: `map` vs `forEach` again

cordova-lib/src/cordova/platform_metadata.js
```diff
@@ -0,0 +1,113 @@
...
+        if(exists){
+            return getPlatVersionsFromFile(platformsJsonFile);
+        }        
+        return getPlatVersionsFromFileSystem(projectRoot);
+    });
+}
+
+// Returns a promise
+function getPlatVersionsFromFile(platformsJsonFile){
+    var platformData = getJson(platformsJsonFile);
```
vladimir-kotikov: You may want to require JSON file directly: `var platformData = require(platformsJsonFile);`

cordova-lib/src/cordova/platform_metadata.js
```diff
@@ -0,0 +1,113 @@
...
+// Retrieves the platforms and their versions from the platforms.json file
+// Returns an array of {platform: platform, version: version} ...
+// ... where version could be '3.4.0', '/path/to/platform' or 'git://...'
+function getVersions(projectRoot) {
+    var platformsDir = path.join(projectRoot, 'platforms');
+    var platformsJsonFile = path.join(platformsDir, 'platforms.json');
+
+    // If the platforms.json file doesn't exist, retrieve versions installed
+    // In this case, we won't be able to know what source the platform came from
+    return Q.nfcall(fs.exists, platformsJsonFile).then(function(exists){
```
vladimir-kotikov: Due to [nodejs.org/api/fs.html](nodejs.org/api/fs.html#fs_fs_exists_path_callback)
> fs.exists() is an anachronism and exists only for historical reasons. There should almost never be a reason to use it in your own code.

Probably it is better to replace this with something like this:
```
return getPlatVersionsFromFile(platformsJsonFile)
.fail(function () {
    return getPlatVersionsFromFileSystem(projectRoot);
});
```

cordova-lib/src/cordova/platform_metadata.js
```diff
@@ -0,0 +1,113 @@
...
+    });
+
+    return Q(platformVersions);
+}
+
+// Returns a promise
+function getPlatVersionsFromFileSystem(projectRoot){
+    var platformVersions = [];
+    var platforms_on_fs = cordova_util.listPlatforms(projectRoot);
+    platforms_on_fs.map(function(platform){
```
vladimir-kotikov: `map` vs `forEach` again

cordova-lib/src/cordova/platform_metadata.js
omefire: @vladimir-kotikov , getting a 404 on the link

omefire: @vladimir-kotikov , getting a 404 on the link. anyway, thanks for the suggestion, modifying accordingly.

vladimir-kotikov: Sorry, my bad. Here is a proper link http://nodejs.org/api/fs.html#fs_fs_exists_path_callback

cordova-lib/src/cordova/platform.js
omefire: Thanks @vladimir-kotikov , nice catch. :)

cordova-lib/src/cordova/platform.js
omefire: Changed all the map into forEach. it's definitely better to use forEach with side-effecting operations.

omefire:
@vladimir-kotikov Thanks for reviewing, All your reviews have been addressed, please take another look.

vladimir-kotikov:
LGTM

gorkem:
Is platforms.json an existing file? Is it used anywhere else? I suspect that it is a new file and I am not sure if it is really needed

cordova-lib/src/cordova/platform_metadata.js
```diff
@@ -0,0 +1,102 @@
...
+    // If the platforms.json file doesn't exist, retrieve versions from platforms installed on the filesystem...
+    // ...Note that in this case, we won't be able to know what source(folder, git-url) the platform came from, we'll just use versions
+    return getPlatVersionsFromFile(platformsJsonFile).fail(function(){
+        return getPlatVersionsFromFileSystem(projectRoot);
+    });
+}
+
+// Returns a promise
+function getPlatVersionsFromFile(platformsJsonFile){
+    var platformData = require(platformsJsonFile);
```
nikhilkh: Why do a require to a data file? JSON.parse is better as it prevents arbitrary code from being loaded.

omefire:
@gorkem, yes platforms.json is a new file in the platforms directory. think of it as the equivalent of the .fetch.json file we have for plugins.

omefire:
I think this PR is ready to be merged in.

cordova-lib/src/cordova/platform_metadata.js
```diff
@@ -0,0 +1,102 @@
...
+
+// Returns a promise
+function getPlatVersionsFromFileSystem(projectRoot){
+    var platformVersions = [];
+    var platforms_on_fs = cordova_util.listPlatforms(projectRoot);
+    platforms_on_fs.forEach(function(platform){
+        var script = path.join(projectRoot, 'platforms', platform, 'cordova', 'version');
+        
+        // Use version script to retrieve installed version of the platform
+        var version = shelljs.exec(script).output;
```
agrieve: `shelljs.exec` is known to have poor performance and lead to EMFILE exceptions. Please use child_process.exec() instead

cordova-lib/src/cordova/platform_metadata.js
```diff
@@ -0,0 +1,102 @@
...
+    
+    return Q(platformVersions); 
+}
+
+// Saves platform@version into platforms.json
+function save(projectRoot, platform, version) {
+    var platformsDir = path.join(projectRoot, 'platforms');
+    var platformJsonFile = path.join(platformsDir, 'platforms.json');
+    if(!fs.existsSync(platformJsonFile)){
+        fs.writeFileSync(platformJsonFile, JSON.stringify({}, null, 4), 'utf-8');
```
agrieve: unless I'm missing something, you overwrite this file a few lines below. No need to write an empty file here.

cordova-lib/src/cordova/platform_metadata.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    var data = require(platformJsonFile);
```
agrieve: It's generally not very flexible to read and write a file in the same method. Better to have clients explicitly read, explicitly mutate (in memory), then explicitly write.

cordova-lib/src/cordova/platform_metadata.js
vladimir-kotikov: NodeJS actually doing JSON parsing when JSON file is required. See http://nodejs.org/api/modules.html#modules_file_modules and http://nodejs.org/api/modules.html#modules_all_together

agrieve: It is a bit confusing to use require() on non-app code though.

cordova-lib/src/cordova/platform_metadata.js
omefire: In case the 'platform.json' file doesn't exist, we have to create it first.
The alternative is to try and write to the file even when it doesn't exist, then handle the exception. 
I think this way is easier to reason about.

cordova-lib/src/cordova/platform_metadata.js
omefire: Thanks, noted. However, I think in this case, it wouldn't make sense to have this logic in the client `platform.js#addHelper` as it doesn't need to know the details of this operation.

cordova-lib/src/cordova/platform_metadata.js
agrieve: writing to a non-existing file should work fine. do you mean read from it?

You're already checking if it exists, so you might as well:

    var data = {}
    if (exists) {
        data = getJson();
    }

omefire: Thanks @agrieve ! ... slow day today. :)

omefire:
@agrieve , I addressed your comments. could you please take another look ?

omefire:
closing because this Pull Request has been merged in.

PR #171, opened 2015-02-22
CB-7827: Allow user to specify android activity name

cjpearson:
This change adds an attribute to config.xml android-activityName that can be used to specify a custom activity name. If none is specified it will still default to MainActivity. See http://markmail.org/message/pj5ig4b54nk4l7dr for a discussion of this issue.

### DIFF
--- a/cordova-lib/src/cordova/platform.js
+++ b/cordova-lib/src/cordova/platform.js
@@ -521,7 +521,7 @@ function getCreateArgs(platDetails, projectRoot, cfg, template_dir, opts) {
     if (/android|ios/.exec(platDetails.platform) && semver.gt(platDetails.version, '3.3.0')) {
         args.push('--cli');
     }
-    
+
     var pkg = cfg.packageName().replace(/[^\w.]/g,'_');
     // CB-6992 it is necessary to normalize characters
     // because node and shell scripts handles unicode symbols differently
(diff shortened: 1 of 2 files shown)


cordova-lib/src/cordova/platform.js
```diff
@@ -521,19 +521,27 @@ function getCreateArgs(platDetails, projectRoot, cfg, template_dir, opts) {
...
     var pkg = cfg.packageName().replace(/[^\w.]/g,'_');
     // CB-6992 it is necessary to normalize characters
     // because node and shell scripts handles unicode symbols differently
     // We need to normalize the name to NFD form since iOS uses NFD unicode form
     var name = platDetails.platform == 'ios' ? unorm.nfd(cfg.name()) : cfg.name();
     args.push(output, pkg, name);
+
+    var activityName = cfg.android_activityName();
+    if (activityName && platDetails.platform === 'android') {
+        activityName = activityName.replace(/W/g, '');
```
agrieve: why not allow any Ws? did you mean `\W`?

cordova-lib/src/cordova/platform.js
```diff
@@ -521,19 +521,27 @@ function getCreateArgs(platDetails, projectRoot, cfg, template_dir, opts) {
...
+	args.push(activityName);
```
agrieve: We need this to continue to work for older versions of cordova-android. You should add this arg only if:

    semver.gte(platDetails.version, '4.0.0-dev')

cordova-lib/src/cordova/platform.js
cjpearson: Yup, typo there.

cordova-lib/src/cordova/platform.js
cjpearson: Thanks for the review. I'll add this.

cjpearson:
@agrieve I've addressed your comments, could you take another look? Thanks!

fredgalvao:
Shouldn't this be documented somewhere? It does not appear when I run `cordova help create` as a valid option.

faenschi:
Should this work with cordova 4.3.1? Nothing happens if I add android-activityName="MyName" to config.xml

PR #174, opened 2015-02-28
CB-8577 BugFix 'cordova plugin add <plugin> --save' should properly save variables into config.xml to enable restoring later on.

omefire:
CB-8577 BugFix 'cordova plugin add <plugin> --save' should properly save variables into config.xml to enable restoring later on.

### DIFF
--- a/cordova-lib/src/configparser/ConfigParser.js
+++ b/cordova-lib/src/configparser/ConfigParser.js
@@ -368,7 +368,7 @@ ConfigParser.prototype = {
     /**
      *This does not check for duplicate feature entries
      */
-    addFeature: function (name, params){
+    addFeature: function (name, params, variables){
         if(!name) return;
         var el = new et.Element('feature');
         el.attrib.name = name;
@@ -378,6 +378,14 @@ ConfigParser.prototype = {
                 p.attrib.name = param.name;
                 p.attrib.value = param.value;
                 el.append(p);
+            });
+        }
+        if (variables) {
+            variables.forEach(function(variable){
+                var v = new et.Element('variable');
+                v.attrib.name = variable.name;
+                v.attrib.value = variable.value;
+                el.append(v);
             });
         }
         this.doc.getroot().append(el);
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -141,6 +141,7 @@ module.exports = function plugin(command, targets, opts) {
                             var pluginInfo =  pluginInfoProvider.get(dir);
                             var existingFeature = cfg.getFeature(pluginInfo.id);
                             if(!existingFeature){
+                                var variables = [];
                                 var params = [{name:'id', value:pluginInfo.id}];
                                 var pluginVersion = versionFromTargetString(target);
                                 if(!pluginVersion && opts.shrinkwrap){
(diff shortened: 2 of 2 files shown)


gorkem:
If we change this now it will just make the migration from <feature> to <plugin> logic more complex.

omefire:
What do you mean ?

gorkem:
Correct, let's do the change together with the <plugin> change. My plan is to have some sort of migration logic with that as well to help old friends. At this moment, this does not affect the functionality a bit. It is just the syntax used on config.xml. I personally do not see any value on changing from param -> variable, and I do not think this was discussed and agreed to be variable.

omefire:
@gorkem , I disagree with your statement that it does not affect functionality. As it currently stands, the code doesn't work for plugins that have variables. it fails on restoring !

omefire:
It currently doesn't work in 100% of the cases but I do agree we can postpone the fix. 

I don't really care what the name of the element is : '<variable>' or something else. As long as it works, I'm fine with it. 

Although it is NOT written anywhere that '<variable>' is the name to be used, the code is currently saying a different story(long before this change). In fact, "ConfigParser.getFeature(id)" does assume that variables are under the '<variable>' element name. That's why changing things to '<variable>' fixes this issue.

Like I said, I'm fine if instead you want to modify the name to something else.

omefire:
It currently doesn't work in 100% of the cases but I do agree we can postpone the fix. 

It doesn't really matter what the name of the element is : '<variable>' or something else. As long as it works, I'm fine with it. 

Although it is NOT written anywhere that '<variable>' is the name to be used, the code is currently saying a different story(long before this change). In fact, "ConfigParser.getFeature(id)" does assume that variables are under the '<variable>' element name. That's why changing things to '<variable>' fixes this issue.

Like I said, I'm fine if instead you want to modify the name to something else.

gorkem:
Ahhh. I see what you mean now. I think we should fix this on the Config.getFeature(id) so that the existing information on config.xml files are not discarded.

omefire:
Fixing it in Config.getFeature(id) would mean that we would have to differentiate between `<param>` types : variables and non-variables. I think that could be unecessarily complex, require special logic and create more subtle bugs. So, I'm in favor on fixing the syntax itself.

omefire:
Fixing it in Config.getFeature(id) would mean that we would have to differentiate between `<param>` types : variables and non-variables. I think that would be unecessarily complex, require special logic and create more subtle bugs. So, I'm in favor on fixing the syntax itself. It also has the benefit of conveying its goal clearer.

gorkem:
I agree, but this is temporary. Special params are going away with the plugin tag. And ConfigParser's current implementation already does a special treatment of FEATURE_SPECIAL_PARAMS. 

I have a commit that I am testing that fixes the getFeature(id) which fixes the tests that depend on the variable too. 

In case interested, this is how it all started https://github.com/apache/cordova-lib/pull/63

omefire:
@gorkem, the change you've committed to fix this issue ( https://github.com/apache/cordova-lib/commit/4c0fec7d30df238cfc4c834a4235b2c8fcf330ad ) introduces another bug : the variables list now contains all the `<param>` such as 'id', 'url', 'installPath', which is not correct !

This is exactly the kind of subtle bugs I was afraid would sneak in if we used the approach you're currently using to fix this issue.

Related PR, with alternative fix : https://github.com/apache/cordova-lib/pull/174

PR #175, opened 2015-03-03
CB-8551 npm registry integration for fetching and removing plugins

stevengill:


cordova-lib/src/plugman/registry/registry.js
```diff
@@ -218,10 +205,18 @@ module.exports = {
...
-function initSettings() {
+function initSettings(npm) {
     var settings = module.exports.settings;
+    var registryURL = 'http://registry.cordova.io';
+
+    //if npm is true, use npm registry
+    if(npm) {
+        registryURL = 'http://registry.npmjs.org';
+    }
+
```
TimBarham: Minor nit - rather than setting registryURL to one value, then potentially setting it to another value, why not just:

    var registryURL = npm ? 'http://registry.npmjs.org' : 'http://registry.cordova.io';

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -236,15 +231,22 @@ function initSettings() {
...
         logstream: fs.createWriteStream(path.resolve(plugmanConfigDir, 'plugman.log')),
         userconfig: path.resolve(plugmanConfigDir, 'config'),
         'cache-min': oneDay
     });
+
+    // if npm is true, use npm registry. 
+    // ~/.plugman/config overides the above config if it exists. 
+    // Need to reset the registry value in settings 
+    if(npm) {
+        settings.registry = 'http://registry.npmjs.org';
```
TimBarham: There's no way to avoid this being overwritten in the first place? If not, rather than duplicating the string, can you either:

1. Store this string in a variable and use the variable twice.
2. Simply use the existing registryURL here (since it must be set to the npm registry).

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -322,3 +324,102 @@ function makeRequest (method, where, what, cb_) {
...
+    });
+}
+
+
+/**
+ * @method fetchPlugReg
+ * @param {Array} with one element - the plugin id or "id@version"
+ * @return {Promise.<string>} Promised path to fetched package.
+ */
+function fetchPlugReg(plugin, client) {
```
TimBarham: This method is identical to `fetchNPM()` other than whether or not we pass `true` to `initSettings()`, and the text of some log messages. Could you create a single helper function modifies those things based on a parameter?

cordova-lib/src/plugman/uninstall.js
```diff
@@ -47,7 +47,7 @@ function uninstall(platform, project_dir, id, plugins_dir, options) {
     options.is_top_level = true;
     options.pluginInfoProvider = options.pluginInfoProvider || new PluginInfoProvider();
     plugins_dir = plugins_dir || path.join(project_dir, 'cordova', 'plugins');
-
+
```
TimBarham: Looks like the only change was adding some spaces :)

cordova-lib/src/cordova/plugin.js
```diff
@@ -238,7 +239,21 @@ module.exports = function plugin(command, targets, opts) {
                 return opts.plugins.reduce(function(soFar, target) {
                     // Check if we have the plugin.
                     if (plugins.indexOf(target) < 0) {
-                        return Q.reject(new CordovaError('Plugin "' + target + '" is not present in the project. See `'+cordova_util.binname+' plugin list`.'));
+                        // Convert target from package-name to package-id if necessary
+                        var keys = Object.keys(pluginMapper);
+                        //Traverse through pluginMapper values to see if it equals our target.
+                        //Cordova-plugin-device would get changes to org.apache.cordova.device
+                        for (var i = 0; i < keys.length; i++) {
```
agrieve: Might as well export the reverse mapping from pluginMapper so that you don't need to do this linear search. Exposing the map as a sub-key of the module is a good idea anyways, since doing it top-level restricts you from ever adding another symbol to it.

cordova-lib/src/plugman/fetch.js
```diff
@@ -147,6 +146,9 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
         checkID(options.expected_id, result.pinfo);
         metadata.save_fetch_metadata(plugins_dir, result.pinfo.id, { source: result.fetchJsonSource });
         return result.dest;
+    })            
+    .fail(function(error) {
```
agrieve: This is a no-op, is it not?

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -236,15 +231,22 @@ function initSettings() {
...
         cache: plugmanCacheDir,
-        registry: 'http://registry.cordova.io',
+        registry: registryURL,
         logstream: fs.createWriteStream(path.resolve(plugmanConfigDir, 'plugman.log')),
         userconfig: path.resolve(plugmanConfigDir, 'config'),
         'cache-min': oneDay
     });
+
+    // if npm is true, use npm registry. 
+    // ~/.plugman/config overides the above config if it exists.
```
agrieve: isn't this by design? Wouldn't you still want to be able to set this?

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -322,3 +324,102 @@ function makeRequest (method, where, what, cb_) {
...
+        return unpack.unpackTgz(package_tgz, pluginDir);
+    })
+    .fail(function(error) {
+        events.emit('log', 'Fetching from cordova plugin registry failed');
+        return Q.reject(error);
+    });
+}
+
+/**
+ * @method checkPluginID
```
agrieve: I don't think we need a comment to tell us what the function name is :)

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -322,3 +324,102 @@ function makeRequest (method, where, what, cb_) {
...
+ * @param {Array} with one element - the plugin id or "id@version"
+ * @return {Promise.<string>} Promised path to fetched package.
+ */
+function checkPluginID(plugin) {
+    //if plugin id is not reverse domain name style, skip CPR and fetch from npm
+
+    //Create regex to for digits, words and dashes and three dots in plugin ids which excludes @VERSION.
+    var re = /([\w-]*\.[\w-]*\.[\w-]*\.[\w-]*[^@])/;
```
agrieve: I think we require only two dots. Might make sense to be a bit less strict here: `/([^@]*\.[^@]*\.[^@]*)/.exec(plugin)`

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -322,3 +324,102 @@ function makeRequest (method, where, what, cb_) {
...
+        return Q.reject(error);
+    });
+}
+
+/**
+ * @method checkPluginID
+ * @param {Array} with one element - the plugin id or "id@version"
+ * @return {Promise.<string>} Promised path to fetched package.
+ */
+function checkPluginID(plugin) {
```
agrieve: "checkPluginId" doesn't make this method's purpose obvious. How splitting it into `isValidCprName()` and `warnIfRenamed()`.

Would also be a bit more obvious if it returned true/false instead of a promise, and to do the logging at the call site.

cordova-lib/package.json
```diff
@@ -19,6 +19,7 @@
   "dependencies": {
     "bplist-parser": "0.0.6",
     "cordova-js": "3.8.0",
+    "cordova-registry-mapper": "0.0.2",
```
agrieve: Where is the source for this?

mmocny: Looks like: https://github.com/stevengill/cordova-registry-mapper

cordova-lib/src/cordova/plugin.js
mmocny: I don't know why we are iterating anyway?  Its already old_id->new_id.  Can't we just `.hasOwnProperty()` ?

mmocny: Nevermind, I see its doing reverse mapping.

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -218,10 +205,18 @@ module.exports = {
 
 /**
  * @method initSettings
+ * @param {Boolean} using npm registry
  * @return {Promise.<Object>} Promised settings.
  */
-function initSettings() {
+function initSettings(npm) {
```
mmocny: nit: `npm` -> `useNpmRegistry`.  Otherwise it looks like a reference to the node module.

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -322,3 +324,102 @@ function makeRequest (method, where, what, cb_) {
...
+function checkPluginID(plugin) {
+    //if plugin id is not reverse domain name style, skip CPR and fetch from npm
+
+    //Create regex to for digits, words and dashes and three dots in plugin ids which excludes @VERSION.
+    var re = /([\w-]*\.[\w-]*\.[\w-]*\.[\w-]*[^@])/;
+    var pluginID = plugin.match(re);
+    //If pluginID equals null, plugin is not reverse domain name style
+    if(pluginID === null) { 
+        events.emit('verbose', 'Skipping CPR');
+        //Q.reject will send us straight to the fail method which is where fetchNPM gets called.
```
mmocny: Bravo, thanks for adding straight-to-npm.

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -322,3 +324,102 @@ function makeRequest (method, where, what, cb_) {
...
+        return Q.reject();
+    } else {
+        //Reverse domain name style plugin ID
+        //Check if a mapping exists for the pluginID
+        //if it does, warn the users to use package-name
+        var packageName = pluginMapper[pluginID[0]];
+        if(packageName) {
+            events.emit('log', 'WARNING: ' + plugin + ' has been renamed to ' + 
+                    packageName + ' and moved to npm. Please use `cordova plugin add ' + 
+                    packageName + '` next time.');
```
mmocny: Suggestion: `'next time'` -> 'Automatically converting to NAME and continuing`

Basically let user know that it didn't fail.

mmocny:
Overall, impressed with the lack of significant impact.

cordova-lib/package.json
stevengill: Yup. I could donate it to Apache if it makes sense. Right now it is easy for me to do releases for it without having to go through apache process. I imagine we will see some PRs coming in for it from other plugin devs who rename.

cordova-lib/src/cordova/plugin.js
stevengill: I agree it should just be exported in pluginMapper. I'll do it up.

cordova-lib/src/plugman/registry/registry.js
stevengill: It actually doesn't convert and continue. It just warns and continues. `cordova plugin add org.apache.cordova.device` would output `WARNING: org.apache.cordova.device has been renamed to cordova-plugin-device and moved to npm. Please use `cordova plugin add cordova-plugin-device next time.` It would then continue on and fetch org.apache.cordova.device from CPR and you would see the log for it succeeding. 

Users should still be able to see that the plugin succeeded with the rest of the output in their terminal. But if you feel like we should be more clear with this message, I can change it up. It is important we make sure our users understand what's up!

cordova-lib/src/plugman/registry/registry.js
stevengill: Just because you asked so nicely! :smile:

cordova-lib/src/plugman/registry/registry.js
stevengill: Good call. I will update it

cordova-lib/src/plugman/registry/registry.js
stevengill: Good suggestions. Agreed

cordova-lib/src/plugman/registry/registry.js
stevengill: LIES!

I just followed the format of the other comments in this file. :smile:

cordova-lib/src/plugman/registry/registry.js
stevengill: Yeah probably. I started off having more differences between the two, but ended up slowly making these very similar. I'll look into simplifying it more.

cordova-lib/src/plugman/registry/registry.js
stevengill: I think it is by design. I don't know the reason though. 

As far as I can tell, `~/.plugman/config` is used by `plugman publish`. We probably don't actually want to change the registry value in the config itself. 

What do you mean by wouldn't you want to be able to still set this?

cordova-lib/src/plugman/registry/registry.js
stevengill: Good call

cordova-lib/src/plugman/fetch.js
stevengill: I'm not 100% sure, but I think I needed to add this during one of my iterations. It very well could be no-op now. I will see if I can hit it and if it is needed.

cordova-lib/src/plugman/registry/registry.js
stevengill: If `~/.plugman/config` exists, it doesn't matter what I passed in above. It gets overwritten by what's in that `config` file. That is why I have to do the check down here. 

I could move the string to a variable. Doesn't really add much though.

cordova-lib/src/plugman/registry/registry.js
stevengill: TBH, I prefer `if(npm) {...}` to `?`. I think it is just easier to read and leaves us open to add other changes if we realize we need to. If I am unaware of some performance boost or something, please let me know.

cordova-lib/src/plugman/registry/registry.js
mmocny: Okay, thanks for clarification.  I guess I agree with the choices, but think we could change the messaging to be more clear.  Perhaps something like:

`"WARNING: 'org.apache.cordova.device' has been renamed to 'cordova-plugin-device' (See: BLOGPOST/GITDOC).  You may not getting the latest version!  We suggest you `cordova plugin rm org.apache.cordova.device` and `cordova plugin add cordova-plugin-device`."`

(i.e. call out what to do to fix, don't disrupt the message with mention of npm)

cordova-lib/src/plugman/registry/registry.js
agrieve: I think it's the same that Tim points out. Normally users don't have a registry override in their plugman/config, but when they do, we should use it. Probably the logic should actually be:

if (custom registry) {
  use it only
} else {
  try CPR and fallback to NPM
}

It's pretty fringe though, so maybe just add a TODO to support custom registries again?

cordova-lib/src/plugman/registry/registry.js
TimBarham: It's just the general principle of avoiding duplicating strings (makes maintenance easier, for a start).

jsoref: fwiw, I'm +1 for not duplicating strings. I tend to lose a lot of time having to adjust things...

cordova-lib/src/plugman/registry/registry.js
stevengill: I think custom registry should still work. 

Ex: `plugman config set registry "MYSPECIALREGISTRY"`
This updates the registry field in `~/.plugman/config`

By the time the code hits line 243, the registry value equals the registry value from config. If npm is true, then it would set it to use npm. But this is currently set to happen on the fail (fallback). 

So it would hit custom registry and then fallback to npm.

cordova-lib/src/plugman/registry/registry.js
stevengill: Fair enough.

cordova-lib/src/plugman/registry/registry.js
jsoref: fwiw, i agree w/ the suggestion.

cordova-lib/src/plugman/registry/registry.js
purplecabbage: +1 
// suggestion:
const NPM_REG_URL = 'http://registry.npmjs.org';
const CPR_REG_URL = 'http://registry.cordova.io';

cordova-lib/src/plugman/registry/registry.js
TimBarham: Ok, fair enough - I get some people don't like `?:`. The usual alternative is:

    var registryURL;
    if (npm) {
        registryURL = 'http://registry.npmjs.org';
    } else {
         registryURL = 'http://registry.cordova.io';
    }

I know this is slightly more verbose, but I think it is cleaner code than assigning something, then immediately assigning something else if a condition is true. That said, it's only a minor nit so I won't complain very loudly if you leave it as is :).

cordova-lib/src/plugman/registry/registry.js
agrieve: ahh, gotcha.

PR #177, opened 2015-03-04
CB-8596 Expose APIs to retrieve platforms and plugins saved in config.xml

omefire:
CB-8596 Expose APIs to retrieve platforms and plugins saved in config.xml  so as to abstract clients away from the config.xml details.

### DIFF
--- a/cordova-lib/spec-cordova/project-metadata-apis.spec.js
+++ b/cordova-lib/spec-cordova/project-metadata-apis.spec.js
@@ -80,8 +80,6 @@ describe('retrieval of project metadata', function(){
             var plugin = plugins[0];
             var retrievedPlugin =  plugin.id + '@' + plugin.version;
             expect(retrievedPlugin).toBe(devicePlugin);
-        }).then(function(){
-            done();
-        });
+        }).fin(done);
     });
 });
--- a/cordova-lib/spec-cordova/project-metadata-apis.spec.js
+++ b/cordova-lib/spec-cordova/project-metadata-apis.spec.js
@@ -22,20 +22,20 @@ var cordova = require('../src/cordova/cordova'),
     shell   = require('shelljs'),
     path    = require('path');
 
-var appId = 'org.testing';
-var appName = 'ProjectMetadataRetrievalTests';
-
-var configNormal = {
-    lib: {
-        www: {
-            url: path.join(__dirname, 'fixtures', 'base', 'www'),
-            version: 'testProjectMetadataRetrieval',
-            id: appName
-        }
-    }
-};
-
 describe('retrieval of project metadata', function(){
+
+    var appId = 'org.testing';
+    var appName = 'ProjectMetadataRetrievalTests';
+
+    var configNormal = {
+        lib: {
+            www: {
+                url: path.join(__dirname, 'fixtures', 'base', 'www'),
+                version: 'testProjectMetadataRetrieval',
+                id: appName
+            }
+        }
+    };
 
     var tmpDir = helpers.tmpDir('project_metadata_test');
     var projectRoot = path.join(tmpDir, appName);
--- a/cordova-lib/spec-cordova/project-metadata-apis.spec.js
+++ b/cordova-lib/spec-cordova/project-metadata-apis.spec.js
@@ -52,7 +52,7 @@ describe('retrieval of project metadata', function(){
     });
     
     it('retrieve platforms and plugins saved in config.xml', function(done){
-        var devicePlugin = 'org.apache.cordova.device@0.2.11';
+        var devicePlugin = 'org.apache.cordova.device@0.2.11'; 
         var androidPlatform = 'android@3.7.0';
 
         process.chdir(tmpDir);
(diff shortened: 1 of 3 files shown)


omefire:
After migrating platforms & plugins to package.json, This implementation will be updated, but clients will remain unchanged.

omefire:
@agrieve , @stevengill  please review.

omefire:
hmmm, the build failure is something that seem to happen randomly.

cordova-lib/src/cordova/project_metadata.js
```diff
@@ -0,0 +1,60 @@
...
+// Return type is a promise that is fulfilled with a list of objects with name and version properties. e.g: [{name: 'android', version: '3.5.0'}, {name: 'wp8', version: 'C:/path/to/platform'}, {name: 'ios', version: 'git://...'}]
+// ToDo: Once we move to npm, this function should be updated to rely on npm instead
+function getPlatforms(projectRoot){
+    var xml = cordova_util.projectConfig(projectRoot);
+    var cfg = new ConfigParser(xml);
+    return Q(cfg.getEngines());
+}
+
+// Returns all the plugins that are currently saved into config.xml
+// Return type is a promise that is fulfilled with a list of objects with name and version properties. e.g: [ {id: 'org.apache.cordova.device', name: 'Device', APP_ID: 'my-app-id', APP_NAME: 'my-app-name'} ]
```
nikhilkh: Consider using JSDoc comments. It's got good syntax and structure to specify return values and types.

cordova-lib/src/cordova/project_metadata.js
```diff
@@ -0,0 +1,60 @@
...
+// ToDO: Once we move to npm, this function should be updated to rely on npm instead
+function getPlugins(projectRoot){
```
nikhilkh: Should we return promises from an API? This couples our implementation with Q. We should ideally return the data itself and document the shape of the data rather than return a Q based promise

nikhilkh:
Looks like your new tests are timing out and you might need to increase the timeout value.

cordova-lib/src/cordova/project_metadata.js
omefire: valid point! however, we already return promises with all our other APIs.

nikhilkh: Ah..in that case it makes sense to remain consistent and return Q. Is this API documented? JSDoc comments are easier to read and serve as good documentation.

cordova-lib/src/cordova/project_metadata.js
omefire: Thanks for the suggestion. definitely cleaner comments...
I just pushed cleaner, more readable comments.

nikhilkh:
I noticed that the plugin metadata that is exposed is the id and version number. What if the source is a git URL - how is that exposed?

cordova-lib/spec-cordova/project-metadata-apis.spec.js
```diff
@@ -0,0 +1,85 @@
...
+        shell.mkdir('-p', tmpDir);
+    });
+
+    afterEach(function() {
+        process.chdir(path.join(__dirname, '..'));  // Windows: Needed to rm the dir on Windows.
+        shell.rm('-rf', tmpDir);
+    });
+    
+    it('retrieve platforms and plugins saved in config.xml', function(done){
+        var devicePlugin = 'org.apache.cordova.device@0.2.11';
```
nikhilkh: Wonder if you need to add test case for git url being specified?

cordova-lib/src/cordova/project_metadata.js
```diff
@@ -0,0 +1,64 @@
...
+ *  ToDo: Once we move to npm, this function should be updated to rely on npm instead
+ */
+function getPlatforms(projectRoot){
+    var xml = cordova_util.projectConfig(projectRoot);
+    var cfg = new ConfigParser(xml);
+    return Q(cfg.getEngines());
+}
+
+/** Returns all the plugins that are currently saved into config.xml
+ *  @return Promise<{{name: string, version: string}[]}> A promise to the list of {name: platformName, version: platformVersion} objects
```
nikhilkh: Did you mean to say plugin name or really plugin id - instead of platformName.

cordova-lib/src/cordova/project_metadata.js
```diff
@@ -0,0 +1,64 @@
...
+ *  e.g: [ {id: 'org.apache.cordova.device', name: 'Device', APP_ID: 'my-app-id', APP_NAME: 'my-app-name'} ]
```
nikhilkh: This example is confusing and does not match with the comment above.

nikhilkh:
This PR needs to be closed as it is a dup of #191.

kamrik:
Was this all taken care of with #191 ?

nikhilkh:
Yes, #191 subsumes this PR. I believe @omefire is on vacation and he will be able to close this when he's back.

PR #178, opened 2015-03-06
CB-8521 `cordova plugin save` should save all installed plugins and their version/git-url/folder-location into config.xml

omefire:
CB-8521 `cordova plugin save` should save all installed plugins and their version/git-url/folder-location into config.xml

### DIFF
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -135,7 +135,10 @@ module.exports = function plugin(command, targets, opts) {
 
                         // Fetch the plugin first.
                         events.emit('verbose', 'Calling plugman.fetch on plugin "' + target + '"');
-                        return plugman.raw.fetch(target, pluginsDir, { searchpath: searchPath, noregistry: opts.noregistry, link: opts.link, pluginInfoProvider: pluginInfoProvider, variables: opts.cli_variables });
+
+                        return plugman.raw.fetch(target, pluginsDir, { searchpath: searchPath, noregistry: opts.noregistry, link: opts.link, 
+                                                                       pluginInfoProvider: pluginInfoProvider, variables: opts.cli_variables,
+                                                                       is_top_level: true });
                     })
                     .then(function(dir){
                         // save to config.xml 
@@ -312,7 +315,7 @@ module.exports = function plugin(command, targets, opts) {
 };
 
 function save(projectRoot, opts){
-    
+     
     var xml = cordova_util.projectConfig(projectRoot);
     var cfg = new ConfigParser(xml);
 
@@ -328,7 +331,7 @@ function save(projectRoot, opts){
         return Q.reject('fetch.json file does not exist: ' + err.message);
     }
 
-    // Then, save plugins and their sources
+    // Then, save top-level plugins and their sources
     return Q().then(function(){
         var jsonFile = path.join(projectRoot, 'plugins', 'fetch.json');
         return JSON.parse(fs.readFileSync(jsonFile, 'utf-8'));
@@ -336,6 +339,12 @@ function save(projectRoot, opts){
         Object.keys(plugins).forEach(function(pluginName){
             var plugin = plugins[pluginName];
             var pluginSource = plugin.source;
+            
+            // If not a top-level plugin, skip it, don't save it to config.xml
+            if(!plugin.is_top_level){
+                return;
+            }
+
             var location = pluginSource.url|| pluginSource.path || pluginSource.id;
 
             var idParam = {name: 'id', value: pluginName};
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -277,6 +277,11 @@ module.exports = function plugin(command, targets, opts) {
                                 events.emit('results', 'config.xml entry for ' +target+ ' is removed');
                             }
                         }
+                    }).then(function(){
+                        // Remove plugin from fetch.json
+                        events.emit('verbose', 'Removing plugin ' + target + ' from fetch.json');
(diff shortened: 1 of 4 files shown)


omefire:
@vladimir-kotikov , please review.

cordova-lib/src/cordova/plugin.js
```diff
@@ -134,7 +135,7 @@ module.exports = function plugin(command, targets, opts) {
 
                         // Fetch the plugin first.
                         events.emit('verbose', 'Calling plugman.fetch on plugin "' + target + '"');
-                        return plugman.raw.fetch(target, pluginsDir, { searchpath: searchPath, noregistry: opts.noregistry, link: opts.link, pluginInfoProvider: pluginInfoProvider});
+                        return plugman.raw.fetch(target, pluginsDir, { searchpath: searchPath, noregistry: opts.noregistry, link: opts.link, pluginInfoProvider: pluginInfoProvider, variables: opts.cli_variables });
```
nikhilkh: Nit: Consider splitting this into two lines - it's rather wide currently - preventing readability.

gorkem:
I think this saves **all** the installed plugins to config.xml. The dependency plugins ( those plugins that were installed as a dependency of another) should never be saved to the config.xml. This may cause the dependency plugins to be restored as a top level plugin.

omefire:
Interesting. @gorkem, could you give an example of a plugin that has dependency plugins ? I'll carry some testing regarding that.

gorkem:
I was just testing com.ibm.mobile.cordova.ibmpush for something else which seems to pull in another large plugin

cordova-lib/src/cordova/plugin.js
```diff
@@ -284,6 +284,11 @@ module.exports = function plugin(command, targets, opts) {
                                 }
                             }
                         }
+                    }).then(function(){
```
nikhilkh: Nit: Consider adding a space after ')' - aids readability. This is an issue in most of the code below.

omefire:
Thanks for raising this issue, @gorkem. I just pushed some changes to handle `dependent` plugins and only save `top-level` plugins in config.xml.

cordova-lib/src/cordova/plugin.js
```diff
@@ -234,7 +239,6 @@ module.exports = function plugin(command, targets, opts) {
             if (!targets || !targets.length) {
                 return Q.reject(new CordovaError('No plugin specified. Please specify a plugin to remove. See `'+cordova_util.binname+' plugin list`.'));
             }
-
```
vladimir-kotikov: Unnecessary change

cordova-lib/src/cordova/plugin.js
```diff
@@ -267,7 +271,8 @@ module.exports = function plugin(command, targets, opts) {
                     .then(function() {
                         // TODO: Should only uninstallPlugin when no platforms have it.
                         return plugman.raw.uninstall.uninstallPlugin(target, path.join(projectRoot, 'plugins'));
-                    }).then(function(){
+                    })
+                    .then(function(){
```
vladimir-kotikov: Unnecessary change

cordova-lib/src/cordova/plugin.js
```diff
@@ -278,9 +283,16 @@ module.exports = function plugin(command, targets, opts) {
...
+                        // Remove plugin from fetch.json
+                        events.emit('verbose', 'Removing plugin ' + target + ' from fetch.json');
+                        var pluginsDir = path.join(projectRoot, 'plugins');
+                        metadata.remove_plugin(pluginsDir, target);
                     });
                 }, Q());
-            }).then(function() {
+            })
+            .then(function() {
```
vladimir-kotikov: Unnecessary change

cordova-lib/src/cordova/plugin.js
```diff
@@ -295,14 +307,87 @@ module.exports = function plugin(command, targets, opts) {
...
+
+    // It might be the case that fetch.json file is not yet existent.
+    // for example: when we have never ran the command 'cordova plugin add foo' on the project
+    // in that case, there's nothing to do except bubble up the error
+    function onFileNotFoundException(err){
+        return Q.reject(err.message);
+    }
+
+    // Then, save top-level plugins and their sources
+    return Q().then(function(){
```
vladimir-kotikov: Since that there is no async code in this function, IMHO it'll be better to avoid wrapping it into a promise and just return fullfilled/rejected promise.

cordova-lib/src/plugman/fetch.js
```diff
@@ -144,7 +144,10 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
         });
     }).then(function(result){
         checkID(options.expected_id, result.pinfo);
-        metadata.save_fetch_metadata(plugins_dir, result.pinfo.id, { source: result.fetchJsonSource });
+        var data = { source: result.fetchJsonSource };
+        data.is_top_level = options.is_top_level;
```
vladimir-kotikov: IMHO it's better to use an object literal here instead of assigning properties.

```javascript
var data = {
    source: result.fetchJsonSource,
    is_top_level: options.is_top_level,
    variables: options.variables || {}
};
```

cordova-lib/src/plugman/util/metadata.js
```diff
@@ -59,3 +59,10 @@ exports.save_fetch_metadata = function(pluginsDir, pluginId, data) {
     fs.writeFileSync(fetchJsonPath, JSON.stringify(metadataJson, null, 4), 'utf-8');
 };
 
+exports.remove_plugin = function(pluginsDir, pluginId){
```
vladimir-kotikov: IMHO it's better to rename this similar to other methods in this file (`remove_fetch_metadata` for example)

PR #183, opened 2015-03-10
CB-8595 Merge platforms.js from cordova & plugman CB-8595

kamrik:
Till now we had two separate places for platform specific code
cordova/metadata exposed via cordova/platforms.js
plugman/platforms exposed via plugman/platforms.js

This change merges the two `platforms.js` files into `platforms/platforms.js`
containing a single class exposing functionality from both places.

The class is instantiated via
prj = platforms.getPlatformProject(platform, project_dir)
`prj` can then be used just like the platform parsers were used before in cordova
e.g. `prj.www_dir()` or the handlers in plugman e.g: `prj.parseProjectFile()`.

This is a compatibility layer so that the platform specific code could be
easier consolidated and eventually moved to the platform repos.


cordova-lib/src/PluginInfo.js
```diff
@@ -288,6 +293,19 @@ function PluginInfo(dirname) {
...
+    self.getFilesAndFrameworks = getFilesAndFrameworks;
+    function getFilesAndFrameworks(platform) {
+        var items = [];
+        // Please avoid changing the order of the calls below, files will be
+        // installed in this order.
+        items = items.concat(self.getSourceFiles(platform));
+        items = items.concat(self.getHeaderFiles(platform));
+        items = items.concat(self.getResourceFiles(platform));
+        items = items.concat(self.getFrameworks(platform));
+        items = items.concat(self.getLibFiles(platform));
```
mmocny: Tip: `concat` actually accepts any number of arguments and will concat all in one go.

Can also do: `[].concat.apply([], [ self.getSourceFiles(..), self.getHeaderFiles(..), ...])`

Which I think looks nicest.

cordova-lib/src/platforms/platforms.js
```diff
@@ -0,0 +1,102 @@
...
+    var self = this;
+    self.root = platformRootDir;
+    self.platform = platform;
+    var ParserConstructor = require(platforms[platform].parser_file);
+    self.parser = new ParserConstructor(platformRootDir);
+    self.handler = require(platforms[platform].handler_file);
+
+    // Expos all public methods from the parser and handler, properly bound.
+    PARSER_PUBLIC_METHODS.forEach(function(method) {
+        if (self.parser[method]) {
```
mmocny: Is it an error to not be there?  Should we throw if this fails?

cordova-lib/src/platforms/platforms.js
```diff
@@ -0,0 +1,102 @@
...
+            self[method] = self.parser[method].bind(self.parser);
+        }
+    });
+
+    HANDLER_PUBLIC_METHODS.forEach(function(method) {
+        if (self.handler[method]) {
```
mmocny: ditto.

cordova-lib/src/platforms/platforms.js
```diff
@@ -0,0 +1,102 @@
...
+    self.getUninstaller = function(type) {
+        return self.handler[type].uninstall;
+    };
+}
+
+// getPlatformProject() should be the only method of instantiating the
+// PlatformProject classes for now.
+function getPlatformProject(platform, platformRootDir) {
+    var cached = cachedProjects[platformRootDir];
+    if (cached && cached.platform == platform) {
```
mmocny: If we have a `cached` PlatformProject already, shouldn't we just assert that `cached.platform == platform` inside the if?  If the platform is wrong, theres a bug right?

cordova-lib/src/platforms/platforms.js
kamrik: Added the if because windows parser doesn't have update_overrides() but apparently update_overrides() is only used internally inside the parsers so I'll remove it from PARSER_PUBLIC_METHODS and remove the if so it will throw if no method is found.

cordova-lib/src/platforms/platforms.js
kamrik: For the handlers the if is still needed because there is actually code that checks if the handler has a parseProjectFile() function before calling it.

cordova-lib/src/plugman/uninstall.js
```diff
@@ -293,58 +293,25 @@ function runUninstallPlatform(actions, platform, project_dir, plugin_dir, plugin
...
     var plugin_dir = pluginInfo.dir;
-    var platform_modules = require('./platforms');
-    var handler = platform_modules[platform];
-    www_dir = www_dir || handler.www_dir(project_dir);
+    var handler = platform_modules.getPlatformProject(platform, project_dir);
+    www_dir = www_dir || handler.www_dir();
     events.emit('log', 'Uninstalling ' + plugin_id + ' from ' + platform);
 
+    var pluginItems = pluginInfo.getFilesAndFrameworks(platform);
     var assets = pluginInfo.getAssets(platform);
```
mmocny: Why are assets different?

cordova-lib/src/PluginInfo.js
kamrik: Good idea, will change.

cordova-lib/src/plugman/uninstall.js
kamrik: Assets are not copied during installation, only during prepare (because www dir is blown on prepare). It wouldn't harm to copy them here as well, just some duplication, but they are also currently dealt with by directly calling common.asset.install(), not from the handlers with a slightly different function signature, so I left it as is for now.

cordova-lib/src/platforms/platforms.js
kamrik: The tests sometimes reuse the same dir for several platforms. Added the platform check after spending a couple of hours figuring this out. Not sure if there is a viable real world case for using the same dir for different platforms.

mmocny: Okay, Fair enough.

cordova-lib/src/PluginInfo.js
purplecabbage: I would recommend :
var items = [];
items = items.concat(self.getSourceFiles(platform),
                                  self.getHeaderFiles(platform),
                                  ... );
// unnecessarily tricky code is a barrier to some, clarity is king

kamrik:
Addressed the comments and merged as a single squashed commit.
Thanks for reviewing.

PR #186, opened 2015-03-13
Allow hyphen in platform name

crisvergara:
amazon-fireos was not being recognized as a platform, because 'cordova-amazon-fireos' did not match the given regex.

This was being returned when attempting to add the platform to a project:

`The provided path does not seem to contain a Cordova platform: /home/cris/.cordova/lib/npm_cache/cordova-amazon-fireos/3.6.3/package`


kamrik:
Looks good, merging.

PR #187, opened 2015-03-18
CB-8703: Add support for semver and device-specific targeting of config-file to Windows

robpaveza:
Allows config-file to target a virtual "package.appxmanifest" (already supported) and specify versions of Windows via new versions and device-target attributes:

 <config-file target="package.appxmanifest" versions=">=8.1.0" ... />
 <config.file target="package.appxmanifest" device-target="phone" />

This change also modifies the related <lib-file> and <framework> elements for Windows to rename the "target" attribute to "device-target" in order to align the meaning across elements. ("target" is preserved as an alias in order to be backwards-compatible).

There is a matching PR for cordova-docs.

### DIFF
--- a/cordova-lib/spec-plugman/platforms/windows.spec.js
+++ b/cordova-lib/spec-plugman/platforms/windows.spec.js
@@ -330,19 +330,19 @@ beforeEach(function () {
                         .then(function () {
                             var path = 'ItemGroup/SDKReference';
                             var incText = 'TestSDK1, Version=1.0';
-                            var targetConditions = {versions: undefined, target: undefined, arch: 'x86'};
+                            var targetConditions = {versions: undefined, deviceTarget: undefined, arch: 'x86'};
                             validateUninstalledProjects('lib-file', libfiles[0], path, incText, targetConditions, ['all']);
 
                             incText = 'TestSDK2, Version=1.0';
-                            targetConditions = {versions: '>=8.1', target: undefined, arch: undefined};
+                            targetConditions = {versions: '>=8.1', deviceTarget: undefined, arch: undefined};
                             validateUninstalledProjects('lib-file', libfiles[1], path, incText, targetConditions, ['windows', 'phone']);
 
                             incText = 'TestSDK3, Version=1.0';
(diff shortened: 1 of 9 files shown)


purplecabbage:
Personally I would much rather just drop support for Windows8 and move to an 8.1 universal only world ... then we can drop some of this wasted complexity.

Also, was there a specific need to rename target->deviceTarget ? this makes the pull request harder to review.

robpaveza:
I think our review indicated that we couldn't yet officially deprecate Windows 8, and even if we could it would require the six month window.

The reason to rename target to deviceTarget was for consistency.

TimBarham:
Jesse - we couldn't use `target` for `<config-file>`, because there was already a `target` attribute that had a different meaning. So it was decided to go with `device-target`. I requested Rob change the equivalent `target` attribute on `<framework>` and `<lib-file>` so that it was consistent across elements.

robpaveza:
@purplecabbage Can we set aside the issue of the Windows 8 deprecation for this PR?  There are still reasons we'll want to target package.appxmanifest with semver support anyway, Windows 8 or no.  Any chance we can get this PR moved along?  Thanks!

purplecabbage:
I am okay with setting aside the deprecation discussion.
Does the renaming of target->deviceTarget have any impact outside this pull request?

robpaveza:
Re: target->deviceTarget: No, all the tests pass, etc.  Since I already did the merge/rebase, I can't really back out the change, but I'll definitely keep things like that in a separate commit in future PRs.

purplecabbage:
LGTM, can I accept this?

nikhilkh:
@purplecabbage. Yes, this one's ready to go! Please help commit this one and close this PR! :bow:

purplecabbage:
This is merged into master, not sure why the pull-request didn't close.
Please review master and close. Thanks.

nikhilkh:
Thanks, @purplecabbage ! There is a corresponding update to docs here that needs someone to commit it: https://github.com/apache/cordova-docs/pull/270

PR #190, opened 2015-03-25
CB-8737 Available platforms list includes extraneous values.

TimBarham:
Using the latest sources, if you list available platforms the output includes two extraneous values (`getPlatformProject` and `PlatformProjectAdapter`):

    $ cordova platform
    Installed platforms: windows
    Available platforms: PlatformProjectAdapter, amazon-fireos, android, blackberry10, browser, firefoxos, getPlatformProject, windows8, wp8

These are methods added to the platforms object. Fix is to make them non-enumerable.

### DIFF
--- a/cordova-lib/src/platforms/platforms.js
+++ b/cordova-lib/src/platforms/platforms.js
@@ -102,5 +102,10 @@ function getPlatformProject(platform, platformRootDir) {
 }
 
 module.exports = platforms;
-module.exports.getPlatformProject = getPlatformProject;
-module.exports.PlatformProjectAdapter = PlatformProjectAdapter;
+
+// We don't want these methods to be enumerable on the platforms object, because we expect enumerable properties of the
+// platforms object to be platforms.
+Object.defineProperties(module.exports, {
+    'getPlatformProject': {value: getPlatformProject},
+    'PlatformProjectAdapter': {value: PlatformProjectAdapter}
+});


TimBarham:
@kamrik would you be able to take a look at this? Thanks!

cordova-lib/src/platforms/platforms.js
```diff
@@ -102,5 +102,10 @@ function getPlatformProject(platform, platformRootDir) {
 }
 
 module.exports = platforms;
-module.exports.getPlatformProject = getPlatformProject;
-module.exports.PlatformProjectAdapter = PlatformProjectAdapter;
+
```
kamrik: Oops, I added those but missed them in the platform list.
getPlatformProject() is used in several places throughout the code and PlatformProjectAdapter is used by a downstream project, so we need them.
Need to find a way to "hide" them for the platform listing.

kamrik:
Will take a look at this today.

cordova-lib/src/platforms/platforms.js
TimBarham: > used by a downstream project, so we need them.

Right - with this change they're still accessible (you should still be able to call them) - they're just not enumerated (in a `for in` loop for example).

kamrik: Ok, got it, looks about right, but for some reason jasmine can't spy on such property, so the tests fail. Looking into this.

kamrik: Ok found it, adding `writable: true` to `defineProperties` solves the test failures. Pushing. Thanks for the pull request.

kamrik:
Apparently I had a typo in the GitHub close command in the commit message :)
@TimBarham , could you please close the PR, thanks.

cordova-lib/src/platforms/platforms.js
TimBarham: Ah, great. Thanks Mark!

PR #191, opened 2015-03-25
CB-8596 Expose APIs to retrieve platforms and plugins from config.xml

TimBarham:
Provides third parties access to platform and plugin information stored in `config.xml` without having to go through ConfigParser.

Based on work by @omefire, with the following changes:

* Tests simplified to use predefined `config.xml`.
* Updated to use `plugin` calls instead of `feature` calls (which have been removed).
* Updates plugin metadata to return `variables` property as an array.
* Maps platform `version` to `src` if it doesn't look like a version (to handle the fact that platform stores either version or source in the `version` property).
* Removed `getProjectMetadata` method and renamed `getProjectMetadata` property to `projectMetadata`. So api is now simply `cordova.projectMetadata.getPlatforms()` and `cordova.projectMetadata.getPlugins()`.


kamrik:
LGTM. Merging.

PR #194, opened 2015-03-27
CB-8754 Auto-restoring a plugin fails when adding a platform.

TimBarham:
Start with a blank Cordova app, then enter the following:

    cordova plugin add org.apache.cordova.camera --save
    cordova plugin remove org.apache.cordova.camera
    cordova platform add browser

The final step should restore the camera plugin, but it fails with the following exception:

    TypeError: Cannot read property 'latest' of undefined
        at next (D:\GIT\Cordova\cordova-lib\cordova-lib\node_modules\npm\lib\cache.js:694:35)
        at D:\GIT\Cordova\cordova-lib\cordova-lib\node_modules\npm\lib\cache.js:682:5
        at RegClient.get_ (D:\GIT\Cordova\cordova-lib\cordova-lib\node_modules\npm\node_modules\npm-registry-client\lib\get.js:105:14)
        at RegClient.<anonymous> (D:\GIT\Cordova\cordova-lib\cordova-lib\node_modules\npm\node_modules\npm-registry-client\lib\get.js:41:12)
        at fs.js:336:14
        at D:\GIT\Cordova\cordova-lib\cordova-lib\node_modules\npm\node_modules\graceful-fs\graceful-fs.js:103:5
        at FSReqWrap.oncomplete (fs.js:99:15)

The core problem here is an issue with how we work with npm. Every time we are going to use npm (when dealing with plugins), we call `npm.load()` passing it our settings. But `npm.load()` can only be called once per session - subsequent calls are ignored. The correct approach is to call `npm.load()` without any settings (to make sure npm is loaded), then call `npm.config.set()` for each setting. This change had been made for platforms, but not plugins.

I also wanted to make sure each time we worked with npm we had a clean config (for example, if we get an npm package from the Cordova Plugin Respository, then later from the npm repository, it would try to get the second package from the CPR because that setting would still be around and not get overridden). So now any code that wants to load and init npm I pass through a central "load and restore" method. You pass this method your npm settings and the chain of promises you want to execute with those settings applied. It loads and initializes npm, executes the promises, then restores npm's configuration to what it was before we started.

Finally, once that problem was fixed there was an additional problem - we were adding plugins too early and the platform wasn't fully initialized, so we'd get an error about not being able to find the platform's `config.xml` file.

### DIFF
--- a/cordova-lib/src/plugman/registry/registry.js
+++ b/cordova-lib/src/plugman/registry/registry.js
@@ -216,7 +216,7 @@ function initSettings(returnEmptySettings) {
 
     // check if settings already set
     if(settings !== null) {
-        return Q(settings);
+        return Q(returnEmptySettings ? {} : settings);
     }
 
     // setting up settings
--- a/cordova-lib/src/plugman/registry/registry.js
+++ b/cordova-lib/src/plugman/registry/registry.js
@@ -29,6 +29,7 @@ var semver = require('semver'),
     Q = require('q'),
     request = require('request'),
     pluginMapper = require('cordova-registry-mapper').oldToNew,
+    npmhelper = require('../../util/npm-helper'),
     home = process.env.HOME || process.env.HOMEPATH || process.env.USERPROFILE,
     events = require('../../events'),
     unpack = require('../../util/unpack'),
@@ -52,11 +53,8 @@ module.exports = {
      * @param {Array} args Command argument
      * @return {Promise.<Object>} Promised configuration object.
      */
-    config: function(args) {
-        return initSettings().then(function(settings) {
-            return Q.ninvoke(npm, 'load', settings);
-        })
-        .then(function() {
+    config: function (args) {
+        return initThenLoadSettingsWithRestore(function () {
             return Q.ninvoke(npm.commands, 'config', args);
         });
     },
@@ -67,9 +65,7 @@ module.exports = {
      * @return {Promise.<void>} Promise for completion.
      */
     owner: function(args) {
-        return initSettings().then(function(settings) {
-            return Q.ninvoke(npm, 'load', settings);
-        }).then(function() {
+        return initThenLoadSettingsWithRestore(function () {
             return Q.ninvoke(npm.commands, 'owner', args);
         });
     },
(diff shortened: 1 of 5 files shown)


TimBarham:
AppVeyor failed because it couldn't connect to github (a problem I was having today too) - is there a way to kickstart it going again?

cordova-lib/src/util/npm-helper.js
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * @description Calls npm.load, then initializes npm.config with the specified settings. Then executes a chain of
+ * promises that rely on those npm settings, then restores npm settings back to their previous value. Use this rather
+ * than passing settings to npm.load, since that only works the first time you try to load npm.
+ * @param {Object} settings
+ * @param {Function} promiseChain
+ */
+function loadWithSettingsThenRestore(settings, promiseChain) {
+    return loadWithSettings(settings).then(promiseChain).then(restoreSettings);
```
agrieve: nit: should restoreSettings be within a .finally() rather than a .then()?

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -223,11 +211,13 @@ module.exports = {
  * @param {Boolean} determines if we are using the npm registry
  * @return {Promise.<Object>} Promised settings.
  */
-function initSettings() {
+function initSettings(returnEmptySettings) {
     var settings = module.exports.settings;
 
     // check if settings already set
-    if(settings !== null) return Q(settings);
+    if(settings !== null) {
```
agrieve: This returns settings even when returnEmptySettings is true. should it?

agrieve:
Nice clean-up! Let me know when comments are addressed and I'll merge it in.

cordova-lib/src/plugman/registry/registry.js
TimBarham: No it should not. Good catch, thanks!

cordova-lib/src/util/npm-helper.js
TimBarham: Yes, indeed it should ... good point :)

TimBarham:
Thanks Andrew. Changes made!

kamrik:
Thanks, merging.
Tests pass ok. Appveyor or GitHub were flaky for the last couple of days.

PR #197, opened 2015-03-27
CB-8755 Plugin --save: Multiple config.xml entries don't get removed.

TimBarham:
Ensures all entries for a plugin get removed on `plugin remove --save`.


kamrik:
LGTM. Merging.

PR #198, opened 2015-03-27
CB-8741 Make plugin --save work more like npm install.

TimBarham:
1. `plugin add --save` now always saves version (in the form `^x.y.z`)
2. `plugin add --save` now always overwrites an existing entry.
3. `plugin add` now honors an existing version specified in `config.xml` when a plugin is added without a specific version, even if the `--save` flag is specified.

I also fixed a bug in `isUrl()` (in `util.js`) - `url.protocol` includes the colon, so the protocol for a windows path is something like `c:` - 2 characters, not one.


kamrik:
LGTM. Merging.

PR #200, opened 2015-04-02
CB-8784 Prepare with no platforms should restore all platforms.

TimBarham:
If `prepare` is called with no platforms specified, we prepare all installed platforms but don't restore any from `config.xml`. In this scenario, we should restore all platforms listed in `config.xml`.

### DIFF
--- a/cordova-lib/src/cordova/restore-util.js
+++ b/cordova-lib/src/cordova/restore-util.js
@@ -30,43 +30,45 @@ exports.installPluginsFromConfigXML = installPluginsFromConfigXML;
 exports.installPlatformsFromConfigXML = installPlatformsFromConfigXML;
 
 
-function installPlatformsFromConfigXML(platforms){
+function installPlatformsFromConfigXML(platforms) {
     var projectHome = cordova_util.cdProjectRoot();
     var configPath = cordova_util.projectConfig(projectHome);
     var cfg = new ConfigParser(configPath);
 
     var engines = cfg.getEngines(projectHome);
-    var targets = engines.map(function(engine){
+    var installAllPlatforms = !platforms || platforms.length === 0;
+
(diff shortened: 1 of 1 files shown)


gorkem:
I am OK with this. I think it was @mmocny who commented that we should only restore explicitly. @mmocny can you comment?

agrieve:
I think the previous conversation was around whether "cordova run android" should first "platform add" android if it's not there (instead of erroring out). We failed to gain consensus that this would be desirable (although it's what we do in `cca`, and I find it super convenient). 

With the addition of platforms in config.xml, I think it's more reasonable to both:
- Restore platforms if none specified (like you're doing here)
- Restore the given platform(s) if they are not yet added but are listed in the config.xml

Still, would be good to start a [discuss] thread about this since it was recently talked about without consensus

mmocny:
+1 to Andrews summary.

purplecabbage:
+1 Restore platforms if none specified (like you're doing here)
+1 Restore the given platform(s) if they are not yet added but are listed in the config.xml
+1  to this pull request

The `cordova run missingPlatform` discussion is somewhat different.  We don't know for certain that that is the user's intent except perhaps if they explicitly said `cordova run android@3.7.1` 
I want to avoid adding android@3.8.0 to a project that is perhaps using an older version of other platforms.  But this could/should be in the discuss thread ...

TimBarham:
Thanks Jesse. I had followed that previous conversation. My thinking here was that we decided to auto-restore on prepare. The gist seemed to be that we would auto-restore any platforms we were preparing. Since specifying no platforms means prepare all, it seemed consistent that in that scenario we should also auto-restore all platforms (and as far as I was aware that wasn't the sticking point of previous discussions).

PR #202, opened 2015-04-06
CB-8799 Save plugin/platform src and version to 'spec' attribute.

TimBarham:
When `--save` option is used, plugin and platform source location or version value is saved to `spec` attribute in `config.xml` (source location takes priority over version). This replaces `version` attribute for platforms, and `src` and `version` attributes for plugins.

Also:

* Updated semver dependency to 2.1.0 to support caret range.
* Ensure when save platform version (to platforms.json and config.xml) we save the version that was actually installed.
* Ensure when save platform version to config.xml we save it as a caret range.

Note that if it is decided we prefer to use `src` or `version` attributes instead of `spec`, a branch with the equivalent change but using the `src` attribute can be found [here](https://github.com/apache/cordova-lib/compare/master...MSOpenTech:CB-8799-src), and for the `version` attribute [here](https://github.com/apache/cordova-lib/compare/master...MSOpenTech:CB-8799-version).


nikhilkh:
@gorkem Any chance you can review this today? Ideally, this can be part of the next TOOLS release.

PR #205, opened 2015-04-08
CB-8807 Platform Add fails to add plugins with variables.

TimBarham:
When a plugin that uses variables is automatically installed for a new platform, it fails with the following exception:

    Installing "[plugin]" for [platform]
    Failed to install '[plugin]':Error: Variable(s) missing: [variables]
        at .../cordova-lib/src/plugman/install.js:299:23
        at ...

The problem here is that we are looking in `config.xml` for information about the plugin's variables, but the plugin is only saved to `config.xml` if it was added with the `--save` flag. This fix updates the logic to look in `fetch.json` instead.

I believe looking in `fetch.json` is the right thing to do here, regardless, for the following reasons:

* `fetch.json` is the one that is guaranteed to be up-to-date.
* the information we store in `config.xml` is used to restore plugins that have been uninstalled. The information we store in `fetch.json` describes currently installed plugins, which is what we're working with here.

Consider the following scenario:

1. Add a plugin with variables, using the `--save` option. Variable information is saved to `config.xml` and `fetch.json`.
2. Remove the plugin (with the `--save` option). The variable information is removed from `fetch.json`, but remains in `config.xml`.
3. Add the plugin again, with different variable values, without the `--save` option. The current variable information is in `fetch.json`, but not in `config.xml`. Also, each platform currently installed will be using the new variable information.
4. Add a platform. If, when we install the plugin for the new platform, we retrieve the variables from `config.xml`, it will be out of sync with platforms that were already installed.

### DIFF
--- a/cordova-lib/src/cordova/platform.js
+++ b/cordova-lib/src/cordova/platform.js
@@ -162,7 +162,7 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
                     return require('./cordova').raw.prepare(prepOpts);
                 }).then(function() {
                     if (cmd == 'add') {
-                        return installPluginsForNewPlatform(platform, projectRoot, cfg, opts);
+                        return installPluginsForNewPlatform(platform, projectRoot, opts);
                     }
                 }).then(function() {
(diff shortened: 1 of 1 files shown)


kamrik:
LGTM. The scenarios seem convincing. Merging.

PR #128, opened 2014-12-05
CB-8043 CB-6462 CB-6105 Refactor orientation preference support

MMrko:
Fixes [CB-8043](https://issues.apache.org/jira/browse/CB-8043), [CB-6462](https://issues.apache.org/jira/browse/CB-6462) and [CB-6105](https://issues.apache.org/jira/browse/CB-6105). Touches [CB-6182](https://issues.apache.org/jira/browse/CB-6182).

Adds support for the global & platform-specific orientation preference (iOS, WP8 and FirefoxOS). Android and Amazon Fire OS already had this functionality. Additionally, the change will allow specifying a custom value  for the platform-specific orientation preference.  If need be I'd be happy to update the [docs](http://cordova.apache.org/docs/en/4.0.0/config_ref_index.md.html) accordingly.

As a future improvement the orientation preference could offer a similar level of granularity to that of the native APIs. For example, iOS & WP8 APIs allow setting both an initial orientation as well as a set of supported orientations for an app. For now, developers will have to resort to a custom approach in their Cordova builds in case they've such needs.

Tested on OS X Mavericks & Windows 8 running VS 2013.

PR #207, opened 2015-04-13
CB-6462 CB-6026 Orientation preference now updates value for iPad

rachel-carvalho:
As dicussed in #128 (more specifically https://github.com/apache/cordova-lib/pull/128#issuecomment-89316079), currently setting this:
```xml
<preference name="Orientation" value="landscape"/>
```
doesn't affect the iPad.

So I added code and specs for updating `UISupportedInterfaceOrientations~ipad`, as well as `UISupportedInterfaceOrientations`.

This is my first time contributing to Cordova, sorry if something is missing! All input is welcome :)

### DIFF
--- a/cordova-lib/spec-cordova/metadata/ios_parser.spec.js
+++ b/cordova-lib/spec-cordova/metadata/ios_parser.spec.js
@@ -151,6 +151,7 @@ describe('ios project parser', function () {
                 getOrientation.andReturn('');
                 wrapper(p.update_from_config(cfg), done, function() {
                     expect(plist_build.mostRecentCall.args[0].UISupportedInterfaceOrientations).toBeUndefined();
+                    expect(plist_build.mostRecentCall.args[0]['UISupportedInterfaceOrientations~ipad']).toBeUndefined();
                     expect(plist_build.mostRecentCall.args[0].UIInterfaceOrientation).toBeUndefined();
                 });
             });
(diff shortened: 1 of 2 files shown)


mmrko:
:+1:

shazron:
Hi Rachel - can you update the title with a prefix of CB-6462 CB-6026 (to enable some auto-tracking with JIRA)

rachel-carvalho:
Done!

shazron:
CB-8783 affects iPad as well, by inspecting the existing code (not yours).

shazron:
Integrated: https://git1-us-west.apache.org/repos/asf?p=cordova-lib.git;a=commit;h=93ab6b61
You can close this issue now, thanks!

christoph-neumann:
How are we supposed to set iPhone and iPad differently now? My app is portrait only for iPhones, but portrait and landscape for iPads. We need a way to have a different policy for iPhones and iPads.

shazron:
Hmm never did thought of that scenario. Not sure how we would tackle this -- I suggest filing a feature request at issues.cordova.io so we can discuss. 

My suggestion is a separate one, something like "Orientation~ipad". If only "Orientation" is set, it sets it for both. If "Orientation~ipad" is set, it overrides the iPad setting that was set in "Orientation". We won't need one for "Orientation~iphone" I don't think but we might have to for completeness.

christoph-neumann:
@shazron Done. Thanks for looking into it!  https://issues.apache.org/jira/browse/CB-9703

PR #209, opened 2015-04-23
Update elementtree dependency to 0.1.6. Note it has a breaking API chang...

shazron:
...e. https://github.com/racker/node-elementtree/issues/24


shazron:
Great - I'll merge this in tomorrow. The only impact I see is if a plugin require'd elementtree in its hooks and is using the old delete API, but that's not something we can really control.

They'll just have to do something like: https://github.com/apache/cordova-plugins/blob/bef0942b8fab7778caf8bf3770c13fd9a8c6e11c/local-webserver/scripts/before_uninstall.js#L51-L57

PR #212, opened 2015-05-05
CB-8757 - Resolve symlinks in order to avoid relative path issues

mbektchiev:
Android projects residing in symlinked directories have problems similar to those listen in https://issues.apache.org/jira/browse/CB-8757 for iOS. I am proposing a better fix which is platform-agnostic and normalizes the paths as early as possible.


cordova-lib/src/plugman/plugman.js
```diff
@@ -89,9 +90,12 @@ plugman.commands =  {
         plugman.owner(cli_opts.argv.remain);
     },
     'install'  : function(cli_opts) {
+        resolvePathsInArgs(cli_opts);
```
agrieve: CLI code doesn't pass through here either I don't think. Within `install.js` is where you should do it I think.

mbektchiev:
@agrieve I moved the fixes to `install.js` and `uninstall.js`. Do you think it's OK now?

cordova-lib/src/plugman/plugman.js
```diff
@@ -228,3 +228,4 @@ plugman.commands =  {
 };
 
 module.exports = plugman;
+
```
agrieve: looks good. just revert this unchanged file and fix the one failing test.

mbektchiev:
Done. I hadn't noticed the failing test, because it's failing only on Mac, but I hadn't run the tests on a Mac. :four_leaf_clover:

PR #215, opened 2015-05-12
CB-9023 Add support to specify a build config file. If none is specified `bui…

nikhilkh:
…ld.json` in the project root is used as a default


vladimir-kotikov:
LGTM

PR #216, opened 2015-05-13
The concat() method returns a new array.

007slm:
the ios plist file can not add more than one dict CFBundleURLTypes arrays  when use cordova plugman add plugin in ios platform.


shazron:
Can you briefly describe what bug this bug fix solves?

purplecabbage:
+1
'concat does not alter this or any of the arrays ... '

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat

purplecabbage:
This should probably also use __.uniq Instead of the double loop, or __.union

purplecabbage:
This change has already been applied here: dff78b0779444d06e259140694e07d5e8c5f5308
Please close this pr.

PR #223, opened 2015-05-20
CB-9051 Plugins don't get re-added if platforms folder deleted.

TimBarham:
To avoid attempting to add a plugin twice while adding a platform (a scenario that can occur now with automatic restore of plugins and platforms), when adding plugins to a newly added platform we check to see if the plugin has already been added, by calling plugman's `PlatformJson.isPluginInstalled()`, which looks in `plugins/<platform>.json` to see if the plugin has been installed for a platform.

If you delete a platform's folder from the platforms folder, it's `plugins/<platform>.json` file still exists, so if you add the platform again, we think all plugins are already installed for it and don't try to install them.

Simple fix is to delete the platform's `plugins/<platform>.json` file before adding a platform, so we start with a clean slate.

### DIFF
--- a/cordova-lib/src/cordova/platform.js
+++ b/cordova-lib/src/cordova/platform.js
@@ -122,6 +122,12 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
                         if (platformAlreadyAdded) {
                             throw new CordovaError('Platform ' + platform + ' already added.');
                         }
+
+                        // Remove the <platform>.json file from the plugins directory, so we start clean (otherwise we
+                        // can get into trouble not installing plugins if someone deletes the platform folder but
+                        // <platform>.json still exists).
+                        removePlatformPluginsJson(projectRoot, target);
+
                         var template_dir = config_json && config_json.lib && config_json.lib[platform] && config_json.lib[platform].template || null;
                         events.emit('log', 'Adding ' + platform + ' project...');
 
@@ -302,8 +308,7 @@ function remove(hooksRunner, projectRoot, targets, opts) {
     .then(function() {
         targets.forEach(function(target) {
             shell.rm('-rf', path.join(projectRoot, 'platforms', target));
-            var plugins_json = path.join(projectRoot, 'plugins', target + '.json');
-            if (fs.existsSync(plugins_json)) shell.rm(plugins_json);
+            removePlatformPluginsJson(projectRoot, target);
         });
     }).then(function() {
         var config_json = config.read(projectRoot);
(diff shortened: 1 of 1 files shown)


cordova-lib/src/cordova/platform.js
```diff
@@ -122,6 +122,12 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
                         if (platformAlreadyAdded) {
                             throw new CordovaError('Platform ' + platform + ' already added.');
                         }
+
+                        // Remove the <platform>.json file from the plugins directory, so we start clean (otherwise we
+                        // can get into trouble not installing plugins if someone deletes the platform folder but
+                        // <platform>.json still exists).
+                        removePlatformPluginsJson(projectRoot, target);
```
omefire: I think this function call should be moved to https://github.com/MSOpenTech/cordova-lib/blob/CB-9051/cordova-lib/src/cordova/platform.js#L174, because the 'addHelper' function is used both by the 'cordova platform add' and 'cordova platform update'. and it is our intention to only do this file deletion when we are performing an 'add' operation.
Another reason might be that we should be doing this file deletion close to the 'installPluginsForTheNewPlatform' function call, to minimize the risk of getting in an inconsistent state where the file has already been deleted even though the 'add' operation ends up throwing an exception down the line.

TimBarham: Regarding your first point: If you look at the context of where the call was added, it is within an `if (cmd == 'add')` check.

Regarding your second point - no, we still want the file deleted even if the `add` operation fails. Essentially we're just putting things into the state they would have been in if the platform had been properly removed, so we want to get into that state as early as possible.

cordova-lib/src/cordova/platform.js
```diff
@@ -663,6 +668,12 @@ function copy_cordovajs_src(projectRoot, platform, platLib) {
     } 
 }
 
+// Remove <platform>.json file from plugins directory.
+function removePlatformPluginsJson(projectRoot, target) {
+    var plugins_json = path.join(projectRoot, 'plugins', target + '.json');
+    if (fs.existsSync(plugins_json)) shell.rm(plugins_json);
```
nikhilkh: Nit: you can use shell.rm('-f', plugins_json) and then you do not need an exists check.

cordova-lib/src/cordova/platform.js
omefire: Fair points! LGTM!

purplecabbage:
LGTM!

PR #224, opened 2015-05-21
CB-8993 Plugin restore ignores search path

marcuspridham:
Passes provided searchpath to the plugin restore after adding a platform.  Without this I would get errors sometimes about "loadLocalPlugins called twice with different search paths.Support for this is not implemented. "

Also allowing the prepare command to pull the searchpath from the .cordova/config.json if it is present.  This aligns with what the platform and plugin commands do.

### DIFF
--- a/cordova-lib/src/cordova/platform.js
+++ b/cordova-lib/src/cordova/platform.js
@@ -160,7 +160,8 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
                 }).then(function () {
                     // Call prepare for the current platform.
                     var prepOpts = {
-                        platforms :[platform]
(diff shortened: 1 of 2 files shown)


omefire:
Good catch. This change looks good to me !

gorkem:
@marcuspridham PR looks. Do you have Apache CLA signed? I can not seem to find it if you have. Before I can merge this one, you need to sign Apache’s Contributor License Agreement (can be done online): http://www.apache.org/licenses/#clas

marcuspridham:
Thanks.  I filled out a Apache CLA back in January and got confirmation from Apache Secretary Craig Russell.  I have done a pull request in the past for CB-8328.

PR #225, opened 2015-05-22
CB-9065 Allow removing plugins by short name.

TimBarham:
Adds logic to `plugin remove` to look up a plugin by "short name" if we don't find it by id. This means, for example, that you can use the following command:

    cordova plugin remove camera

Instead of:

    cordova plugin remove cordova-plugin-camera

Also removes a bunch of unecessary calls to `path.join(projectRoot, 'plugins')`, since we already have this path stored in a variable.

### DIFF
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -64,9 +64,8 @@ module.exports = function plugin(command, targets, opts) {
     var platformList = cordova_util.listPlatforms(projectRoot);
 
     // Massage plugin name(s) / path(s)
-    var pluginPath, plugins;
-    pluginPath = path.join(projectRoot, 'plugins');
-    plugins = cordova_util.findPlugins(pluginPath);
+    var pluginPath = path.join(projectRoot, 'plugins');
+    var plugins = cordova_util.findPlugins(pluginPath);
     if (!targets || !targets.length) {
         if (command == 'add' || command == 'rm') {
             return Q.reject(new CordovaError('You need to qualify `'+cordova_util.binname+' plugin add` or `'+cordova_util.binname+' plugin remove` with one or more plugins!'));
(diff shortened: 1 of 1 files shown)


purplecabbage:
Interesting, I like it.
How about removing by path/url? 
Does it make sense having installed via `cordova plugin add ../myplugin/` 
that I could then remove it with `cordova plugin rm ../myplugin/` ? 
Not entirely sure this makes sense, or adds value, just throwing it out there.

TimBarham:
Hah, yeah, I think that'd definitely be doable. I won't include it as part of this change, though.

purplecabbage:
So, I discussed this a bit with Steve, and he brought up an interesting point.  
A stretch goal is to not have a plugin.xml file at all, and simply use package.json for the entire plugin definition.  
Given that the `<name>` tag only exists in plugin.xml, could be removed, and is not unique ( which I see you handle ) does it still make sense to push users to rely on a convenience function that we may end up removing?  The name property in a package.json IS the id.
Potentially we could also add a cordova.name field to package.json

TimBarham:
Hmmm... Interesting point. I like the convenience of a short name, but I don't think it is worth the weirdness of adding a custom field to package.json just for that. If I check this in, and people get used to it, they'll expect it to be supported going forward.

So, here's a possible alternatively that will work now and in the future... rather than supporting removing by name, we could support removing by a "short id". That is - you could drop the `cordova-plugin-` from the start. So `camera`, for example, would match `cordova-plugin-camera`. I think that makes sense - having to specify `cordova-plugin-` is kinda superfluous. We should support this shortcut when *adding* a plugin, too.

What do you think?

purplecabbage:
Yeah, I like that better.
If we fail to find it in our typical id search, we then retry with the addition of `cordova-plugin-`
Incidentally, internally here we discussed doing the same thing with install ... adding entries to the plugin-mapper so you could `cordova plugin install camera` and get `cordova-plugin-camera`.  Unfortunately the plugin-mapper is currently only outputting warnings and not redirecting, so that will have to wait.

muratsu:
:+1:

TimBarham:
Ok, updated with that functionality - if we don't find it installed, and we don't find it in the mapper list, then we add `cordova-plugin-` to the start (if it isn't already there) and try again.

PR #226, opened 2015-05-25
Draft implementation for Unified platform API

vladimir-kotikov:
Draft implementation of Unified Platform API for cordova, that inspired by [this thread](http://markmail.org/thread/3dw4mis4qo5d4ecz) in cordova-dev mailing list and [this proposal](https://github.com/kamrik/cordova-discuss/blob/master/proposals/PlatformProject.md)


cordova-lib/src/cordova/metadata/PlatformHandler.js
```diff
@@ -0,0 +1,100 @@
...
+
+    this.platform = platform || '';
+    this.root = this.path = projectPath || '';
+
+    // Extend with a ParserHelper instance
+    Object.defineProperty(this, 'helper', {
+        value: new ParserHelper(this.platform),
+        enumerable: true,
+        configurable: false,
+        writable: false
```
kamrik: The writeable/configurable:false, has the tendency to interfere with jasmine spies which causes our tests to fail with very obscure error messages. Just in case you encounter such errors, otherwise I'm all for leaving this as false.

kamrik:
Looks great! 
Looks like figuring out the tests will be pretty difficult. But it might be worth it just throwing away some of the older tests that rely on a lot of mocks/spies.

cordova-lib/src/cordova/metadata/PlatformHandler.js
```diff
@@ -0,0 +1,100 @@
...
+    return path.resolve(platformSource, 'cordova-js-src');
+};
+
+/**
+ * Base implementation for updateWww.
+ */
+PlatformHandler.prototype.updateWww = function() {
+    var projectRoot = util.isCordova(this.root);
+    var appWww = util.projectWww(projectRoot);
+    var platformWww = path.join(this.root, 'platform_www');
```
kamrik: Looks like we would like to avoid using util.isCordova() here. One can easily create single platform projects that do not maintain the traditional Cordova Project directory structure. Kind of like it was with plugman based workflow or with my latests attempts at the gulp based workflow.

Myabe just pass the appWww as a parameter to this function, and for now as a compatibility feature, get it from util.isCordova() if it's not provided.

cordova-lib/src/cordova/metadata/PlatformHandler.js
vladimir-kotikov: We can always change behavior of `parser` if we'll have any problems later so i would prefer to leave it as-is.

cordova-lib/src/cordova/metadata/PlatformHandler.js
vladimir-kotikov: Updated.

vladimir-kotikov:
Yeah, reworking tests will be painful.
BTW, i'm still wondering that so small number of tests is became failing.

PR #227, opened 2015-05-26
CB-9073 Allow to add platform if project path contains `&` symbol

vladimir-kotikov:
This fixes [CB-9073](https://issues.apache.org/jira/browse/CB-9073)


sgrebnov:
LGTM

PR #228, opened 2015-05-27
CB-8627: Only update fetch.json when plugins are successfully installed.

omefire:
CB-8627: Only update fetch.json when plugins are successfully installed.


omefire:
This change makes sure that the fetch.json file only gets updated with plugin information once the plugin has been successfully installed. fetch.json is considered the source of truth with regards to what plugins are installed.

One particular case this change addresses is that a case of missing variables during `cordova plugin add` (e.g: for plugins like facebook-plugin) will not end up with fetch.json modified even though the plugin addition failed.

omefire:
ping ... @TimBarham

TimBarham:
As I general design point, I don't like the idea of breaking the `save_fetch_metadata()` call out so it needs to be called separately, for two reasons:

1. First and most importantly, `plugman.fetch()` is a public API, and this changes its behavior (it no longer writes to `fetch.json`, and it returns a different value).

2. You've tightened the coupling between `cordova.plugin.add` and `plugman` - `cordova.plugin.add` now needs to know stuff about that result object, and needs to use it to call `save_fetch_metadata()`.

I think a better approach would be for `fetch()` to support an (*optional*) method that it calls before it writes `fetch.json` - this method would return a promise (or reject if something fails). Let's say this parameter was called `validate`, the end of `fetch()` might look something like this (though you'd need to handle the fact that this method would need to be optional):

``` js
        }).then(function(result) {
            options.plugin_src_dir = result.pinfo.dir;
            return Q.when(copyPlugin(result.pinfo, plugins_dir, options.link && result.fetchJsonSource.type == 'local'))
            .then(function(dir) {
                result.dest = dir;
                return result;
            });
        });
    }).then(function(result) {
        return result;
    }).then(validate).then(function (result) {
        var data = { source: result.fetchJsonSource };
        data.is_top_level = options.is_top_level; 
        data.variables = options.variables || {};
        metadata.save_fetch_metadata(plugins_dir, result.pinfo.id, data);
        return result.dest;
    });
}
```
Some additional points:

* In the actual implementation, you should find a cleaner way to pass `result` through to the final piece of code, rather than requiring the `validate()` method return it like my example would. Maybe `validate()` doesn't need to be asynchronous (that is, it could just return a Boolean rather than a promise), but it is certainly more flexible if it supports it being async.
* Since `plugman.fetch()` is responsible for creating the plugin's folder and its contents, it should also be responsible for removing it if the process fails (rather than `plugin.add()` doing it as happens now).

dblotsky:
What's the status on this PR?

dblotsky:
Ping!

cordova-lib/src/cordova/plugin.js
```diff
@@ -206,10 +178,41 @@ module.exports = function plugin(command, targets, opts) {
...
+                            return Q(result);
+                        });
+                    })                    
+                    .then(function(result){
+                        metadata.save_fetch_metadata(pluginsDir, result.pinfo.id, { source: result.fetchJsonSource, variables: opts.cli_variables, is_top_level: true });
+                        return Q(result.dest);
+                    })
+                    .then(function(dir){
+                        // save to config.xml
+                        if(saveToConfigXmlOn(config_json,opts)){
```
robpaveza: Usage: Function name `saveToConfigXmlOn` implies that it performs the saving, but your branch seems to do so.  Perhaps rename to `shouldSaveToConfigXml`.

cordova-lib/src/cordova/plugin.js
```diff
@@ -206,10 +178,41 @@ module.exports = function plugin(command, targets, opts) {
...
+                                events.emit('verbose', 'Calling plugman.install on plugin "' + result.dest + '" for platform "' + platform + '" with options "' + JSON.stringify(options)  + '"');
+                                return plugman.raw.install(platform, platformRoot, path.basename(result.dest), pluginsDir, options);
                             });
-                        }, Q());
+                        }, Q()).then(function(){
+                            return Q(result);
+                        });
+                    })                    
+                    .then(function(result){
+                        metadata.save_fetch_metadata(pluginsDir, result.pinfo.id, { source: result.fetchJsonSource, variables: opts.cli_variables, is_top_level: true });
```
robpaveza: Why not camelCased instead of separated_by_underscore?

cordova-lib/src/plugman/fetch.js
```diff
@@ -146,11 +145,7 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
         });
     }).then(function(result){
         checkID(options.expected_id, result.pinfo);
-        var data = { source: result.fetchJsonSource };
-        data.is_top_level = options.is_top_level; 
-        data.variables = options.variables || {};
-        metadata.save_fetch_metadata(plugins_dir, result.pinfo.id, data);
-        return result.dest;
+        return result;
```
robpaveza: Function name `checkID` seems to do something and return a result, but you don't capture that result, nor do you return anything with that result.  Because you pass the value of `result.pinfo` by value, `checkID` has no opportunity to mutate or modify anything.

What happens if `checkID` "fails", if it can?  If it can't, why is it called `check`?

cordova-lib/src/cordova/plugin.js
```diff
@@ -206,10 +178,41 @@ module.exports = function plugin(command, targets, opts) {
...
+
+                            var attributes = {};
+                            attributes.name = pluginInfo.id;
+
+                            var src = parseSource(target, opts);
+                            attributes.spec = src ? src : '^' + pluginInfo.version;
+
+                            var variables = [];
+                            if (opts.cli_variables) {
+                                for (var varname in opts.cli_variables) {
```
robpaveza: `Object.keys(opts.cli_variables)` rather than `for...in`.

cordova-lib/src/cordova/plugin.js
```diff
@@ -206,10 +178,41 @@ module.exports = function plugin(command, targets, opts) {
...
+                        metadata.save_fetch_metadata(pluginsDir, result.pinfo.id, { source: result.fetchJsonSource, variables: opts.cli_variables, is_top_level: true });
+                        return Q(result.dest);
+                    })
+                    .then(function(dir){
+                        // save to config.xml
+                        if(saveToConfigXmlOn(config_json,opts)){
+                            var pluginInfo =  pluginInfoProvider.get(dir);
+
+                            var attributes = {};
+                            attributes.name = pluginInfo.id;
```
robpaveza: I'm confused by the dichotomy of `name` and `id`.

omefire:
Closing this as not relevant anymore.

PR #229, opened 2015-05-28
CB-9001 Set WMAppManifest.xml Author, Description and Publisher attri…

daserge:
…butes based on config.xml

Used getNodeTextSafe in ConfigParser' description getter

[JIRA issue](https://issues.apache.org/jira/browse/CB-9001)


sgrebnov:
LGTM

PR #230, opened 2015-06-01
Updated Android pinned version for Cordova 4.3.x to 3.7.2 (security patch).

TimBarham:
Also update pinned platforms to use '~' version, so we get patch updates automatically.


nikhilkh:
LGTM!

TimBarham:
I'm going to pull the tilde version change because of the impact on the `--usegit` scenario.

PR #232, opened 2015-06-02
CB-9108 Handle version ranges when add platform with --usegit.

TimBarham:
Adding a platform using a version range didn't work with `--usegit`, since it would try to use that version range as a git tag. This was a problem if you added the platform using `platform add <platform>@<range>`, and also because our pinned platforms now use tilde versions (which are shorthand for a version range).

Fix is to use `npm view` to find the most recent version that matches the range, and use that.

We need to do the same when installing from npm, when we do our check to see if the requested version has previously been downloaded using git.

### DIFF
--- a/cordova-lib/src/cordova/lazy_load.js
+++ b/cordova-lib/src/cordova/lazy_load.js
@@ -29,7 +29,7 @@ var path          = require('path'),
     events        = require('../events'),
     request       = require('request'),
     config        = require('./config'),
-    HooksRunner        = require('../hooks/HooksRunner'),
+    HooksRunner   = require('../hooks/HooksRunner'),
     zlib          = require('zlib'),
     tar           = require('tar'),
     URL           = require('url'),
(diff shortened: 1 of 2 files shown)


TimBarham:
Hey @stevengill ... could you take a look at this? Thanks!

stevengill:
Sure. I'll get to it later today

omefire:
I spent some time reviewing this, it looks good to me even though I think the --usegit flag should be eliminated (some time in the future) now that we have the capability to add platforms using git-urls : `cordova platform add https://github.com/apache/cordova-android.git`

stevengill:
You make a good point @omefire. Lets kick up a discuss thread to depreciate --usegit and if we decide to move forward, we will want to have it around for about 6 months (our depreciation policy) as we give users notice about its removal.

omefire:
@stevengill, starting the discuss thread on the mailing list.

TimBarham:
Ok. I've track down and handled that weird Zlib error you were seeing @stevengill.

TimBarham:
Sounds good regarding the discuss thread, @omefire and @stevengill. In the mean time, I've pushed this so we can proceed with the tools release.

stevengill:
Great! I will proceed with the release!

PR #233, opened 2015-06-03
Fix four failing tests on Windows.

TimBarham:
Two tests fail on Windows in a non-admin command window because they rely on creating symlinks. Since creating the symlinks isn't the purpose of the test but just setup for the test, if that fails and the platform is `win32`, we now just bail on the test.

Another test failed because a folder unexpectedly exists when run from a non-admin prompt. I'm not 100% sure why the folder exists, but the test was bogus... the test actually ends up testing that we throw if `src` does not exist, which is something that we verify in an earlier test, while the intent of the test was to verify we throw if `dest` *already* exists. Since it is `copyNewFile()` that is supposed to throw in that scenario, so I updated the test to check for that.

Finally, a fourth test failed on Windows (admin or non-admin) because of different line endings. Changed that to handle either line ending.

Verified all tests now pass on Windows and still pass on OS X.

Note: I don't believe this is the cause of the AppVeyor failures we are seeing.


purplecabbage:
Awesome, I was just looking into this, waiting for appveyor now ...

TimBarham:
Bummer... AppVeyor is still broken (which I pretty much expected since its failure is different from what I was seeing - but it would have been nice :smile:).

purplecabbage:
Yeah, gotta love the 'weird error' when it happens.  This does resolve the other errors I had.  I think we can still pull this in.

PR #235, opened 2015-06-04
Implementation for Unified platform API 

vladimir-kotikov:
An mplementation of Unified Platform API for cordova, that inspired by [this thread](http://markmail.org/thread/3dw4mis4qo5d4ecz) in cordova-dev mailing list and [this proposal](https://github.com/kamrik/cordova-discuss/blob/master/proposals/PlatformProject.md)

This PR is the updated version of https://github.com/apache/cordova-lib/pull/226

### DIFF
--- a/cordova-lib/src/cordova/prepare.js
+++ b/cordova-lib/src/cordova/prepare.js
@@ -18,12 +18,8 @@
 */
 
 var cordova_util      = require('./util'),
-    ConfigParser      = require('../configparser/ConfigParser'),
     path              = require('path'),
     platforms         = require('../platforms/platforms'),
-    fs                = require('fs'),
-    shell             = require('shelljs'),
-    et                = require('elementtree'),
     HooksRunner       = require('../hooks/HooksRunner'),
     events            = require('../events'),
     Q                 = require('q'),
@@ -39,7 +35,8 @@ var PluginInfoProvider = require('../PluginInfoProvider');
 exports = module.exports = prepare;
 function prepare(options) {
     var projectRoot = cordova_util.cdProjectRoot();
-    var xml = cordova_util.projectConfig(projectRoot);
+    var projectWww = cordova_util.projectWww(projectRoot);
+    var projectConfig = cordova_util.projectConfig(projectRoot);
 
     if (!options) {
         options = {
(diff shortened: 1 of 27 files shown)


cordova-lib/src/platforms/PlatformApi.js
```diff
@@ -0,0 +1,365 @@
...
+ * Base implementation for updateProject. Does nothing
+ * since implementation is heavily depends on platform specifics.
+ * Always should be overridden by platform.
+ */
+BasePlatformApi.prototype.updateProject = function(configSource) {
+    if (this.parser && this.parser.update_project) {
+        var configParser = new ConfigParser(configSource || this.getConfigXml());
+        return Q.when(this.parser.update_project(configParser));
+    }
+    return Q();
```
kamrik: Should we throw in this case? If there is no parser and no overriding implementation, ti's probably a bug.

kamrik:
Looks great! The need to duplicate common code like ConfgiParser in cordova-android is a bit annoying and is bound to generate some versioning confusion. But probably it should be dealt with later.

One option could be to pass modules like ConfigParser to coraodva-android/Api.js from code that requires Api.js (usually cordova-lib). It's weird and unintuitive, but would solve the code duplication project.

cordova-lib/src/platforms/PlatformApi.js
vladimir-kotikov: Well, in some cases platform may want to do nothing in `updateProject`. Example is a browser platform, which needs to update config.xml  and www content only during prepare.

vladimir-kotikov:
@kamrik, thanks for feedback.

Dependency injection of cordova logic to platform was our very first idea, but in addition to limitations, you mentioned, it seems that it also will be impossible to use PlatformAPI without CLI in this case.

Another approach, we've been thinking about - factoring out common classes and utilities, like ConfigHelper, superspawn, Requirements-related stuff, etc., to separate module, which then can be used by platforms and cordova-lib. This approach is quite similar to @TimBarham's PR for cordova-serve: https://github.com/apache/cordova-lib/pull/238

What do you think about this?

kamrik:
I don't have strong opinion here, but a bit more inclined towards dependency injection. For some modules like ConfigParser having different versions imported by different platforms can get very frustrating. Even when cordova is used without thc CLI, large parts of cordova-lib would still be needed, whether required()'d directly by the user or implicitly via cordova-android. So as an extreme option we can dependency-inject the entire cordova-lib (properly exposing the relevant classes). Somewhat more elegantly, we could construct a module inside cordova-lib with just the needed classes and inject it when instantiating cordova-android API. It might be useful to later split out this module as a separate npm package (or several packages), but for time being they could still stay part of the lib.

Superspawn is definitely a candidate for being a totally separate module. But modules like PluginInfo and ConfigParser are very unlikely to be used separately from each other. Maybe in the long run they should live in some new package named like "cordova-core", or we could move the CLI specific stuff out of cordova-lib and into cordova-cli, then the lib would become that core.

That said, I'm fine with the approach you suggested.

cordova-lib/src/platforms/PlatformApi.js
```diff
@@ -0,0 +1,365 @@
...
+ * @return {Promise}         Promise, either resolved with array of requirements
+ *                                    or rejected with error.
+ */
+BasePlatformApi.prototype.requirements = function(options) {
+    var modulePath = path.join(this.root, 'cordova', 'lib', 'check_reqs');
+    return require(modulePath).check_all();
+};
+
+// getPlatformApi() should be the only method of instantiating the
+// PlatformProject classes for now.
```
omefire: :nit PlatformProject -> PlatformApi ?

cordova-lib/src/platforms/PlatformApi.js
```diff
@@ -0,0 +1,365 @@
...
+    platforms = require('./platformsConfig.json'),
+    ConfigParser = require('../configparser/ConfigParser');
+
+var BasePluginHandler = require('../plugman/platforms/PluginHandler');
+var ParserHelper = require('../cordova/metadata/parserhelper/ParserHelper');
+
+// Avoid loading the same platform projects more than once (identified by path)
+var cachedProjects = {};
+
+// A single class that exposes functionality from platform specific files from
```
omefire: does this comment still apply ? should we remove it ?

nikhilkh:
It will help to generate API docs for the public API surface now. Reviewing the implementation guts and the API in code is not easy. In any case, the platform API might need to be documented for anyone looking to create a platform. Might as well do it now.

For that, using jsdoc or other tools to generate docs for the API from code would be super useful.

nikhilkh:
Overall, it feels like the API is closely bound to current implementation and order as opposed to be more meaningful and self-descriptive. I know there is a constraint of a lot of legacy baggage to support here.

cordova-lib/src/platforms/PlatformApi.js
```diff
@@ -0,0 +1,365 @@
...
+        }
+    }
+
+    return ctor;
+}
+
+function runShellSilent(callback) {
+    var silentstate = shell.config.silent;
+    shell.config.silent = true;
+    callback();
```
nikhilkh: Should there be a try..finally here?

cordova-lib/src/platforms/PlatformApi.js
```diff
@@ -0,0 +1,365 @@
...
+/**
+ * Default implementation for platform run. Uses executable script shipped with platform
+ * to deploy and run project. Could be overridden using PlatformApi implementation if it
+ * is provided by platform.
+ * @param  {Object}  options Complex object that provides cordova API to method
+ * @return {Promise}         Promise, either resolved or rejected with error code.
+ */
+BasePlatformApi.prototype.run = function(options) {
+    var cmd = path.join(this.root, 'cordova', 'run');
+    return superspawn.spawn(cmd, options.options, { printCommand: options.verbose, stdio: 'inherit' });
```
nikhilkh: Looks like you need `chmod:true`

cordova-lib/src/platforms/PlatformApi.js
vladimir-kotikov: Updated in https://github.com/MSOpenTech/cordova-lib/commit/c78eae57929ed541ac2c2a47a164938433133309

omefire:
Since you've already ported the android implementation from cordova-lib to cordova-android, shouldn't you delete the android parser file from the metadata directory ? 

As long as it exists, BasePlatformApi.update_www (and potentially other functions) refers to it instead of the intended new code.

vladimir-kotikov:
`android_parser` is left for backward compatibility, since user can install older version of cordova-android, which doesn't provides own `PlatformAPI`. In this case the code from `android_parser` will be used.

If you're instantiating `PlatformApi` properly - via `getPlatformApi` getter, and not directly (`new BasePlatformApi()`) - the `updateWww` will _always_ be owerwritten by platform-provided method. (of course if platform provides it :smiley:  )

cordova-lib/src/platforms/PlatformApi.js
```diff
@@ -0,0 +1,414 @@
...
+
+/**
+ * Base implementation for updateConfig. Reset platform's config to default.
+ * @return {String} Path to platform's config.xml file
+ */
+BasePlatformApi.prototype.updateConfig = function(optSource) {
+    var defaultConfig = path.join(this.root, 'cordova', 'defaults.xml');
+    var ownConfig = this.getConfigXml();
+    // If defaults.xml is present, overwrite platform config.xml with it
+    if (fs.existsSync(defaultConfig)) {
```
omefire: The old version of the code also handled the case when both defaults.xml and project_dir/config.xml were not present.
In that case, it would grab 'www/config.xml' and copy it to project_dir/config.xml.

Any reason that case isn't handled anymore ?

omefire:
cool. (ref: android_parser & backward compatibility).

cordova-lib/src/platforms/PlatformApi.js
omefire: The old version of the code also handled the case when both defaults.xml and project_dir/config.xml were not present.
In that case, it would grab 'www/config.xml' and copy it to project_dir/config.xml.

Any reason that case isn't handled anymore ?

cordova-lib/src/cordova/prepare.js
```diff
@@ -56,147 +53,44 @@ function prepare(options) {
...
-                    shell.cp('-f', xml, parser.config_xml());
+            var platformApi = platforms.getPlatformApi(platform, platformPath);
+
+            return platformApi.updateWww(projectWww)
+            .then(function () {
+                return platformApi.updateConfig(projectConfig);
+            }).then(function () {
+                // Call plugman --prepare for this platform. sets up js-modules appropriately.
+                var plugins_dir = path.join(projectRoot, 'plugins');
+                var platformPath = path.join(projectRoot, 'platforms', platform);
```
omefire: platformPath has already been defined.

omefire:
Ah, I see. (ref: android_parser & backward compatibility).

However, in case a platform is implemented via the new way, shouldn't we be actually using that ? currently, we're still using the old code, which means the new one never runs.

On top of that, if someone wanted to fix or debug an issue, he'd do it on the new code instead of the old one, which is confusing.

I suggest, we always prefer the new API when available, and only fall back to old one when the new API isn't available.

cordova-lib/src/platforms/PlatformApi.js
```diff
@@ -0,0 +1,414 @@
...
+            if (parser)return parser;
+
+            var ParserConstructor;
+            try {
+                ParserConstructor = require(platforms[this.platform].parser_file);
+            } catch (e) { }
+
+            // parser === null is the special case which means that we've tried
+            // to get embedded platform parser and failed. In this case instead of
+            // parser's methods will be called PlatformApi default implementations.
```
omefire: this sentence is a little confusing. 
Maybe reword it to 'In this case, the default behavior as implemented by PlatformApi will be used' ?

vladimir-kotikov:
Closing this since the PlatformApi needs for more detailed design. The discussion is moved to cordova/cordova-discuss#9

PR #236, opened 2015-06-05
Change cordova-lib to execute plugin hooks without the need of Cordova project structure

galexandrov:
Change cordova-lib to execute plugin hooks defined in plugin.xml without the need of Cordova project structure and extend Plugman with extra command argument (--nohooks) for preventing plugin hooks execution. By default the hooks will be executed.


omefire:
What's the reason behind this proposed change ?  What use case does this fulfill ?

galexandrov:
In our case we don't use the Cordova CLI we use cordova-{platform} scripts for project creation and that's why we don't have the Cordova CLI Project structure. But it seems that the hooks execution is tightly bound with the Cordova Project structure and more specific with the existence of cordova folder. 
The condition for hook execution in plugmna was cordovaUtil.isCordova(), which means that when we use cordova-{platform} scripts for project creation and cordova-plugman for plugin management the hooks will not be executed.

galexandrov:
ping @omefire

omefire:
@galexandrov , will take a look when possible...

cordova-lib/src/hooks/HooksRunner.js
```diff
@@ -60,11 +64,23 @@ HooksRunner.prototype.fire = function fire(hook, opts) {
 };
 
 /**
+ * Tries to create a CordovaHooksRunner for passed project root.
+ * @constructor
+ */
+function CordovaHooksRunner(projectRoot) {
+    DefaultHooksRunner.call(this, projectRoot);
+    var root = cordovaUtil.isCordova(projectRoot);
+    if (!root) throw new CordovaError('Not a Cordova project ("' + projectRoot + '"), can\'t use hooks.');
```
omefire: :nit please use `if (...) { } else { }` (with curly braces).

cordova-lib/src/hooks/HooksRunner.js
```diff
@@ -80,6 +96,15 @@ HooksRunner.prototype.prepareOptions = function(opts) {
     return opts;
 };
 
+function HooksRunner(projectRoot) {
```
omefire: Could you please provide the 'why' we are moving to have both CordovaHooksRunner and DefaultHooksRunner ? So that other maintainers are not confused as to the reason behind this shift.

cordova-lib/src/hooks/scriptsFinder.js
```diff
@@ -105,6 +105,11 @@ function getApplicationHookScriptsFromDir(dir) {
  */
 function getScriptsFromConfigXml(hook, opts) {
     var configPath = cordovaUtil.projectConfig(opts.projectRoot);
+
+    if (!(fs.existsSync(configPath))) {
```
omefire: a comment here that lets code readers know that there are cases when the cordova project structure is not available would be helpful.

cordova-lib/src/plugman/install.js
```diff
@@ -357,7 +356,7 @@ function runInstall(actions, platform, project_dir, plugin_dir, plugins_dir, opt
                     }
                 };
 
-                var hooksRunner = new HooksRunner(projectRoot);
+                var hooksRunner = new HooksRunner(cordovaUtil.isCordova() || project_dir);
```
omefire: why `cordovaUtil.isCordova() || project_dir` ?
wouldn't just using `project_dir` achieve the same result ?

cordova-lib/src/plugman/uninstall.js
```diff
@@ -283,7 +283,7 @@ function runUninstallPlatform(actions, platform, project_dir, plugin_dir, plugin
             }
         };
 
-        var hooksRunner = new HooksRunner(projectRoot);
+        var hooksRunner = new HooksRunner(cordovaUtil.isCordova() || project_dir);
```
omefire: why cordovaUtil.isCordova() || project_dir ?
wouldn't just using project_dir achieve the same result ?

cordova-lib/spec-cordova/HooksRunner.spec.js
```diff
@@ -81,12 +81,6 @@ describe('HooksRunner', function() {
         process.chdir(path.join(__dirname, '..'));  // Non e2e tests assume CWD is repo root.
     });
 
-    it('should throw if provided directory is not a cordova project', function() {
```
omefire: Please add a test case that deals with a NON-cordova project.

omefire:
I feel like the plugman changes ( `run_hooks` ) should be part of a separate pull request.

cordova-lib/src/plugman/uninstall.js
stevengill: Sending project_dir should be enough

stevengill:
I'd like to revive this PR. Hooks should run when using create scripts + plugman.

PR #237, opened 2015-06-05
CB-9126 Fix ios pbxproj' resources paths when adding ios platform on non-OSX environment

daserge:
[Jira issue with details](https://issues.apache.org/jira/browse/CB-9126)

### DIFF
--- a/cordova-lib/package.json
+++ b/cordova-lib/package.json
@@ -42,7 +42,7 @@
     "underscore": "1.7.0",
     "unorm": "1.3.3",
     "valid-identifier": "0.0.1",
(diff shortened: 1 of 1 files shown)


nikhilkh:
LGTM. Good find!

vote539:
Linking in [another JIRA issue](https://issues.apache.org/jira/browse/CB-5898).  This bug was first observed in January '14, and was diagnosed and solved upstream in node-xcode in April '15.  Thanks for catching that the newer version of node-xcode wasn't being pulled into CordovaLib.

PR #238, opened 2015-06-05
CB-9127 Implements cordova-serve module.

TimBarham:
This takes code and functionality from `cordova-browser` and `cordova-lib` (`cordova serve` command) and combines it in a single module. There's a lot of code here, but by far the bulk of it comes directly from those two places. See readme.md for more information.

Once this is checked in, I will:

* Update COHO release tools to work with this module.
* Work to get a `cordova-serve` release out.
* Update `cordova-lib` and `cordova-browser` to use `cordova-serve`.

### DIFF
--- a/cordova-serve/src/browser.js
+++ b/cordova-serve/src/browser.js
@@ -1,0 +1,100 @@
+/**
+ Licensed to the Apache Software Foundation (ASF) under one
+ or more contributor license agreements.  See the NOTICE file
+ distributed with this work for additional information
+ regarding copyright ownership.  The ASF licenses this file
+ to you under the Apache License, Version 2.0 (the
+ "License"); you may not use this file except in compliance
+ with the License.  You may obtain a copy of the License at
+
+ http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+ */
+
+var exec = require('./exec'),
+    Q = require('q');
+
+/**
+ * Launches the specified browser with the given URL.
+ * Based on https://github.com/domenic/opener
+ * @param {{target: ?string, url: ?string, dataDir: ?string}} opts - parameters:
+ *   target - the target browser - ie, chrome, safari, opera, firefox or chromium
+ *   url - the url to open in the browser
+ *   dataDir - a data dir to provide to Chrome (can be used to force it to open in a new window)
+ * @return {Q} Promise to launch the specified browser
+ */
+module.exports = function (opts) {
+    //target, url, dataDir
+    var target = opts.target || 'chrome';
+    var url = opts.url || '';
+
+    return getBrowser(target, opts.dataDir).then(function (browser) {
+        var args;
+        switch (process.platform) {
+            case 'darwin':
+                args = ['open'];
+                if (target == 'chrome') {
+                    // Chrome needs to be launched in a new window. Other browsers, particularly, opera does not work with this.        
+                    args.push('-n');
+                }
+                args.push('-a', browser);
+                break;
+            case 'win32':
+                // On Windows, we really want to use the "start" command. But, the rules regarding arguments with spaces, and 
+                // escaping them with quotes, can get really arcane. So the easiest way to deal with this is to pass off the 
+                // responsibility to "cmd /c", which has that logic built in. 
+                // 
+                // Furthermore, if "cmd /c" double-quoted the first parameter, then "start" will interpret it as a window title, 
+                // so we need to add a dummy empty-string window title: http://stackoverflow.com/a/154090/3191 
+                args = ['cmd /c start ""', browser];
+                break;
+            case 'linux':
+                // if a browser is specified, launch it with the url as argument
+                // otherwise, use xdg-open.
+                args = [browser];
+                break;
+        }
+        args.push(url);
+        var command = args.join(' ');
+        console.log('Executing command: ' + command);
+        return exec(command);
+    });
+};
+
+function getBrowser(target, dataDir) {
+    dataDir = dataDir || 'temp_chrome_user_data_dir_for_cordova';
+
+    var chromeArgs = ' --user-data-dir=/tmp/' + dataDir;
+    var browsers = {
+        'win32': {
+            'ie': 'iexplore',
+            'chrome': 'chrome --user-data-dir=%TEMP%\\' + dataDir,
+            'safari': 'safari',
+            'opera': 'opera',
+            'firefox': 'firefox'
+        },
+        'darwin': {
+            'chrome': '"Google Chrome" --args' + chromeArgs,
+            'safari': 'safari',
+            'firefox': 'firefox',
+            'opera': 'opera'
+        },
+        'linux' : {
+            'chrome': 'google-chrome' + chromeArgs ,
+            'chromium': 'chromium-browser' + chromeArgs,
+            'firefox': 'firefox',
+            'opera': 'opera'
+        }
+    };
+    target = target.toLowerCase();
+    if (target in browsers[process.platform]) {
+        return Q(browsers[process.platform][target]);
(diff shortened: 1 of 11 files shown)


cordova-serve/src/server.js
```diff
@@ -0,0 +1,137 @@
...
+            response.end();
+        }
+
+        function do304() {
+            console.log('304 ' + request.url);
+            response.writeHead(304, {'Content-Type': 'text/plain'});
+            response.end();
+        }
+
+        function isFileChanged(path) {
```
vladimir-kotikov: IMO this is a good candidate to factor out to `./util` module with signature `function isFileChanged(path, modifiedSince) {}`

cordova-serve/src/server.js
```diff
@@ -0,0 +1,137 @@
...
+                        response.write('<li><a href="' + file + '">' + file + '</a></li>\n');
+                    }
+                });
+
+                response.write('</ul>');
+                response.end();
+            } else if (!isFileChanged(filePath)) {
+                do304();
+            } else {
+                (opts.streamHandler && opts.streamHandler(filePath, request, response)) || stream(filePath, request, response);
```
vladimir-kotikov: nit: This is more readable and do not require `-W030` for JSHint:
```
var streamHandler = opts.streamHandler || stream;
streamHandler(...params...)
```

cordova-serve/src/server.js
```diff
@@ -0,0 +1,137 @@
...
+                    do302(request.url + '/');
+                    return;
+                }
+                console.log('200 ' + request.url);
+                response.writeHead(200, {'Content-Type': 'text/html'});
+                response.write('<html><head><title>Directory listing of ' + urlPath + '</title></head>');
+                response.write('<h3>Items in this directory</h3>');
+                response.write('<ul>');
+                fs.readdirSync(filePath).forEach(function (file) {
+                    if (file) {
```
vladimir-kotikov: This is always `true`.

cordova-serve/src/server.js
```diff
@@ -0,0 +1,137 @@
...
+
+        fs.exists(filePath, function (exists) {
+            if (!exists) {
+                do404();
+                return;
+            }
+            if (fs.statSync(filePath).isDirectory()) {
+                var index = path.join(filePath, 'index.html');
+                try {
+                    if (fs.statSync(index)) {
```
vladimir-kotikov: Why `try...catch` instead of just `fs.existSync`?

cordova-serve/src/server.js
```diff
@@ -0,0 +1,137 @@
...
+            filePath = result.filePath;
+        }
+        if (!filePath) {
+            if (!root) {
+                throw new Error('No server root directory HAS BEEN specified!');
+            }
+            filePath = path.join(root, urlPath);
+        }
+
+        fs.exists(filePath, function (exists) {
```
vladimir-kotikov: The logic below will look better if factored into `defaultUrlProcessor`, so you can do something like
```javascript
var urlProcessor = opts.urlPathProcessor || defaultUrlProcessor;
var result = urlProcessor(urlPath, request, response, do302, do404);
```
in [line 69](https://github.com/apache/cordova-lib/pull/238/files#diff-28759fa7c601ddb3ffab33e34a6587d2R69)

cordova-serve/src/platform.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Launches a server where the root points to the specified platform in a Cordova project.
+ * @param {string} platform - Cordova platform to serve.
+ * @param {{root: ?string, port: ?number, urlPathProcessor: ?function, streamHandler: ?function, serverExtender: ?function}} opts
+ *   root - cordova project directory, or any directory within it. If not specified, cwd is used. This will be modified to point to the platform's www_dir.
+ *   All other values are passed unaltered to launchServer().
+ * @returns {*|promise}
+ */
+module.exports = function (platform, opts) {
+    if (!platform) {
```
vladimir-kotikov: It would be better to wrap function into `Q.when()` to return a promise even if error is thrown. Otherwise you'll need to wrap this code into `try...catch` in caller code.

cordova-serve/src/platform.js
```diff
@@ -0,0 +1,55 @@
...
+        throw new Error('A platform must be specified');
+    }
+
+    opts.root = util.getPlatformWwwRoot(findProjectRoot(opts.root), platform);
```
vladimir-kotikov: `opts` can be `undefined`, so accessing `opts.root` will throw.

cordova-serve/src/server.js
```diff
@@ -0,0 +1,137 @@
...
+var fs     = require('fs'),
+    http   = require('http'),
+    url    = require('url'),
+    path   = require('path'),
+    Q      = require('q'),
+    stream = require('./stream');
+
+/**
+ * @desc Launches a server with the specified options and optional custom handlers.
+ * @param {{root: ?string, port: ?number, urlPathProcessor: ?function, streamHandler: ?function, serverExtender: ?function}} opts
```
vladimir-kotikov: IMO `urlPathProcessor`, `serverExtender` and `streamHandler` deserves additional description.

cordova-serve/src/browser.js
```diff
@@ -0,0 +1,100 @@
...
+ *   url - the url to open in the browser
+ *   dataDir - a data dir to provide to Chrome (can be used to force it to open in a new window)
+ * @return {Q} Promise to launch the specified browser
+ */
+module.exports = function (opts) {
+    //target, url, dataDir
+    var target = opts.target || 'chrome';
+    var url = opts.url || '';
+
+    return mapTarget(target, opts.dataDir).then(function (browser) {
```
vladimir-kotikov: nit: it might sounds better if the function is called `getBrowser`

cordova-serve/src/server.js
TimBarham: Hah, fair point. Though I'm tempted to make it:

    (opts.streamHandler || stream)(filePath, request, response);

Jut for kicks :smile:

cordova-serve/src/server.js
TimBarham: Yeah, this was just copied and pasted code, but I'll fix it.

cordova-serve/src/server.js
TimBarham: This was just copied and pasted code (from cordova-lib `serve` module), so not sure why :smile: ... Perhaps existsSync wasn't around at the time? I also thought I had fixed this in one of my iterations on this stuff. Guess not. Will fix it now.

cordova-serve/src/server.js
TimBarham: Yeah, good suggestion. It's not quite as simple as that, since a custom `urlPathProcessor()` might be provided but not want to process a particular path and fall back on default logic. I'll take a look, though, because I'm not happy with how this works currently (the logic of what `urlPathProcessor()` returns is too weird).

cordova-serve/src/server.js
TimBarham: Yeah, possibly. I might skip this one, however (this is just copied and pasted existing code, and I don't think this one is high enough priority).

cordova-serve/src/platform.js
TimBarham: Done.

cordova-serve/src/platform.js
TimBarham: Fixed.

cordova-serve/src/server.js
TimBarham: Done.

cordova-serve/src/browser.js
TimBarham: Heh, sure - this was also just copied and pasted code (from cordova-browser), but sounds reasonable.

cordova-serve/src/server.js
TimBarham: So, I've done this slightly differently, but it ends up cleaning up the logic here quite a bit, and also in the implementations I have for `urlPathProcessor()`.

Max104t: is `urlPathProcessor` still used today? the scenario I have is to have a local proxy to short-circuit the CORS failures I get when I use XHR from browser.. I was hoping to use it as a redirect to local proxy.. thanks

PR #239, opened 2015-06-06
CB-8052: [webOS] Update cordova-lib for webOS support

JayCanuck:
Allows for Cordova CLI and Plugman to work with webOS in the Cordova project environment.

Naturally it relies on https://github.com/apache/cordova-webos/pull/3 having been merged and added to NPM to work via the webos shorthand.

### DIFF
--- a/cordova-lib/src/plugman/platforms/webos.js
+++ b/cordova-lib/src/plugman/platforms/webos.js
@@ -55,12 +55,12 @@ module.exports = {
         return widget_doc._root.attrib['id'];
     },
     'source-file':{
-        install:function(source_el, plugin_dir, project_dir, plugin_id) {
-            var dest = path.join(source_el.attrib['target-dir'], path.basename(source_el.attrib['src']));
-            common.copyFile(plugin_dir, source_el.attrib['src'], project_dir, dest);
+        install:function(obj, plugin_dir, project_dir, plugin_id, options) {
+            var dest = path.join(obj.targetDir, path.basename(obj.src));
+            common.copyFile(plugin_dir, obj.src, project_dir, dest);
         },
-        uninstall:function(source_el, project_dir, plugin_id) {
-            var dest = path.join(source_el.attrib['target-dir'], path.basename(source_el.attrib['src']));
+        uninstall:function(obj, project_dir, plugin_id, options) {
+            var dest = path.join(obj.targetDir, path.basename(obj.src));
             common.removeFile(project_dir, dest);
         }
     },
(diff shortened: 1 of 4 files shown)


cordova-lib/src/plugman/platforms/webos.js
```diff
@@ -0,0 +1,99 @@
...
+        } else {
+            config_path = preferred_path;
+        }
+        var widget_doc = xml_helpers.parseElementtreeSync(config_path);
+        return widget_doc._root.attrib['id'];
+    },
+    'source-file':{
+        install:function(obj, plugin_dir, project_dir, plugin_id, options) {
+            var dest = path.join(obj.targetDir, path.basename(obj.src));
+            common.copyFile(plugin_dir, obj.src, project_dir, dest);
```
omefire: Every other platform uses 'copyNewFile' to install 'source-file'. Maybe we should do the same here ?

JayCanuck: I modeled it after firefoxos's extry, given the similar OS nature and they had the copyFile entry. If copyNewFile is the current stand, I can update it to that tomorrow morning no problem

JayCanuck: Looks like browser platform uses it as well. Is there some advantage to copyNewFile that I'm missing that would make the change worth it?

stevengill:
Merged!

cordova-lib/src/plugman/platforms/webos.js
omefire: To be honest, I don't know. I've been looking through history to understand why, but I haven't yet. There's a bug I can easily fix elsewhere if I have the answer to these questions :
 - why copyFile vs copyNewFile ? 
 - when was 'copyNewFile' included and what problem did it fix ? 
 - is it necessary ?

purplecabbage: The only difference in that copyNewFile throws an error if the target exists.
common.copyNewFile checks if the file exists, then actually calls common.copyFile

omefire: Thanks @purplecabbage, I see that. 
What I don't know/understand is what would happen/break if we changed it back to 'copyFile' ?
Because this ends up failing in cases two plugins have the same dependencies.

purplecabbage: It was added here: 
https://github.com/apache/cordova-lib/commit/ae2ce7ac6325d1c6558365b4be82e5a8ff0983b9#diff-bf0ba2178fefa5505235315252ccc98c

Not sure why it is required, other than the somewhat meaningless 'refactor'  commit message.

PR #242, opened 2015-06-09
fix CB-9145 prepare can lose data during config munge

tony--:
I added a test that adds access tags as described in CB-9145.  This test fails when run against master.
This is because the current exact match detection assumes that the query returning a result is sufficient to identify an exact attribute match.  However, the query may also return results with additional attributes and these should be excluded.  

Originally I added a utility function that handles this case by performing a 2-way comparison.  However, we already know that foundChild.attr matches srcChild.attr because foundChild is a query result.  We only need to determine if foundChild has extra attributes, so simply comparing the number of attributes is sufficient.  

After making this change, all tests are green - including the newly added test that fails against master.

### DIFF
--- a/cordova-lib/spec-cordova/prepare.spec.js
+++ b/cordova-lib/spec-cordova/prepare.spec.js
@@ -271,4 +271,25 @@ describe('prepare._mergeXml', function () {
         var testElements = dstXml.findall('preference');
         expect(testElements.length).toEqual(2);
     });
+
+    it('should not skip partial duplicate non singelton children', function () {
+        //remove access tags from dstXML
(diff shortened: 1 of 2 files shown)


purplecabbage:
lgtm :+1:

tony--:
Thanks @purplecabbage!

tony--:
AND I screwed up my first commit comment by prepending the word "fix" before the JIRA issue.  I will adhere to the formatting guidelines in the future.

tony--:
Merged commit [2598231](https://github.com/tony--/cordova-lib/commit/25982317b26c831dbea9d6b71dd57a90b933fc68) into apache:master, closing.

PR #243, opened 2015-06-09
Adds license headers to files in templates folder.

TimBarham:
Also, updates code that uses the template files when building plugins to strip the license so its not included in users' projects (note we don't need to strip the license from the XML files as we immediately generate an element tree with `ElementTree`).


nikhilkh:
LGTM

PR #244, opened 2015-06-09
CB-9147 Adding a platform via caret version adds wrong version

TimBarham:
It adds the very latest version rather than the latest matching version.

Internally we are using a version of `npm` that doesn't understand caret versions, so it just gets the latest. I missed this previously because when I tried this, the latest version **was also** the latest matching version.

The same problem also applies to plugins.

I've made a slightly different fix in each case:

For platforms, we already had the latest matching version as a result of our call to `getLatestMatchingNpmVersion()`, so I just use that.

For plugins, in this specific scenario (a caret version) I turn it into a standard range using a version of `semver` that understands caret ranges (so `^1.2.3` becomes `>=1.2.3 <2.0.0`) and use that instead.

Note that this is currently an issue with platforms and plugin versions saved to `config.xml`.

Also, I noticed that we failed to find the version in `config.xml` if you added it using the new form (`cordova-plugin-camera`, for example), but it was stored in `config.xml` using the id, so fixed that.

### DIFF
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -360,6 +360,14 @@ function getPluginVariables(variables){
 
 function getVersionFromConfigFile(plugin, cfg){
     var pluginEntry = cfg.getPlugin(plugin);
+    if (!pluginEntry) {
+        // If the provided plugin id is in the new format (e.g. cordova-plugin-camera), it might be stored in config.xml
+        // under the old format (e.g. org.apache.cordova.camera), so check for that.
+        var oldStylePluginId = pluginMapper[plugin];
+        if (oldStylePluginId) {
+            pluginEntry = cfg.getPlugin(oldStylePluginId);
+        }
+    }
     return pluginEntry && pluginEntry.spec;
 }
 
--- a/cordova-lib/src/plugman/registry/registry.js
+++ b/cordova-lib/src/plugman/registry/registry.js
@@ -349,7 +349,7 @@ function fetchPlugin(plugin, client, useNpmRegistry) {
 
     return initThenLoadSettingsWithRestore(useNpmRegistry, function () {
         events.emit('log', 'Fetching plugin "' + plugin + '" via ' + registryName);
-        return Q.ninvoke(npm.commands, 'cache', ['add', plugin])
+        return Q.ninvoke(npm.commands, 'cache', ['add', processPluginVersion(plugin)])
         .then(function (info) {
             var cl = (client === 'plugman' ? 'plugman' : 'cordova-cli');
             bumpCounter(info, cl);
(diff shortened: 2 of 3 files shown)


cordova-lib/src/plugman/registry/registry.js
```diff
@@ -364,6 +364,28 @@ function fetchPlugin(plugin, client, useNpmRegistry) {
...
+    if (!version || version.charAt(0) !== '^') {
+        return plugin;
+    }
+
+    var validRange = semver.validRange(version, /* loose */ true);
+    if (!validRange) {
+        return plugin;
+    }
+
+    console.log('Returning processed plugin range: ' + parts[0] + '@"' + validRange + '"');
```
stevengill: I would change console.log to events.emit('log', MSG) or ' events.emit('verbose', MSG)

TimBarham: Grrr... That was actually debug code I intended to remove. Thought I caught them all.

stevengill: Haha it happens to me all the time too.

cordova-lib/src/cordova/plugin.js
```diff
@@ -360,6 +360,12 @@ function getPluginVariables(variables){
 
 function getVersionFromConfigFile(plugin, cfg){
     var pluginEntry = cfg.getPlugin(plugin);
+    if (!pluginEntry) {
+        var oldStylePluginId = pluginMapper[plugin];
```
omefire: a comment about why you are doing this would be great.

cordova-lib/src/cordova/plugin.js
```diff
@@ -360,6 +360,12 @@ function getPluginVariables(variables){
...
+        if (oldStylePluginId) {
```
omefire: 'oldStylePluginId' is not the correct name for what the registry-mapper (pluginMapper) returns. it returns the package name, which is the new way of doing things. so, I'd call it something like 'packageName' ..

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -364,6 +364,26 @@ function fetchPlugin(plugin, client, useNpmRegistry) {
     });
 }
 
+function processPluginVersion(plugin) {
+    // If plugin includes a version that is a caret range, the ancient version of npm we're using won't now how to
```
omefire: :nit 'now' -> 'know'

omefire:
Hey Tim, would "~" work fine in these scenarios ? does the semver version currently used by npm correctly handle "~" ?

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -364,6 +364,26 @@ function fetchPlugin(plugin, client, useNpmRegistry) {
     });
 }
 
+function processPluginVersion(plugin) {
```
omefire: How much work would it be to make use of 'getLatestMatchingNpmVersion' ? that way we don't have to maintain two separate ways of doing the same thing (platforms vs plugins).

omefire:
How much work would it be to make use of 'getLatestMatchingNpmVersion' ? that way we don't have to maintain two separate ways of doing the same thing (platforms vs plugins).

cordova-lib/src/cordova/plugin.js
TimBarham: Nope, it really is oldStylePluginId... Have a look at the declaration of pluginMapper.

cordova-lib/src/cordova/plugin.js
TimBarham: Good idea. Will add.

cordova-lib/src/plugman/registry/registry.js
TimBarham: I'm intentionally not doing that... Calling getLatestMatchingNpmVersion is expensive - I only use it for platform because we were already calling it..

cordova-lib/src/plugman/registry/registry.js
TimBarham: Will fix.

TimBarham:
Regarding '~' - yep it handles that fine. I also considered whether it would be worth just doing this for any range (in the plugin code - the platform code will already handle any range), to future-proof against further range shortcuts that might come in the future, but once the Cordova Plugin Registry goes read-only we'll be able to update our version of npm (and hopefully keep it up-to-date) so that shouldn't be an issue.

PR #245, opened 2015-06-10
Use cp instead of mv, as it would fail if tmp_dir is mounted on a dif…

sencenan:
Use cp instead of mv, as it would fail if tmp_dir is mounted on a different device from the plugin_dir. The tmp_dir is cleaned after copy already.

This fixes:
```
EXDEV, cross-device link not permitted '/tmp/git/1433949300428/Example'
```
when installing from git repository

### DIFF
--- a/cordova-lib/src/plugman/util/plugins.js
+++ b/cordova-lib/src/plugman/util/plugins.js
@@ -50,8 +50,14 @@ module.exports = {
             var plugin_dir = path.join(plugins_dir, plugin_id);
             events.emit('verbose', 'Copying fetched plugin over "' + plugin_dir + '"...');
             shell.mkdir('-p', plugin_dir);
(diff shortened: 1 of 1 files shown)


nikhilkh:
@TimBarham @omefire to take a quick look

omefire:
I'm trying to reproduce this error. @sencenan does this happen as well when you use different partitions of the same device ? or is it only on totally different devices ?

sencenan:
@omefire my workspace and /tmp are mounted on different harddrives.

Please also see this comment:

https://github.com/apache/cordova-lib/blob/024b970b26d3d8b7f9a6b2f0eab2325d7716deef/cordova-lib/src/cordova/lazy_load.js#L299

This is a known issue.

omefire:
LGTM !

PR #246, opened 2015-06-10
CB-9122 Added tests for platform/plugin add/rm/update with --save flag

alsorokin:
https://issues.apache.org/jira/browse/CB-9122

### DIFF
--- a/cordova-lib/spec-cordova/helpers.js
+++ b/cordova-lib/spec-cordova/helpers.js
@@ -17,12 +17,25 @@
     under the License.
 */
 
-var path = require('path'),
-    fs = require('fs'),
-    shell = require('shelljs'),
-    os = require('os');
+var path         = require('path'),
+    fs           = require('fs'),
+    shell        = require('shelljs'),
+    os           = require('os'),
+    ConfigParser = require('../src/configparser/ConfigParser');
 
-module.exports.tmpDir = function(subdir) {
+// Just use Android everywhere; we're mocking out any calls to the `android` binary.
+module.exports.testPlatform = 'android';
+
+function getConfigPath(dir) {
+    // if path ends with 'config.xml', return it
+    if (dir.indexOf('config.xml') == dir.length - 10) {
+        return dir;
+    }
+    // otherwise, add 'config.xml' to the end of it
+    return path.join(dir, 'config.xml');
+}
+
+module.exports.tmpDir = function (subdir) {
     var dir = path.join(os.tmpdir(), 'e2e-test');
     if (subdir) {
         dir = path.join(dir, subdir);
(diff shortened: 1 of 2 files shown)


cordova-lib/spec-cordova/save.spec.js
```diff
@@ -0,0 +1,435 @@
...
+            if (exists) {
+                cordova.raw.platform('rm', platformName).then(done, done);
+            } else {
+                done();
+            }
+        });
+    });
+
+    describe('(preparing fixtures)', function () {
+        it('preparing temp dir', function () {
```
omefire: Move this to 'beforeEach'.

cordova-lib/spec-cordova/save.spec.js
```diff
@@ -0,0 +1,435 @@
...
+            shell.rm('-rf', tempPath);
+            shell.mkdir(tempPath);
+        });
+
+        it('creating mocks', function () {
+            revert_copy_cordova_js = platform.__set__('copy_cordova_js', function () {});
```
omefire: Please, move this to 'beforeEach'.

cordova-lib/spec-cordova/save.spec.js
```diff
@@ -0,0 +1,435 @@
...
+        it('cloning platform', function (done) {
+            shell.exec('git clone ' + platformGitUrl + ' --depth 1 ' + platformLocalPath, { silent: true }, function (err) {
+                expect(err).toBe(0);
+                done();
+            });
+        }, timeout);
+    });
+
+    describe('(platform add)', function () {
+        it('spec.1 cusom tgz support: cordova platform add android@url_to_tgz', function (done) {
```
omefire: :nit 'cusom' -> 'custom'

cordova-lib/spec-cordova/save.spec.js
```diff
@@ -0,0 +1,435 @@
...
+                expect(true).toBe(false);
+                console.log(err.message);
+                done();
+            });
+        }, timeout);
+    });
+
+    describe('plugin update', function () {
+        //plugin update functionality is not yet implemented in cordova-cli
+        xit('spec.19 cordova plugin update cordova-plugin-console --save', function (done) {
```
omefire: forgot to change this back to 'it' ?

cordova-lib/spec-cordova/save.spec.js
```diff
@@ -0,0 +1,435 @@
...
+                done();
+            }).catch(function (err) {
+                expect(true).toBe(false);
+                console.log(err.message);
+                done();
+            });
+        }, timeout);
+
+        //plugin update functionality is not yet implemented in cordova-cli
+        xit('spec.20 cordova plugin update cordova-plugin-console@0.3.12 --save', function (done) {
```
omefire: forgot to change this back to 'it' ?

cordova-lib/spec-cordova/helpers.js
```diff
@@ -43,17 +54,101 @@ if (host.match(/win/)) {
...
+module.exports.setEngineSpec = function (appPath, spec) {
+    appPath = getConfigPath(appPath);
+    var parser = new ConfigParser(appPath);
+
+    parser.removeEngine(module.exports.testPlatform);
+    parser.addEngine(module.exports.testPlatform, spec);
+    parser.write();
+};
+
+module.exports.getEngineSpec = function (appPath) {
```
omefire: consider making this function more general by allowing the engine to be passed in as a separate parameter, instead of using the global variable module.exports.testPlatform.

cordova-lib/spec-cordova/helpers.js
```diff
@@ -43,17 +54,101 @@ if (host.match(/win/)) {
...
+
+    for (var i = 0; i < engines.length; i++) {
+        if (engines[i].name === module.exports.testPlatform) {
+            return engines[i].spec;
+        }
+    }
+    return null;
+};
+
+module.exports.removeEngine = function (appPath) {
```
omefire: please, pass in the engine to be removed as a separate parameter. I think that makes the intent of the function clearer.
Hate global vars, :)

cordova-lib/spec-cordova/save.spec.js
alsorokin: I didn't find this functionality in cordova code (and the test always fails) so I decided to mark it pending. Should I leave it failing?

cordova-lib/spec-cordova/save.spec.js
alsorokin: If this is moved, then 'creating test app' should also be moved, but creating test app before each test could significantly increase test run duration (289 -> 348 sec on my machine).

cordova-lib/spec-cordova/save.spec.js
omefire: That's because there's no 'cordova plugin update' functionality. Please, delete those tests.

omefire:
LGTM !

omefire:
Added a few minor comments.

omefire:
strange thing, why is there a commit from me saying: 'omefire reviewing alsorokin's code changes' ? :)

omefire:
Hmm @alsorokin review your second commit. it seems like you mistakenly committed something else ?

alsorokin:
This commit came from you, @omefire 
I can delete it if you want.

omefire:
Please, do.

alsorokin:
Done. Is it good to go?

cordova-lib/spec-cordova/helpers.js
```diff
@@ -17,12 +17,23 @@
...
+    shell        = require('shelljs'),
+    os           = require('os'),
+    ConfigParser = require('../src/configparser/ConfigParser');
 
-module.exports.tmpDir = function(subdir) {
+// Just use Android everywhere; we're mocking out any calls to the `android` binary.
+module.exports.testPlatform = 'android';
+
+function getConfigPath(dir) {
+    if (dir.indexOf('config.xml') !== dir.length - 10) {
```
omefire: Hey Alex, why 'dir.length - 10' ? I don't understand.

omefire: I see what you're doing there. I think you should use underscore (which we already include) to do that: http://aganov.github.io/underscore-strings/docs/underscore.strings.html#section-10

omefire:
LGTM !
@alsorokin, can you please:
 - Disable console logging (if possible, if not that's ok too.) ?
 - Use underscore's endWith function instead writing that hard to understand logic

cordova-lib/spec-cordova/helpers.js
purplecabbage: A simple line comment stating the intent to see if dir 'endsWith' config.xml would be helpful.
underscore would probably be overkill for this one method.

omefire: @purplecabbage, I was leaning towards a comment at first, but then I realized we already have underscore which provides this functionality. I think we should just use it and I don't find it to be overkill.

alsorokin:
- Disabling console logging seems to be really tricky in this case.Most of it comes from android platform code which runs in child_process.spawn.
i've tried wrapping it and always pass stdout: 'ignore' option to it, but it seem to break some functionality so I decided to leave it as is.
- the underscore library which is included to cordova-lib doesn't have the module for string processing, so it doesn't have 'endsWith' function. I've just added a clarifying comment.

vladimir-kotikov:
Merging. The AppVeyor failures, caused by node-xcode fixed in https://github.com/apache/cordova-lib/pull/253

omefire:
@alsorokin, sounds good. Thanks!

PR #247, opened 2015-06-12
CB-9162 - Default Variables

ogoguel:


cordova-lib/spec-plugman/install-browserify.spec.js
```diff
@@ -46,6 +46,7 @@ var install = require('../src/plugman/install'),
         'com.cordova.engine-android' : path.join(plugins_dir, 'com.cordova.engine-android'),
         'org.test.plugins.childbrowser' : path.join(plugins_dir, 'org.test.plugins.childbrowser'),
         'com.adobe.vars' : path.join(plugins_dir, 'com.adobe.vars'),
+        'org.test.defaultvariables' : path.join(plugins_dir, 'org.test.defaultvariables'),
```
TimBarham: @ogoguel - could you make the equivalent change to `install.spec.js`?

cordova-lib/spec-plugman/plugins/org.test.defaultvariables/plugin.xml
```diff
@@ -0,0 +1,18 @@
...
+    <name>Use Default Variables</name>
+
+    <!-- android -->
+    <platform name="android">
+		   <preference name="ENABLE_LOG" default="yes" />
+            <preference name="API_KEY" />
+    </platform>
+
+   
+</plugin>
```
TimBarham: Clean up whitespace (mixture of tabs and spaces).

cordova-lib/src/cordova/plugin.js
```diff
@@ -168,10 +168,19 @@ module.exports = function plugin(command, targets, opts) {
...
                         // Validate top-level required variables
-                        var pluginVariables = pluginInfoProvider.get(dir).getPreferences(),
-                            missingVariables = pluginVariables.filter(function (v) {
+                        var pluginVariables = pluginInfoProvider.get(dir).getPreferences();
+                        var requiredVariables = [];
+                        for(var i in pluginVariables)
+                        {
+                            var v = pluginVariables[i];
+                            // discard variables with default value
+                            if (!v.default)
```
TimBarham: When I test this adding a plugin that requires variables (for example, `com.phonegap.plugins.facebookconnect`) and I don't specify any, this line throws (`v` is `null`). The logic here doesn't seem to coincide with the actual value returned by `getPreferences()`.

TimBarham:
Thanks @ogoguel! A couple of things need fixing - I'll merge this once they're sorted.

cordova-lib/src/cordova/plugin.js
ogoguel: Beside adding a plugin, how this line of code can be unit-tested ?

ogoguel:
Fixed requiredVariables
Fixed whitespace in plugin.xml

cordova-lib/src/cordova/plugin.js
TimBarham: Have a look in `spec-cordova/plugin.spec.js`. You should be able to add something there. Create a `plugin.xml` with the required bits and add it under `spec-cordova/fixtures/plugins`.

ogoguel:
Added missing test for variables parsing

TimBarham:
Thanks for the update, @ogoguel. Squashed and merged.

cordova-lib/spec-plugman/plugins/org.test.defaultvariables/plugin.xml
```diff
@@ -0,0 +1,16 @@
...
+    id="org.test.defaultvariables"
+    version="3.0.0">
+
+    <name>Use Default Variables</name>
+
+    <!-- android -->
+    <platform name="android">
+            <preference name="ENABLE_LOG" default="yes" />
+            <preference name="API_KEY" />
+    </platform>
```
fujunwei: I only found the default `ENABLE_LOG` in `cordova/plugins/android.json`, but there is no similar preference `<preference name="ENABLE_LOG" value="yes" />` in config.xml, how to use the Default variable.

ogoguel: If you want to override the default variable, just specify it during the plugin installation : 
ex :` cordova plugin add <plugin_id> --variable ENABLE_LOG=no`  
This will update the `cordova/plugins/android.json`accordingly, as this is where the variables are being stored (and not in the config.xml)

fujunwei: How to use the variable updated in the `cordova/plugins/android.json`.

TimBarham: If you are creating a plugin, and need to use variables because some values will need to be defined by the app developer, you *use* the variable by including a reference to it in values in your plugin's `plugin.xml`. The value of the variable (whether the default value specified in the `preference` element, or a value supplied by the app developer when the plugin is added to their project) is substituted for references to the variable in values in `plugin.xml` when those values are used. See [here](https://cordova.apache.org/docs/en/5.1.1/plugin_ref_spec.md.html#variables) for some more info.

PR #251, opened 2015-06-20
CB-9225 Adds plugman support for the Windows platform.

TimBarham:


omefire:
LGTM !
Tim, question: Is this mainly to help guide people new to plugin development ?

omefire:
Also, If it's a feature people use a lot, do you think it might be a good idea to do something similar for Windows Phone ? (given that people mostly use cordova for mobile apps)

TimBarham:
Mefire - in response to your first question, yeah - `plugman` provides some tools to get started building your plugin - creates the initial structure and `plugin.xml`, and adds a basic bits when you add a platform. In response to your second question - the `windows` platform is the preferred way to build Windows mobile apps using Cordova.

PR #252, opened 2015-06-25
CB-8965 Prevent cli from copying cordova.js and cordova-js-src/ multiple times

surajpindoria:


stevengill:
Looks good to me!

cordova-lib/src/cordova/platform.js
```diff
@@ -161,8 +161,17 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
                     }
                     return superspawn.spawn(bin, args, copts);
                 }).then(function() {
-                    copy_cordova_js(projectRoot, platform);
-                    copy_cordovajs_src(projectRoot, platform, platDetails.libDir);
+                    var platform_www = path.join(projectRoot, 'platforms', platform, 'platform_www');
+
+                    // only want to copy cordova_js once, when the platform is added
+                    if (!fs.existsSync(path.join(platform_www, 'cordova.js'))) {
```
stevengill: You don't need this check. These functions only get run once when the platform gets added.

cordova-lib/src/cordova/platform.js
```diff
@@ -161,8 +161,17 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
...
+                        copy_cordova_js(projectRoot, platform);
+                    }
+
+                    // only want to copy cordova-js-src once, when the platform is added
+                    if (!fs.existsSync(path.join(platform_www, 'cordova-js-src'))) {
```
stevengill: Keep this check as we are moving this functionality to platform create scripts and don't want to run this command again if it already exists.

PR #254, opened 2015-07-08
CB-9330 Removes 'plugman publish' related functionality

vladimir-kotikov:
See [CB-9330](https://issues.apache.org/jira/browse/CB-9330) for more details.


cordova-lib/src/plugman/plugman.js
```diff
@@ -169,35 +160,6 @@ plugman.commands =  {
             }
         });
     },
-
-    'publish'  : function(cli_opts) {
```
nikhilkh: I wonder if these commands should still be exposed but with a meaningful console.error message on either having to use the older version of plugman < current or if we decide to make CPR read-only - stating that as such.

purplecabbage: +1 to leaving the function stubs around for a bit and simply have them output that this functionality is not supported anymore.

nikhilkh:
There is also an update to the docs that is needed.

vladimir-kotikov:
@nikhilkh done. See https://github.com/apache/cordova-docs/pull/296

PR #258, opened 2015-07-13
Add support for --target=edge to launch app in Edge browser.

TimBarham:
Note that this requires a fairly recent Windows 10 build (`10166` or later for Edge to support `localhost`, and probably around `10158` or later for the command line launch of Edge to work).


nikhilkh:
LGTM

PR #259, opened 2015-07-14
Fix prepare to wait the promise from plugman prepare

giuseppelt:
This fixes the issue where after_prepare hook is fired before the prepare process ends.

### DIFF
--- a/cordova-lib/src/cordova/prepare.js
+++ b/cordova-lib/src/cordova/prepare.js
@@ -106,28 +106,29 @@ function prepare(options) {
             if (options.browserify) {
                 plugman.prepare = require('../plugman/prepare-browserify');
             }
-            plugman.prepare(platformPath, platform, plugins_dir, null, true, pluginInfoProvider);
-
-            // Make sure that config changes for each existing plugin is in place
-            var platformJson = PlatformJson.load(plugins_dir, platform);
-            var munger = new PlatformMunger(platform, platformPath, plugins_dir, platformJson, pluginInfoProvider);
-            munger.reapply_global_munge();
-            munger.save_all();
-
-            // Update platform config.xml based on top level config.xml
-            var cfg = new ConfigParser(xml);
-            var platform_cfg = new ConfigParser(parser.config_xml());
-            exports._mergeXml(cfg.doc.getroot(), platform_cfg.doc.getroot(), platform, true);
-
-            // CB-6976 Windows Universal Apps. For smooth transition and to prevent mass api failures
-            // we allow using windows8 tag for new windows platform
-            if (platform == 'windows') {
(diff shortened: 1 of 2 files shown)


cordova-lib/src/cordova/prepare.js
```diff
@@ -106,28 +106,29 @@ function prepare(options) {
...
-            // we allow using windows8 tag for new windows platform
-            if (platform == 'windows') {
-                exports._mergeXml(cfg.doc.getroot(), platform_cfg.doc.getroot(), 'windows8', true);
-            }
-
-            platform_cfg.write();
-
-            return parser.update_project(cfg);
+            var preparePromise =  plugman.prepare(platformPath, platform, plugins_dir, null, true, pluginInfoProvider);
+            return Q(preparePromise).then(function () {
```
TimBarham: Since plugman.prepare() returns a promise, you should just be able to do this:

```js
return plugman.prepare(platformPath, platform, plugins_dir, null, true, pluginInfoProvider)
.then(function () {
```

TimBarham:
Thanks @giuseppelt. Looks good to me apart from the one stylistic comment. @stevengill does this look right to you?

cordova-lib/src/cordova/prepare.js
vladimir-kotikov: Actually `plugman.prepare` doesn't return promise, but `prepare-browserify` does. To handle both of them we need to wrap call into promise forcibly.

TimBarham: Oh, that's right. My bad. I had looked at this previously and forgot that I had updated `plugman.prepare()` to `return Q()` so I could use the simpler logic here.

giuseppelt: I used Q, because I checked the standard plugman prepare, and it doesn't return a promise. Instead browserify prepare returns a promise. 
So, to avoid the edit of the standard prepare to add a last line of 'return q.resolve', I decided to wrap with Q to handle "to promise or to not promise" business.

TimBarham: Yep, realized that eventually - sorry about that :smile:.

However, because ultimately the goal will be to replace the current prepare with browserify prepare , I'd prefer to keep this code simple and consistent with how browserify prepare works, and modify the `plugman.prepare` to return a promise. Then when the regular prepare goes away, we won't be left with unnecessary complexity here.

giuseppelt:
I made the changes to use the promise.
I signed the icla and mailed to secretary at apache. Do I have to do further actions?

TimBarham:
Thanks @giuseppelt. Saw your ICLA came through. Change is merged.

PR #260, opened 2015-07-15
CB-9371: Fix prepare deleting orientation preferences

cjpearson:


christoph-neumann:
This is a needed fix! Why has this not been merged?

chrskrchr:
Agreed - this is a useful change.  I want to be able to specify different orientation settings for iPhone vs. iPad at the Xcode project level, and this change would allow me to do that without Cordova clobbering my project-level settings.

mattrayner:
Is there any news on this?

mattrayner:
I'm guessing there's no news on this?

cjpearson:
@shazron, could you please take a look at this PR?

shazron:
@cjpearson unfortunately this patch can't be applied since the code has changed. Can you re-work it based on the master code? Also, the code you are trying to patch is the default fallback if a platform does not conform to the Platform API (older platforms). For newer platforms that do, like cordova-ios@4.x, they have their own Platform API implementations -- so the corresponding code would exist here: https://github.com/apache/cordova-ios/blob/08e00b68efbf2bfee21866f0bedd9c01ee8255e5/bin/templates/scripts/cordova/lib/prepare.js#L252-L275

cjpearson:
Thanks, @shazron. I've rebased this pr and opened a new one here https://github.com/apache/cordova-ios/pull/233

stevengill:
I believe we are deleting ios_parser.js for cordova@7 since that functionality has all been moved into  cordova-ios platform now. This would only be useful for older versions of cordova-ios (pre-API.js) now and we are dropping support for those older versions in cordova-7. I think this PR can be closed.

Thoughts @shazron?

shazron:
I think we can close this since its already in the Platform API for cordova-ios, and users will have to upgrade if they want this fix for cordova-ios < 4.

PR #262, opened 2015-07-17
restore-util.js can't handle git urls

hstaudacher:
When adding a plugin to the config.xml using a git url e.g. <plugin name="https://..." /> a spec is not needed because the plugin needs to be checked out by git. The restore-util.js has a bug that uses the spec of a plugin when the spec is not valid. This does not make sense because a) a spec without a name is not a valid plugin and b) a git url doe snot has a spec. This commit changes the resture-util to use the pluginEntry.name instead of the spec if the spec is not valid.

### DIFF
--- a/cordova-lib/src/cordova/restore-util.js
+++ b/cordova-lib/src/cordova/restore-util.js
@@ -102,7 +102,7 @@ function installPluginsFromConfigXML(args) {
         // Install from given URL if defined or using a plugin id. If spec isn't a valid version or version range,
         // assume it is the location to install from.
         var pluginSpec = pluginEntry.spec;
-        var installFrom = semver.validRange(pluginSpec, true) ? pluginEntry.name + '@' + pluginSpec : pluginSpec;
+        var installFrom = semver.validRange(pluginSpec, true) ? pluginEntry.name + '@' + pluginSpec : pluginEntry.name;
 
         // Add feature preferences as CLI variables if have any
         var options = {cli_variables: pluginEntry.variables,


omefire:
Hey @hstaudacher, I think you're using the plugin syntax within config.xml in a wrong way.

It's not supposed to be : <plugin name="https://..." />
It SHOULD be : <plugin name="name" spec="https://..." />

For example: <plugin name="cordova-plugin-file" spec="https://github.com/apache/cordova-plugin-file.git" />

I think we should file a documentation bug if it isn't clear enough.

hstaudacher:
sorry don't see your examples

omefire:
Ooops, updated my comments, now you should see the examples.

omefire:
Also keep in mind that instead of modifying the config.xml manually, you can do : `cordova plugin add https://github.com/apache/cordova-plugin-file.git --save` and it will do the appropriate modifications of config.xml for you.

TimBarham:
To clarify - the `name` is always the plugin name/id. The `spec` is either a version, a version range, or a "location" (locar folder, tarball, git repo) to get the plugin from. If `spec` is a version or range, then the name is necessary and we fully identify the plugin as `<name>@<version>`. If `spec` is a location, then the name is not necessary (since we will install whatever we find at the specified location, regardless of the name), and we can fully identify the plugin with just the location. So the existing code logic is correct.

hstaudacher:
Alright, thanks guys, I couldn't find any documentation for this. Will close the request now.

PR #265, opened 2015-07-17
CB-9384 Plugman should be able to checkout a commit SHA

purplecabbage:
actually checkout git_ref because it may be a branch OR a commit SHA

This functionality was lost when CB-8908 was implemented.

The fix will still clone with depth=1 IF a branch or a commit sha was not specified.

### DIFF
--- a/cordova-lib/src/gitclone.js
+++ b/cordova-lib/src/gitclone.js
@@ -21,7 +21,6 @@ var  Q             = require('q'),
      shell         = require('shelljs'),
      events        = require('./events'),
      path          = require('path'),
-     semver        = require('semver'),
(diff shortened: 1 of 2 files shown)


omefire:
I wasn't a fan of the `cryptic` optimizations we were doing to begin with, as it rendered the code more cryptic than it needed to be.

omefire:
@purplecabbage LGTM !
But could you also add a comment to specify all the forms that 'git_ref' could assume ? (branch, commit SHA, etc...). That might prevent attempts at optimizations that will break this scenario again.

purplecabbage:
Yeah. I'll add it to the docs too

omefire:
Awesome ! Thanks !

PR #266, opened 2015-07-21
CB-9278: BugFix: Restoring multiple platforms fails

omefire:


cordova-lib/src/cordova/restore-util.js
```diff
@@ -56,24 +57,23 @@ function installPlatformsFromConfigXML(platforms) {
...
+    // Otherwise, we get a bug where the following line: https://github.com/apache/cordova-lib/blob/0b0dee5e403c2c6d4e7262b963babb9f532e7d27/cordova-lib/src/util/npm-helper.js#L39
+    // gets executed more simultaneously by each platform and leads to an exception being thrown
+    return promiseutil.Q_chainmap_graceful(targets, function(target) {
         if (target) {
             events.emit('log', 'Restoring platform ' + target + ' referenced on config.xml');
             return cordova.raw.platform('add', target);
         }
         return Q();
+    }, function(err) {
+        events.emit('log', err);
```
nikhilkh: Should these be logged as 'error' instead of 'log'

nikhilkh:
One small comment. LGTM otherwise.

omefire:
@nikhilkh, log level updated.

vladimir-kotikov:
LGTM. I would also avoid unnecessary whitespace and formatting changes.

cordova-lib/src/cordova/restore-util.js
```diff
@@ -56,24 +57,23 @@ function installPlatformsFromConfigXML(platforms) {
...
+
+    // Run `platform add` for all the platforms separately
     // so that failure on one does not affect the other.
-    var promises = targets.map(function (target) {
+
+    // CB-9278 : Run `platform add` serially, one platform after another
+    // Otherwise, we get a bug where the following line: https://github.com/apache/cordova-lib/blob/0b0dee5e403c2c6d4e7262b963babb9f532e7d27/cordova-lib/src/util/npm-helper.js#L39
+    // gets executed simultaneously by each platform and leads to an exception being thrown
+    return promiseutil.Q_chainmap_graceful(targets, function(target) {
         if (target) {
```
vladimir-kotikov: nit: I wonder if `target` can be falsy? Probably this condition is not required.

cordova-lib/src/util/promise-util.js
```diff
@@ -33,4 +33,21 @@ function Q_chainmap(args, func) {
     });
 }
 
+// Behaves similar to Q_chainmap but gracefully handles failures.
+// When a promise in the chain is rejected, it will call the failureCallback and then continue the processing, instead of stopping
```
vladimir-kotikov: Nit: The line is twice as longer than previous one. It might look better if split into two lines.
Another nit: Since this method is 'public', the comment might be rewritten using JSDoc notation.

nikhilkh:
@omefire - Is this ready to be merged? It will be good to get this merged before the next cordova-lib release.

omefire:
@nikhilkh , it's ready to be merged in.

nikhilkh:
Thinking about it - wonder if you can add test case for this to prevent future regression. It's really a basic use case here.

dblotsky:
Also tried on OS X with two platforms, and it worked. +1 for adding a test case for this.

TimBarham:
Made this comment offline, but decided it should be included here too... I don't think this is the right fix - I think a more general approach would serve use better in the long run. There may be other scenarios (now or in the future) where we end up calling into `npm-helper` with multiple `npm` actions that aren't triggered sequentially. I think it would be better/safer to update `npm-helper` to handle that properly, rather than having to modify existing cases (or remember to always do it sequentially in the future).

Since `npm-helper`'s `loadWithSettingsThenRestore()` method ensures we always restore (even if something fails), if subsequent calls come in before we've restored from a previous call (which is what currently triggers the error), we can queue up those subsequent calls and process each them (and hence fulfill the corresponding promise) only once we've restored for all the calls that preceded it.

This shouldn't be too hard to code up, and provides us with a more robust solution at the root of the problem.

omefire:
@TimBarham , I've explored that possibility before landing with the current solution.
I think messing around with a queue-based solution or the event loop is going to take more time and more testing.
My goal is to have this issue fixed and checked in ASAP since it impacts such a basic scenario.
I can revisit making the change to npmhelper in the future.

omefire:
@TimBarham , I've explored that possibility before landing with the current solution.
I think messing around with a queue-based solution or the event loop is going to take more time and more testing.
My goal is to have this issue fixed and checked in ASAP since it impacts such a basic scenario.

I can revisit making the change to npmhelper in the future. we can even go ahead and create another JIRA bug to track that refactoring, but I don't think it should stop us from moving forward with the fix to this issue.

PR #268, opened 2015-07-24
CB-9408: Added support for `windows-packageVersion` on `<widget>`.

robpaveza:
Corresponds to a similar change on cordova-windows.


omefire:
LGTM !

PR #269, opened 2015-07-28
CB-9421 Fixed searchpath parameter being ignored

alsorokin:
https://issues.apache.org/jira/browse/CB-9421

### DIFF
--- a/cordova-lib/src/plugman/fetch.js
+++ b/cordova-lib/src/plugman/fetch.js
@@ -236,7 +236,8 @@ function findLocalPlugin(plugin_src, searchpath, pluginInfoProvider) {
 
     versions.forEach(function(pinfo) {
         // Ignore versions that don't satisfy the the requested version range.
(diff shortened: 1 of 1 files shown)


cordova-lib/src/plugman/fetch.js
```diff
@@ -236,7 +236,7 @@ function findLocalPlugin(plugin_src, searchpath, pluginInfoProvider) {
 
     versions.forEach(function(pinfo) {
         // Ignore versions that don't satisfy the the requested version range.
-        if (!semver.satisfies(pinfo.version, versionspec)) {
+        if (!semver.satisfies(pinfo.version.replace(/-dev$/, ''), versionspec)) {
```
vladimir-kotikov: This is not obvious, why this change is required. Could you please add a link to JIRA/semver docs or short explanation, why this needed.

purplecabbage:
lgtm!

PR #270, opened 2015-07-28
CB-9420 Fixes malformed require calls in browserify bundle

vladimir-kotikov:
This fixes browserify bundle generation problems, desctibed in [CB-9420](https://issues.apache.org/jira/browse/CB-9420)


stevengill:
LGTM! I'll test and merge it in!

PR #271, opened 2015-07-28
Add platform project instance in hook options context

galexandrov:
Add platform project instance in hook options context for install/uninstall plugman commands.
This is handy when you need to make changes on the project from the hook script. 
Sample scenario is when you need to change project file from hook script. With this change it will be possible to get the parsed project file

``` javascript
var projectRoot = context.opts.projectRoot,
    projectFile = context.opts.cordova.project.parseProjectFile(projectRoot);
```


galexandrov:
run ci

galexandrov:
ping @omefire

PR #274, opened 2015-07-30
CB-9407 FFOS permissions missing in manifest

zalun:
CB-9407
running an update on platform_cfg allows permissions from plugins to feed the manifest.


zalun:
R? please

PR #275, opened 2015-07-31
CB-9436 Removes `require-tr` bundle transformation

vladimir-kotikov:
This PR replaces excess `require-tr` browserify transform (- 2 deps and lot of code) with `aliasify`.

It also solves two more problems from [Browserify Master Issue](https://issues.apache.org/jira/browse/CB-8801):
  * browserify and mobilespec (see https://issues.apache.org/jira/browse/CB-8802) - because all bundled modules now exposed outside of bundle
  * excess plugins' dirs in `www` folder (see https://issues.apache.org/jira/browse/CB-9038)

NOTE: This PR depends on apache/cordova-js#124.


cordova-lib/src/plugman/prepare-browserify.js
```diff
@@ -21,7 +21,7 @@
 
 var platform_modules   = require('../platforms/platforms'),
     path               = require('path'),
-    through            = require('through2'),
+    aliasify           = require('aliasify'),
```
nikhilkh: Should the 'through2' dependency be removed from package.json?

stevengill:
Vladimir! This is a much needed refactor and it looks great! Thank you! The previous spaghetti code was becoming hard to manage and understand. 

Things I like:
* Using `cordova_plugins.js` in the browserify build and removing `cordova_requires`
* Adding the module name field in `cordova_plugins.js` for each module.
* Using browserify expose to map `cordova/plugin_list` to `cordova_plugins.js`. (Didn't even know about expose)
* Great find with aliasify! Much cleaner way to handle the bad legacy requires in plugins compared to using requiresTr.

Quick Question without digging to deep:
* Wondering why you had to add require('codrdova/init') to the bundle? https://github.com/MSOpenTech/cordova-lib/commit/510516135ef41b735b2e610c1b8d95fc77bd571e#diff-28a16640f3807e6ee2300ea7ce8c2f1dR196. How was that getting added previously? 

I think you fixed all of the remaining bugs with your changes. @surajpindoria will continue to test on all of the devices we have here to see if we can find bugs.

I am going to dive deeper into your cordova.js changes now.

vladimir-kotikov:
> Wondering why you had to add require('codrdova/init') to the bundle? MSOpenTech@5105161#diff-28a16640f3807e6ee2300ea7ce8c2f1dR196. How was that getting added previously?

Previously this was being added by `cordova-js/tasks/lib/bundle-browserify.js` as `cordova/bootstrap` module. The problem is when you adding plugin modules into bundle after bootstrap is added, they only get loaded _after_ cordova initialization and `deviceready` event. The only way i found to fix this, is to add `bootstrap` as the last file in the bundle.

stevengill:
Make sense. Thanks! Lets merge it in!

PR #276, opened 2015-08-04
CB-9278: Tests to prevent future regressions while restoring multiple platforms

omefire:
Tests to prevent future regressions while restoring multiple platforms.


omefire:
@muratsu, please review.

PR #280, opened 2015-08-19
CB-9434: Adding a ToDO. 'cordova run' should run build related hooks …

omefire:
…: 'before_build' & 'after_build' hooks.


omefire:
For further details, see : https://issues.apache.org/jira/browse/CB-9434

PR #281, opened 2015-08-19
CB-9517 Adding a plugin on iOS/OSX that uses a private framework does…

tripodsan:
see also related PR https://github.com/apache/cordova-osx/pull/14

todo:
* update the project templates for osx / ios if desired or create the copy-files build phase automatically
* talk to the guys from node-xcode if we should add this to their code


shazron:
Fails the cordova-lib npm tests.
jshint -- "src/plugman/platforms/ios.js: line 107, col 34, Mixed double and single quotes."

PR #282, opened 2015-08-21
CB-9597 Adds implementation for PlatformApiPoly class according to PlatformApi spec

vladimir-kotikov:
This implements PlatformApiPoly class according to PlatformApi spec, which allows to:
  * create/update platform
  * execute platform's actions (build/run/add/update)
  * do a prepare (needed for CLI workflow only)
  * install/uninstall plugins

Other noticeable changes:
  * removes `getPlatformProject` and PlatformProject method/class in favor of PlatformApiPoly/getPlatformApi
  * make assets and js-modules installing/uninstalling through ActionStack
  * refactor configChanges to not require plugins_dir in constructor
  * moves mergeXml helper to xml-helpers

This should be used along with cordova-cli/platformApi branch: https://github.com/apache/cordova-cli/pull/220


kamrik:
Looks good! and very clean.

One note about tests, jasmine spies seem to give you a lot of trouble. In the previous refactoring round I was doing, I found it very useful to entirely replace some tests that rely on spies too much with something based on filesystem fixtures with less spying.

cordova-lib/src/plugman/util/config-changes.js
```diff
@@ -168,22 +167,18 @@ function remove_plugin_changes(plugin_name, plugin_id, is_top_level) {
     }
 
     // Remove from installed_plugins
-    if (is_top_level) {
-        delete platform_config.installed_plugins[plugin_id];
-    } else {
-        delete platform_config.dependent_plugins[plugin_id];
-    }
+    self.platformJson.addPlugin(pluginInfo.id, is_top_level);
```
vladimir-kotikov: FIXME: probably should be `removePlugin()`

vladimir-kotikov:
Thanks for advice, Mark. I have some plans about tests update and definitely will try fixtures-based approach.

stevengill:
Looks great Vlad!  I'm fine with you merging this in and us spending some time testing it on master to make sure it works with all of our platforms.

PR #286, opened 2015-08-26
Fixed usage of npm plugins on Ubuntu Phone

tsschaffert:
Plugins with an id in the form of cordova-plugin-<name> could not be added to the ubuntu platform. Quick and dirty fix to avoid this problem, maybe someone with more knowledge about the project can come up with a better solution.


tsschaffert:
Duplicate of #294

PR #287, opened 2015-08-26
CB-9556 Don't uninstall dependent plugin if it was installed as a top-level after

vladimir-kotikov:
This fixes [CB-9556](https://issues.apache.org/jira/browse/CB-9556)

### DIFF
--- a/cordova-lib/src/plugman/install.js
+++ b/cordova-lib/src/plugman/install.js
@@ -274,7 +274,12 @@ function runInstall(actions, platform, project_dir, plugin_dir, plugins_dir, opt
 
     if (platformJson.isPluginInstalled(pluginInfo.id)) {
         if (options.is_top_level) {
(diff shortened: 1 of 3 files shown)


cordova-lib/src/plugman/install.js
```diff
@@ -273,8 +273,13 @@ function runInstall(actions, platform, project_dir, plugin_dir, plugins_dir, opt
     var platformJson = PlatformJson.load(plugins_dir, platform);
 
     if (platformJson.isPluginInstalled(pluginInfo.id)) {
+        var msg = 'Plugin "' + pluginInfo.id + '" already installed on ' + platform + '.';
```
omefire: :nit please move this line within the if statement : `if (options.is_top_level) ...`

omefire:
LGTM !

cordova-lib/src/plugman/install.js
vladimir-kotikov: Done.

PR #288, opened 2015-08-26
CB-9560 Issue using plugin restore for plugins with common dependencies

marcuspridham:
Similar change to CB-9278.  This change makes it so only one plugin is restored at a time.

### DIFF
--- a/cordova-lib/src/cordova/restore-util.js
+++ b/cordova-lib/src/cordova/restore-util.js
@@ -90,7 +90,11 @@ function installPluginsFromConfigXML(args) {
         return Q.all('No config.xml plugins to install');
     }
 
-    var promises = plugins.map(function(featureId) {
+    // CB-9560 : Run `plugin add` serially, one plugin after another
+    // We need to wait for the plugin and its dependencies to be installed
+    // before installing the next root plugin otherwise we can have common
+    // plugin dependencies installed twice which throws a nasty error.
+    return promiseutil.Q_chainmap_graceful(plugins, function(featureId) {
         var pluginPath = path.join(plugins_dir, featureId);
         if (fs.existsSync(pluginPath)) {
             // Plugin already exists
@@ -111,5 +115,4 @@ function installPluginsFromConfigXML(args) {
         };
         return plugin('add', installFrom, options);
     });
-    return Q.allSettled(promises);
 }


nikhilkh:
@omefire to review

omefire:
@marcuspridham, could you please add a comment like this one ?
That way, we could signal other devs not to try and optimize this piece of code by parallelizing it ( switching back to 'map' function).

omefire:
@marcuspridham, could you please add a comment like the one added for CB-9278 ?
That way, we could signal other devs not to try and optimize this piece of code by parallelizing it ( switching back to 'map' function).

omefire:
Also, does this happen when you use the following two plugins : cordova-plugin-file and cordova-plugin-file-transfer ?
Otherwise, could you please provide some example plugins which elicit this behavior ?

omefire:
Also, does this happen when you use the following two plugins : cordova-plugin-file and cordova-plugin-file-transfer ?
Otherwise, could you please provide some example plugins which elicit this behavior ?
I'd like to test using them.

marcuspridham:
Added comments.  I'll try to put together a sample that shows the issue.  The ones I am using are not available on NPM.

marcuspridham:
Yep it does seem to happen with cordova-plugin-file-transfer and cordova-plugin-file.  I added the following to the default cordova project config.xml and added a platform
    <plugin name="cordova-plugin-file-transfer" spec="~1.2.1" />
    <plugin name="cordova-plugin-file" spec="~3.0.0" />

You can see in the output that it is trying to install the file plugin twice.  With my change the error does not happen...

Cyar:cdvtest marcuspridham$ cordova platform add android
Adding android project...
Creating Cordova project for the Android platform:
	Path: platforms/android
	Package: io.cordova.hellocordova
	Name: HelloCordova
	Activity: MainActivity
	Android target: android-22
Copying template files...
Android project created with cordova-android@4.1.1
Discovered plugin "cordova-plugin-whitelist" in config.xml. Installing to the project
Discovered plugin "cordova-plugin-file-transfer" in config.xml. Installing to the project
Discovered plugin "cordova-plugin-file" in config.xml. Installing to the project
Fetching plugin "cordova-plugin-whitelist@1" via npm
Fetching plugin "cordova-plugin-file-transfer@~1.2.1" via npm
Fetching plugin "cordova-plugin-file@~3.0.0" via npm
Installing "cordova-plugin-whitelist" for android
Installing "cordova-plugin-file-transfer" for android
Fetching plugin "cordova-plugin-file" via npm
Installing "cordova-plugin-file" for android
Installing "cordova-plugin-file" for android

The Android Persistent storage location now defaults to "Internal". Please check this plugins README to see if you application needs any changes in its config.xml.
      
If this is a new application no changes are required.
      
If this is an update to an existing application that did not specify an "AndroidPersistentFileLocation" you may need to add:
      
      "<preference name="AndroidPersistentFileLocation" value="Compatibility" />"
      
to config.xml in order for the application to find previously stored files.
      
    
Error during processing of action! Attempting to revert...
Failed to install 'cordova-plugin-file':Error: Uh oh!
"/Users/marcuspridham/cdvtest/platforms/android/src/org/apache/cordova/file/EncodingException.java" already exists!

omefire:
Cool @marcuspridham . thanks for the confirmation.

omefire:
This issue arises because [common.copyFileNew] (https://github.com/apache/cordova-lib/blob/dff78b0779444d06e259140694e07d5e8c5f5308/cordova-lib/src/plugman/platforms/common.js#L64) throws an exception when it tries to copy a file that already exists, having been placed there by a previously installed plugin.
So, I'm curious why does this not show up when we do the copy in serial, but shows up when we do it in parallel ?
Seems fishy to me.

omefire:
This issue arises because [common.copyFileNew] (https://github.com/apache/cordova-lib/blob/dff78b0779444d06e259140694e07d5e8c5f5308/cordova-lib/src/plugman/platforms/common.js#L64) throws an exception when it tries to copy a file that already exists, having been placed there by a previously installed plugin.
So, I'm curious why does this not show up when we do the copying in serial, but shows up when we do it in parallel ?
Seems fishy to me.

omefire:
This issue arises because [common.copyFileNew] (https://github.com/apache/cordova-lib/blob/dff78b0779444d06e259140694e07d5e8c5f5308/cordova-lib/src/plugman/platforms/common.js#L64) throws an exception when it tries to copy a file that already exists, having been placed there by a previously installed plugin.
So, I'm curious why does this not show up when we do the copying in serial, but shows up when we do it in parallel ? [the way the code is structured, it should happen no matter what.].
Seems fishy to me.

dpogue:
A related issue about dependency versions from config.xml not being respected when they are dependencies of another plugin is [CB-9525](https://issues.apache.org/jira/browse/CB-9525)

marcuspridham:
In serial it seems to be able to detect the file plugin is already installed and does not try to install it again.  The parallel way you can see in the logs that it does not see the plugin is installed yet so it runs the install again. There is a race condition.

omefire:
I see. The parallel mode exhibits a race condition with regards to this check : [platformJson.isPluginInstalled](https://github.com/apache/cordova-lib/blob/bed6d3a201e0193b2ffdb60fa6de74de0c870cec/cordova-lib/src/plugman/install.js#L275).

omefire:
@dpogue , I see what's going on there. This might call for us to be able to handle dependent plugin's versioning in Cordova.

omefire:
LGTM ! @marcuspridham, I'll pull these changes in.

omefire:
Merged !

purplecabbage:
@omefire These pull requests should just close themselves if you merge->push
Are you rebasing?

omefire:
@purplecabbage This one did close itself. (asfgit closed it)

PR #290, opened 2015-08-27
CB-7015 Add support for parent and custom attributes in lib-file element

mbektchiev:
This enables referring to libraries from the Android SDK directory.

Refs https://issues.apache.org/jira/browse/CB-7015

### DIFF
--- a/cordova-common/src/PluginInfo/PluginInfo.js
+++ b/cordova-common/src/PluginInfo/PluginInfo.js
@@ -228,6 +228,8 @@ function PluginInfo(dirname) {
             return {
                 itemType: 'lib-file',
                 src: tag.attrib.src,
+                parent: tag.attrib.parent,
+                custom: !isStrFalse(tag.attrib.custom), // default to true if not set
                 arch: tag.attrib.arch,
                 Include: tag.attrib.Include,
                 versions: tag.attrib.versions,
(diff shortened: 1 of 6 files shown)


mbektchiev:
@stevengill Can you please review this PR?

stevengill:
I'll review them next week (off tomorrow). Thanks for the ping.

mbektchiev:
ping @stevengill

EddyVerbruggen:
It would be really great to have this feature as this is one of my main headaches when developing (and maintaining) Cordova Android plugins. This would definitely help make plugins play nice with each other.

stevengill:
Why add new functionality when you can just use framework tag? See push plugin from @macdonst https://github.com/phonegap/phonegap-plugin-push/blob/master/plugin.xml#L60

mbektchiev:
Because these dependencies are only when using gradle. For projects that work with ant you have no way to refer a single jar.

macdonst:
I'm not in favor of this PR being merged into master as I don't think we should add the functionality CB-7015 asks for. There is no need to add a new tag to support as the framework tag accomplishes exactly the same functionality. Yes, this requires that your Android project use Gradle but Gradle support for Cordova Android has been around for almost a year so it is time to upgrade your Cordova projects to use it.

It doesn't make sense to add this code that duplicates functionality for a build system that is deprecated. If we add this code we will be responsible for testing and maintaining it going forward.

-1

EddyVerbruggen:
Hey @macdonst do you know if all PhoneGap Build cordova-android builds are Gradle-enabled? Thx

mbektchiev:
Won't fix.

PR #291, opened 2015-08-28
CB-9559 Adding a plugin with caret in version results in an error

TimBarham:
When Cordova used `npm 1.3.x` internally, we had to process "caret versions" since `npm 1.3.x` didn't understand them. As part of that, we'd turn the caret version into a fully defined version range, which contained a space, and `npm 1.3.x` required that we wrap such a range in quotes.

This logic broke when we upgraded to `npm 2.10.x` - it no longer requires the quotes in this scenario, and in fact fails if we include them. However, we also no longer need to do this processing (since `npm 2.10.x` understands caret versions), so we can just remove that code altogether.

### DIFF
--- a/cordova-lib/src/plugman/registry/registry.js
+++ b/cordova-lib/src/plugman/registry/registry.js
@@ -19,8 +19,7 @@
 
 /* jshint laxcomma:true */
 
-var semver = require('semver'),
-    npm = require('npm'),
+var npm = require('npm'),
     path = require('path'),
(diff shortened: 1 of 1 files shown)


omefire:
+1

purplecabbage:
Thanks for the background on this Tim.

TimBarham:
Np :smile:

macdonst:
+1

PR #292, opened 2015-08-31
CB-8993 Plugin restore ignores search path

marcuspridham:
When restoring a platform the search path needs to be provided otherwise
plugins will always be resolved to npm.

### DIFF
--- a/cordova-lib/src/cordova/prepare.js
+++ b/cordova-lib/src/cordova/prepare.js
@@ -56,7 +56,7 @@ function prepare(options) {
     var hooksRunner = new HooksRunner(projectRoot);
     return hooksRunner.fire('before_prepare', options)
     .then(function(){
-        return restore.installPlatformsFromConfigXML(options.platforms);
+        return restore.installPlatformsFromConfigXML(options.platforms, { searchpath : options.searchpath });
     })
     .then(function(){
         options = cordova_util.preProcessOptions(options);
--- a/cordova-lib/src/cordova/restore-util.js
+++ b/cordova-lib/src/cordova/restore-util.js
@@ -32,7 +32,7 @@ exports.installPluginsFromConfigXML = installPluginsFromConfigXML;
 exports.installPlatformsFromConfigXML = installPlatformsFromConfigXML;
 
 
-function installPlatformsFromConfigXML(platforms) {
+function installPlatformsFromConfigXML(platforms, opts) {
     var projectHome = cordova_util.cdProjectRoot();
     var configPath = cordova_util.projectConfig(projectHome);
     var cfg = new ConfigParser(configPath);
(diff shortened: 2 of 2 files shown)


cordova-lib/src/cordova/prepare.js
```diff
@@ -56,7 +56,7 @@ function prepare(options) {
     var hooksRunner = new HooksRunner(projectRoot);
     return hooksRunner.fire('before_prepare', options)
     .then(function(){
-        return restore.installPlatformsFromConfigXML(options.platforms);
+        return restore.installPlatformsFromConfigXML(options.platforms, { searchpath : options.searchpath });
```
omefire: I don't see this searchpath ption being used by [the downstream function]( https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/cordova/restore-util.js#L35 ).

omefire: Oops never mind, I see the changes you have made to restore-utils.

omefire:
Thanks for suggesting these changes.

omefire:
Thanks for suggesting these changes. 

However, I have a suggestion :
As it currently stands, it seems like adding adding a platform or restoring from config.xml is broken.
It doesn't handle cases where the searchpath comes from config.json or the case when the searchpath could be an array of multiple paths.

I propose we include that fix with this one. What do you think ?

For references, see : [Handling searchpaths as string or array](https://github.com/apache/cordova-lib/blob/41765f1fd7393be1fff2b39ba7cf00b5bf6f1ef1/cordova-lib/src/cordova/plugin.js#L97).

marcuspridham:
Thanks.  I just tried these cases it appears to already work. The search path provided with prepare passes through https://github.com/apache/cordova-lib/blob/41765f1fd7393be1fff2b39ba7cf00b5bf6f1ef1/cordova-lib/src/cordova/plugin.js#L97 before a plugin is added so it is always processed to handle config.json and multiple paths.

I tried these cases they all appear to work for me.

1. Restore plugins using plugin_search_path in .cordova/config.json.
2. Restore plugins using combination of plugin_search_path and cli provided search path.  One plugin located in each location.
3. Restore plugins using 2 search paths on cli.  cordova prepare --searchpath ../plugin_cache:../plugin_cache_2/

omefire:
Awesome, Thanks for validating. 
The case where we specify a searchpath on the CLI and another one in the ./cordova/config.json doesn't go through. I will open up a separate issue for that.

omefire:
LGTM !! will be merged in.

omefire:
@marcuspridham , did you sign the [ICLA](http://www.apache.org/licenses/#clas) ?

marcuspridham:
Yep.  I signed the ICLA awhile ago.  I have done some pull requests to Cordova before.

omefire:
#merged !

omefire:
#merged !
Please, close this PR.

marcuspridham:
Thanks! Please look at merging CB-9560 when you time.

omefire:
CB-9560 is merged as well.

PR #293, opened 2015-09-01
CB-9587 Check if browser platform added properly before creating parser

vladimir-kotikov:
This fixes weird error when trying to run app on browser platform but the platform isn't added to project. See [CB-9587](https://issues.apache.org/jira/browse/CB-9587) for details and repro steps.


TimBarham:
Seems reasonable. One minor concern... The check for the `cordova` directory doesn't really validate that this is a `browser` platform directory. Not sure of what might be better, however :smile:.

vladimir-kotikov:
Yes, i didn't found any 'browser'-specific things to check too.

PR #295, opened 2015-09-03
CB-9604 Fix error adding browser platform with PlatformApi polyfill.

TimBarham:
Don't try to create `ConfigParser()` in `PlatformApiPoly` constructor, since we may not have copied the `config.xml` file into place yet. Instead, create it once we know we have the `config.xml` file.


TimBarham:
@vladimir-kotikov - does this fix look right to you?

vladimir-kotikov:
Looks good. However there is another chance to access `this._config` before it will be initialized: https://github.com/MSOpenTech/cordova-lib/blob/CB-9604/cordova-lib/src/platforms/PlatformApiPoly.js#L162.

Probably it makes sense to eliminate "private" `this._config` and always use `platformInfo.projectConfig` instead, which need to be a dynamic property with getter? Your thoughts? I can refactor this by myself after.

PR #297, opened 2015-09-07
CB-9617 Fixes incorrect project state after adding/removing plugins

vladimir-kotikov:
This fixes [CB-9617](https://issues.apache.org/jira/browse/CB-9617)

The cause of problem is that `plugman.prepare` functionality was eliminated (since the new flow assumes that plugin installation is totally handled by `PlatformApi` instance) and plugin's assets and js-files are now get copied into app's `www` folder by `PlatforfmApi.prepare`.
Due to this at the time after _plugin installation_ and before _prepare_ the platform folder might exist in incorrect state.


vladimir-kotikov:
@TimBarham, please review when possible.

TimBarham:
LGTM. I've merged it.

vladimir-kotikov:
:+1:  Thanks

PR #299, opened 2015-09-07
icon parameter should not be relative to the www directory on Ubuntu

david-barth-canonical:


tsschaffert:
Shouldn't this be handled similar to how it is done in the manifest generation of cordova-ubuntu? Then the 'www' part should be replaced by '../..' (otherwise icons in a top-level res folder are not found).

See https://github.com/apache/cordova-ubuntu/blob/master/bin/templates/project/cordova/lib/manifest.js#L48

david-barth-canonical:
Yes, makes sense. Icon management needs more tweaks, but this way it will work better already Thanks for the review.

PR #300, opened 2015-09-08
CB-9601 Fix <framework>.versions support on Windows after semver update

sgrebnov:
https://issues.apache.org/jira/browse/CB-9601

Adds logic to convert Windows specific four-digits version to three-digits semver compatible version.

### DIFF
--- a/cordova-lib/src/util/windows/jsprojManager.js
+++ b/cordova-lib/src/util/windows/jsprojManager.js
@@ -264,7 +264,8 @@ jsprojManager.prototype = {
 
         if (target || versions) {
             var matchingProjects = this.projects.filter(function (project) {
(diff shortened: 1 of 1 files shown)


nikhilkh:
It will be great to add test cases for this scenario.

sgrebnov:
Problem w/ tests is that we have e2e tests but they miss windows10 proj target. Since I've added more functionality and corresponding tests (+ missing windows10 proj target) in another PR - #301 I've decided not to duplicate that code here as it will conflict. So if we merge #301 corresponding tests will be there.

PS. It is hard to split tests on those PRs separately as they both require same changes like missing windows10 target.

nikhilkh:
LGTM

PR #301, opened 2015-09-08
CB-8615, CB-9588 Better Cordova plugin support for Windows 10 assemblies

sgrebnov:
* `target-dir` support for `<framework>` elements  so that plugin devs can specify different subdirs for platform specific libs and avoid naming conflict(CB-8615)
* `<resource-file>` support(CB-9588)
* missing Windows10 project tests
* tests to cover CB-9588 and CB-8615

Some of unit tests might fail as we added tests for Windows10 project and there is the following fix which is required for correct semver work (CB-9601): https://github.com/apache/cordova-lib/pull/300

### DIFF
--- a/cordova-lib/spec-plugman/platforms/windows.spec.js
+++ b/cordova-lib/spec-plugman/platforms/windows.spec.js
@@ -73,7 +73,8 @@ beforeEach(function () {
             all: 'CordovaApp.projitems',
             phone: 'CordovaApp.Phone.jsproj',
             windows: 'CordovaApp.Windows.jsproj',
-            windows8: 'CordovaApp.Windows80.jsproj'
+            windows8: 'CordovaApp.Windows80.jsproj',
+            windows10: 'CordovaApp.Windows10.jsproj'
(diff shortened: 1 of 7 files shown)


nikhilkh:
Looks like there are some CI test failures

nikhilkh:
@robpaveza to have a close look

sgrebnov:
Tests fail as we added corresponding unit tests including tests for windows 10 proj which has semver related issue(CB-9601): https://github.com/apache/cordova-lib/pull/300 . If I apply that patch, then all tests pass.
PS. Do you  want me to combine two PRs into a single one so that we have all changes, tests and required patches in one place?

sgrebnov:
Rebased on top of master which now includes #300

sgrebnov:
Tests pass now

vladimir-kotikov:
Tested, looks good.

vladimir-kotikov:
@sgrebnov, please update apache/cordova-docs with new attributes description.

PR #303, opened 2015-09-09
CB-9631 Save plugin to config.xml only if installation succeeds

vladimir-kotikov:
This fixes [CB-9631](https://issues.apache.org/jira/browse/CB-9631)

The idea is that we need to save installed plugins to config.xml only if installation succeeds (similar to platform save)


vladimir-kotikov:
@omefire, review please

omefire:
LGTM !

vladimir-kotikov:
Thanks for fast response!

PR #304, opened 2015-09-09
CB-9617 Do not restore plugins immediately after plugin removal

vladimir-kotikov:
https://github.com/apache/cordova-lib/commit/14675051c400a6811a6c6171fbf92f3475244630 introduced `cordova.prepare` call after each plugin uninstallation (see [CB-9617](https://issues.apache.org/jira/browse/CB-9617)). This results in a weird behaviour, when removed plugin immediately restored if it was saved in config.xml.

### DIFF
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -31,8 +31,7 @@ var cordova_util  = require('./util'),
     pluginMapper  = require('cordova-registry-mapper').newToOld,
     events        = require('../events'),
     metadata      = require('../plugman/util/metadata'),
-    chainMap      = require('../util/promise-util').Q_chainmap,
-    cordova       = require('./cordova');
+    chainMap      = require('../util/promise-util').Q_chainmap;
 
 // Returns a promise.
 module.exports = function plugin(command, targets, opts) {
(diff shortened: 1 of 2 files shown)


vladimir-kotikov:
@omefire, @TimBarham, please review when possible

omefire:
Will review tomorrow

TimBarham:
It seems like an ok fix for the immediate problem, but it does raise a concern in my mind - before the Platform API changes, we didn't call `prepare` when adding/removing plugins. This meant we didn't call `installPluginsFromConfigXML()` even when *adding* a plugin.

So, I have a question: is it perhaps overkill to call `prepare()` when adding/removing plugins? We end up doing a bunch of stuff we didn't used to do - such as calling `installPluginsFromConfigXML()`, calling `installPlatformsFromConfigXML()`, firing `before_prepare` and `after_prepare` hooks. Maybe other stuff? Is this desirable, or should we only be calling a subset of prepare() functionality when adding/removing plugins?

vladimir-kotikov:
@TimBarham, agree with you, calling a `cordova.prepare` on each plugin add/rm might be excess. Reworked.

cordova-lib/src/cordova/plugin.js
```diff
@@ -283,6 +283,31 @@ module.exports = function plugin(command, targets, opts) {
...
+            locations: {
+                plugins: path.join(projectRoot, 'plugins'),
+                www: cordova_util.projectWww(projectRoot)
+            }
+        };
+
+        return platforms.getPlatformApi(platform).prepare(project);
+    }));
+}
+
```
TimBarham: Logic looks good to me. Just one suggestion - the guts of `preparePlatforms()` is an exact duplicate of code within `prepare.js`, correct? Why not add a `preparePlatform()` method in `prepare.js`, and call that from within the loop here and within the loop in `prepare`?

I guess one argument against that would be that this logic will become pretty simplistic once there is a real `projectInfo` instance for the project, but on the other hand, sharing this code means you don't have to update that in two places. So... I guess up to you :smile:.

vladimir-kotikov: Yup, factored this to separate function. Another thing that i've missed, is that we still need to do a `browserify` if corresponding option is specified.

TimBarham:
:+1:

PR #308, opened 2015-09-11
CB-9521 Do not pass removed platform to `after_platform_rm` hook

vladimir-kotikov:
This fixes annoying issue, already described in [CB-9521](https://issues.apache.org/jira/browse/CB-9521).

### DIFF
--- a/cordova-lib/src/cordova/platform.js
+++ b/cordova-lib/src/cordova/platform.js
@@ -363,6 +363,7 @@ function remove(hooksRunner, projectRoot, targets, opts) {
             platformMetadata.remove(projectRoot, target);
         });
     }).then(function() {
+        opts.cordova.platforms = cordova_util.listPlatforms(opts.projectRoot);
         return hooksRunner.fire('after_platform_rm', opts);
     });
 }


cordova-lib/src/cordova/platform.js
```diff
@@ -363,6 +363,7 @@ function remove(hooksRunner, projectRoot, targets, opts) {
             platformMetadata.remove(projectRoot, target);
         });
     }).then(function() {
+        opts.cordova.platforms = cordova_util.listPlatforms(opts.projectRoot);
```
daserge: @vladimir-kotikov probably the root reason of the issue is in hooksRunner.opts.platforms [being set in platforms.js](https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/cordova/platform.js#L570) and then used in [refineOpts in HooksRunner](https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/hooks/HooksRunner.js#L69) - we will need to refine the way context.opts.cordova.platforms is used and properly document it.

vladimir-kotikov:
Closing this as there is some uncertainty in hook parameters specification.

PR #309, opened 2015-09-16
CB-9598 Introduces `cordova-common` module

vladimir-kotikov:
This is an implementation for [CB-9598](https://issues.apache.org/jira/browse/CB-9598) and the next step of platformApi refactoring. The idea is to move all code that potentially will be shared between different platforms and LIB into a separate module that will be used by both LIB an platforms.

This PR only copies existing code from cordova-lib into cordova-common, but doesn't switches cordova-lib to use it. Once this PR is merged we can remove this functionality from cordova-lib and switch to using cordova-common.


vladimir-kotikov:
@nikhilkh, regarding your concern about iOS specific dependencies (`xcode`, `plist` and `bplist-parser`) for this - the reason is that `ConfigChanges` class still uses them to do some special config processing.

We can get rid of this ios specifics only when we'll end supporing old platforms (`ios` in particular) via `PlatformApiPoly`.

BTW there is exactly the same situation with Windows platform - there is some logic that spreads appxmanifest changes to all appxmanifest files in project (see [generate_plugin_config_munge](https://github.com/MSOpenTech/cordova-lib/commit/9245da38f30e61ab98595e8916aea175f7d88f56#diff-cd932e6542e2f88b0f2c0dd0907b7481R274)), which we'll need to keep there while supporting non-platformApi-compatible platform.

vladimir-kotikov:
@sgrebnov, this is ready to merge, please take a look.

sgrebnov:
sounds great, @vladimir-kotikov - I plan to complete this on weekend

sgrebnov:
:+1: - reviewed `cordova-common` folder, run `npm install` and `npm test`

PR #312, opened 2015-10-01
CB-9569 - Support <access> and <allow-navigation> tag translation to Application Transport Security (ATS) Info.plist directives

shazron:
Tests are included as well. &lt;access&gt; and &lt;allow-navigation&gt; tags have two new attributes to account for ATS: `minimum-tls-version` (defaults to 'TLSv1.2') and `requires-forward-secrecy` (defaults to 'true').

If the defaults are set, there will be *no* corresponding entry in ATS, to reduce verbosity in the Info.plist.

### DIFF
--- a/cordova-lib/spec-cordova/ConfigParser.spec.js
+++ b/cordova-lib/spec-cordova/ConfigParser.spec.js
@@ -219,6 +219,14 @@ describe('config.xml parser', function () {
                 });
                 expect(pluginNames).not.toContain('org.apache.cordova.legacyplugin');
             });
+            it('it should read <access> tag entries', function(){
+                var accesses = cfg.getAccesses();
+                expect(accesses.length).not.toEqual(0);
+            });
+            it('it should read <allow-navigation> tag entries', function(){
+                var navigations = cfg.getAllowNavigations();
+                expect(navigations.length).not.toEqual(0);
+            });
         });
     });
 });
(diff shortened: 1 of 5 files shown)


shazron:
Not sure why the tests are failing (--save flag stuff) which is unrelated to my code :/

vladimir-kotikov:
They fails because of facebook plugin has changed its id: https://github.com/Wizcorp/phonegap-facebook-plugin/commit/a2f6dbb80a54eff29a0e09836aebf0dd905fd60a and we [still use old id](https://github.com/apache/cordova-lib/blob/master/cordova-lib/spec-cordova/save.spec.js#L52) in our tests to check if the plugin's variables saved into config.xml

shazron:
lol, I was just investigating this, and wanted to update this PR comment thread and fixed it in https://issues.apache.org/jira/browse/CB-9737 :)

shazron:
Hmm one problem that I can see is, this assumes the last item to "clobber" the ATS entry is the last one standing. There is no merge strategy. For example:
```
<access origin="https://*.google.com />
<access origin="https://google.com />
```

We would only pick up the last item there, were subdomains are *not* allowed. We can't possibly decipher the intent of the user here, it is up to the user to write a coherent whitelist policy.

tombell:
Is there a way to completely disable this? I don't like the configuration it generates as it nullifies our prior configuration will bad defaults.

shazron:
The default configuration is `<access origin="*" />`. The ideal way is to remove that tag, and it shouldn't alter any ATS entries, but right now it adds it if it is missing (as a default), so that won't work. Will that work better? Do you mind filing an issue at issues.cordova.io so we can track this and you can be notified? 

I would like to hear about the configuration issues you are encountering and provide solutions (including the issue above I outlined).

tombell:
I created https://issues.apache.org/jira/browse/CB-10011

We have an Android platform as well in our project, so we cannot remove the `<access origin="*" />`.

I would rather a way to fully disable this "feature".

PR #313, opened 2015-10-03
CB-9589 auto convert IDs to new style IDs if in registry mapper

stevengill:
@nikhilkh @vladimir-kotikov take a look. Thoughts?

### DIFF
--- a/cordova-lib/src/plugman/install.js
+++ b/cordova-lib/src/plugman/install.js
@@ -36,6 +36,7 @@ var path = require('path'),
     plugman = require('./plugman'),
     HooksRunner = require('../hooks/HooksRunner'),
     isWindows = (os.platform().substr(0,3) === 'win'),
+    pluginMapper = require('cordova-registry-mapper').oldToNew,
     cordovaUtil = require('../cordova/util');
 
 var superspawn = require('../cordova/superspawn');
@@ -48,7 +49,8 @@ var PluginInfoProvider = require('../PluginInfoProvider');
    providing a high-level logic flow overview.
    1. module.exports (installPlugin)
      a) checks that the platform is supported
-     b) invokes possiblyFetch
+     b) converts oldIds into newIds (CPR -> npm)
+     c) invokes possiblyFetch
    2. possiblyFetch
      a) checks that the plugin is fetched. if so, calls runInstall
      b) if not, invokes plugman.fetch, and when done, calls runInstall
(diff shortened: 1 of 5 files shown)


cordova-lib/src/plugman/install.js
```diff
@@ -79,6 +80,19 @@ module.exports = function installPlugin(platform, project_dir, id, plugins_dir,
 
     var current_stack = new action_stack();
 
+    // Split @Version from the plugin id if it exists.
+    var splitVersion = id.split('@');
+    //Check if a mapping exists for the plugin id
+    //if it does, convert id to new name id 
+    var newId = pluginMapper[splitVersion[0]];
+    if(newId) {
+        events.emit('log', 'Notice: ' + id + ' has been automatically converted to ' + newId + ' and fetched from npm. This is due to our old plugins registry shutting down.');
```
vladimir-kotikov: Shouldn't this be emitted with `warn` level to encourage users to switch to new plugin id instead of relying on plugin mapper?

cordova-lib/src/plugman/install.js
```diff
@@ -404,6 +418,20 @@ function installDependencies(install, dependencies, options) {
     return dependencies.reduce(function(soFar, dep) {
         return soFar.then(
             function() {
+                // Split @Version from the plugin id if it exists.
+                var splitVersion = dep.id.split('@');
+                //Check if a mapping exists for the plugin id
+                //if it does, convert id to new name id 
+                var newId = pluginMapper[splitVersion[0]];
+                if(newId) {
+                    events.emit('log', 'Notice: ' + dep.id + ' has been automatically converted to ' + newId + ' and fetched from npm. This is due to our old plugins registry shutting down.');
```
vladimir-kotikov: Shouldn't this be emitted with `warn` level to encourage users to switch to new plugin id instead of relying on plugin mapper?

cordova-lib/src/plugman/uninstall.js
```diff
@@ -142,8 +143,17 @@ module.exports.uninstallPlugin = function(id, plugins_dir, options) {
...
         var deps = pluginInfo.getDependencies();
+        var deps_path;
         deps.forEach(function (d) {
-            if (toDelete.indexOf(d.id) === -1) {
+            var splitVersion = d.id.split('@');
+            deps_path = path.join(plugin_dir, '..', splitVersion[0]);
+            if (!fs.existsSync(deps_path)) {
+                var newId = pluginMapper[splitVersion[0]];
+                if (newId && toDelete.indexOf(newId) === -1) {
+                   toDelete.push(newId);
```
vladimir-kotikov: IMHO it makes sense to add verbose message here to show to user what's happened.

cordova-lib/src/plugman/uninstall.js
```diff
@@ -256,6 +266,15 @@ function runUninstallPlatform(actions, platform, project_dir, plugin_dir, plugin
         promise = promiseutil.Q_chainmap(danglers, function(dangler) {
             var dependent_path = path.join(plugins_dir, dangler);
 
+            //try to convert ID if old-id path doesn't exist. 
+            if (!fs.existsSync(dependent_path)) {
+                var splitVersion = dangler.split('@');
+                var newId = pluginMapper[splitVersion[0]];
+                if(newId) {
+                    dependent_path = path.join(plugins_dir, newId);
```
vladimir-kotikov: Same here.

stevengill:
@vladimir-kotikov switched to warnings and added some verbose output. Also added a conversion step to fetch.js as it was needed in there since cordova-lib calls fetch instead of install. I originally had my ID conversion code in registry.js, but this lead to some failing checks in the code for what ID was expected and what was actually retrieved. i think it is all good now. Let me know if you have other thoughts before I merge.

vladimir-kotikov:
@stevengill, looks good :+1:

stevengill:
merged

PR #314, opened 2015-10-03
Refactor cordova-serve to use Express.

TimBarham:
This simplifies a lot of code, but more importantly provides a more standardized and modular way to customize the server (via Express middleware).

This is a breaking change (so next release should be bumped to `0.2.0`).

#### API Changes:

* Now call module as a constructor that returns a `cordova-serve` instance. This allows creation of multiple servers listening to different ports.
* `launchBrowser()` is a static method of `cordova-serve`
* `servePlatform()` and `launchServer()` are instance methods.
* `sendStream()` method removed (not needed - just use `send` module).
* `opts` parameter loses the following properties: `urlPathHandler`, `streamHandler` and `serverExtender`. These can all be done in other ways (primarily by attaching Express middleware to the `app` property exposed on the `cordova-serve` instance).
* `opts` parameter gains the following properties: `router` (an optional Express router that will be attached to server) and `events` (an optional `EventEmitter` which will be used for logging - via `events.emit('log', msg)` - if provided).


purplecabbage:
How does this impact the cordova-browser platform? Do browser targeting apps now depend on express and node on the server? Or can they be deployed as static sites?

TimBarham:
It only impacts the `cordova run browser` command (which already runs on a local node server) - it has no impact on how users might deploy their app when targeting the browser platform.

purplecabbage:
Great, thanks. 
I love express, just wanted to be sure.

TimBarham:
No problem :smile:

cordova-serve/serve.js
```diff
@@ -17,9 +17,41 @@
...
+        });
+        next();
+    });
+
+    // Turn on compression
+    this.app.use(compression());
+
+    this.servePlatform = require('./src/platform');
+    this.launchServer = server;
+}
```
omefire: +1

cordova-serve/src/server.js
```diff
@@ -17,121 +17,51 @@
...
-
-        function serveFile(filePath) {
-            if (!filePath) {
-                if (!root) {
-                    throw new Error('No server root directory HAS BEEN specified!');
-                }
-                filePath = path.join(root, urlPath);
-            }
+    if (opts.router) {
+        app.use(opts.router);
```
omefire: I wonder if 'middleware' is a more familiar term to express users. WDYT ?

omefire:
LGTM!

cordova-serve/src/server.js
TimBarham: "Router" is definitely a term that express users would be familiar with. While you could pass any middleware in as this option, the more likely scenario is to pass in a router (which is like a "mini-app" - it provides a single entry point for a while collection of middleware).

PR #315, opened 2015-10-05
CB-9747 Fix Android Parser

ogoguel:
Fix parser error if the activity name refers a full package class, instead of just the class
Added tests


stevengill:
We are deleting android_parser.js for cordova@7 since that functionality has all been moved into cordova-android platform now. This would only be useful for older versions of cordova-android (pre-API.js) now and we are dropping support for those older versions in cordova@7. I think this PR can be closed.

PR #316, opened 2015-10-06
CB-9712 CLI 5.3 breaks with node 3.3.3

daserge:
I could repro the issue using npm@3.3.6 as npm@3.3.5 seems to have issues with Cordova installation.

Requires [cordova-app-hello-world#14](https://github.com/apache/cordova-app-hello-world/pull/14) changes in order for require to work.

[Jira issue](https://issues.apache.org/jira/browse/CB-9712)

### DIFF
--- a/cordova-lib/src/cordova/create.js
+++ b/cordova-lib/src/cordova/create.js
@@ -113,7 +113,13 @@ function create(dir, optionalId, optionalName, cfg) {
         cfg.lib.www.url = cfg.lib.www.url || cfg.lib.www.uri;
 
         if (!cfg.lib.www.url) {
-            cfg.lib.www.url = path.join(__dirname, '..', '..', 'node_modules', 'cordova-app-hello-world');
+            try {
+                cfg.lib.www.url = require('cordova-app-hello-world').dirname;
+            } catch (e) {
+                // Falling back on npm@2 path hierarchy
+                // TODO: Remove fallback after cordova-app-hello-world release
+                cfg.lib.www.url = path.join(__dirname, '..', '..', 'node_modules', 'cordova-app-hello-world');
+            }
         }
 
         // TODO (kamrik): extend lazy_load for retrieval without caching to allow net urls for --src.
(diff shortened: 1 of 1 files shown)


daserge:
Tests are passing with the updated cordova-app-hello-world: #317

nikhilkh:
LGTM. @stevengill for FYI. Next tools release will have to also release cordova-app-hello-world to make this work.

stevengill:
LGTM. We should do a new release soon!

nikhilkh:
@daserge Does the dev workflow break with this change? Does `coho npm-link` need to be modified to include linking cordova-app-hello-world for cordova committers who are linking to cloned version of the repo for this to work.

daserge:
@nikhilkh yes I think coho will need an update.
Although I still have issues with local dev deployment specifically with [npm linking](https://github.com/npm/npm/issues/9766) so can't verify this for now.

nikhilkh:
Let's check this in anyway.

omefire:
I don't understand. Why would we need a new release of 'cordova-app-hello-world' for this to work ? This code doesn't seem to require anything extra as it currently stands.

daserge:
This update needs [cordova-hello-world-module to export its path](https://github.com/apache/cordova-app-hello-world/pull/14/files#diff-168726dbe96b3ce427e7fedce31bb0bcR3) as its used in `cfg.lib.www.url` assignment directly. 
Probably I should add a comment to this line for clearness.
Alternatively we can use [require.resolve](https://nodejs.org/api/globals.html#globals_require_resolve) here instead - what do you think?

nikhilkh:
`require.resolve` seems a more elegant way to handle this. As long as that is supported in older versions of node.

daserge:
And actually the reason why this does not work with old `cordova-hello-app-world` is the fact that it did not have an entry point (index.js).

omefire:
@nikhilkh , but us using ```require.resolve``` here doesn't make it necessary to update the hello-world app.

omefire:
@daserge , Aha! that was a question I asked myself. I didn't see any entry point and assumed node.js figures it out via some weird rules.

Can we also update package.json in the hello-world app to explicitly mention the entry point ? (index.js)

daserge:
@omefire, thank you for the review, 
I've addressed it - exposed `cordova-app-hello-world.dirname` explicitly, also included a few places with the similar path issues.

vladimir-kotikov:
LGTM

omefire:
LGTM!

PR #318, opened 2015-10-08
CB-9758 Mobilespec crashes adding plugins on OS X

daserge:
Avoid empty keys in NSExceptionDomains

[Jira issue](https://issues.apache.org/jira/browse/CB-9758)

### DIFF
--- a/cordova-lib/spec-cordova/metadata/ios_parser.spec.js
+++ b/cordova-lib/spec-cordova/metadata/ios_parser.spec.js
@@ -496,6 +496,15 @@ describe('ios project parser', function () {
                     expect(d.NSExceptionRequiresForwardSecrecy).toEqual(false);
                 });
             });
+            it('<allow-navigation> - should ignore wildcards like data:*, https:*, https://*', function(done) {
+                wrapper(p.update_from_config(cfg), done, function() {
+                    var ats = plist_build.mostRecentCall.args[0].NSAppTransportSecurity;
+                    var exceptionDomains = ats.NSExceptionDomains;
+                    expect(exceptionDomains['']).toBeUndefined();
+                    expect(exceptionDomains['null']).toBeUndefined();
+                    expect(exceptionDomains['undefined']).toBeUndefined();
+                });
+            });
         });
         describe('www_dir method', function() {
             it('should return /www', function() {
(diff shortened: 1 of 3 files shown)


daserge:
@shazron, could you please review this?

shazron:
Just a question, this is for the case for example: `http://*` or `https://*` correct? A failing test should be added to the existing tests.

daserge:
The current code results in such an entry in `mobilespec-Info.plist`:
```xml
        <key></key>
        <dict/>
```
This happens with mobilespec `config.xml`' [`<allow-navigation href="data:*" />`](https://github.com/apache/cordova-mobile-spec/blob/78a865153615ae20b9d8b6305b7b1a49496bbbf7/config.xml#L55)(`scheme:*` if generalize) and also for `http://*` or `https://*`.

daserge:
BTW `<allow-navigation href="http:*" />` and `<allow-navigation href="https:*" />` don't crash and result in 
```xml
        <key>null</key>
        <dict>
          <key>NSExceptionAllowsInsecureHTTPLoads</key>
          <true/>
        </dict>
```
and 
```xml
        <key>null</key>
        <dict/>
```
correspondingly.
Not sure if they work in this case though.

shazron:
Great, since you've identified the stuff that crashes, you should add tests for them in the jasmine tests (there are already tests for other cases).

daserge:
Sure, I'm looking into this.

daserge:
@shazron, could please explain the difference in these tests? Is there a typo?
[`it('<access> - should handle wildcard')`](https://github.com/apache/cordova-lib/commit/0c3614e2fff6f3579a9220611ad439abde9abde1#diff-dfa42d6bc7630716b3099bfc4d33da05R191)
[`it('<allow-navigation> - should handle wildcard')`](https://github.com/apache/cordova-lib/commit/0c3614e2fff6f3579a9220611ad439abde9abde1#diff-dfa42d6bc7630716b3099bfc4d33da05R308)

shazron:
No, its correct. The two tags are different.

daserge:
Don't these tests do the same thing?
Or is there any state changing between tests I'm missing?

shazron:
They do the same thing, almost. The only thing different is access tags do not handle wildcards in the schemes. I've opted to keep them separate for completeness, and ease of maintenance. 

Ideally the allow-navigation test should be part of the whitelist plugin, since it is only part of that.

daserge:
Got it, thanks for the explanation, @shazron.

daserge:
@shazron, could you please review again?

shazron:
looks good to me. Thanks @daserge!

PR #320, opened 2015-10-12
CB-9782 Update create/update signatures for PlatformApi polyfill

vladimir-kotikov:
This PR updates parameters set for `PlatformApi.create/update` methods according to https://github.com/apache/cordova-android/pull/226


stevengill:
As far as I can tell, this pretty much fills out the polyfill with real variables. LGTM

PR #322, opened 2015-10-15
CB-9813 Keep module-to-plugin mapping at hand.

busykai:
This information is handy for various use-cases. E.g. for preview application to
provide only the plugins which are used by the application (instead of all of
them, since they are built-in in the preview app on the device). The other is
simulation (taco-simulation does need this info).

Instead of trying to (unreliably) recover this information from module list,
just keep track of it when it's known (at the prepare stage).

CC: @tony--

### DIFF
--- a/cordova-lib/src/platforms/PlatformApiPoly.js
+++ b/cordova-lib/src/platforms/PlatformApiPoly.js
@@ -566,7 +566,8 @@ PlatformApiPoly.prototype._addModulesInfo = function(plugin, targetDir) {
         var moduleName = plugin.id + '.' + ( moduleToInstall.name || moduleToInstall.src.match(/([^\/]+)\.js/)[1] );
         var obj = {
             file: ['plugins', plugin.id, moduleToInstall.src].join('/'),
-            id: moduleName
+            id: moduleName,
+            pluginId: plugin.id
         };
         if (moduleToInstall.clobbers.length > 0) {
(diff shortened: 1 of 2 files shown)


cordova-lib/src/platforms/PlatformApiPoly.js
```diff
@@ -566,7 +566,8 @@ PlatformApiPoly.prototype._addModulesInfo = function(plugin, targetDir) {
         var moduleName = plugin.id + '.' + ( moduleToInstall.name || moduleToInstall.src.match(/([^\/]+)\.js/)[1] );
         var obj = {
             file: ['plugins', plugin.id, moduleToInstall.src].join('/'),
-            id: moduleName
+            id: moduleName,
+            plugin: plugin.id
```
TimBarham: `pluginId` perhaps?

busykai: yes, `pluginId` is more descriptive.

TimBarham:
LGTM, @busykai.

tony--:
[JIRA issue CB-9813](https://issues.apache.org/jira/browse/CB-9813)

busykai:
Thanks, Tony! Updated the commit message.

TimBarham:
Thanks @tony-- - @busykai, could you update the PR title to include the JIRA issue?

PR #324, opened 2015-10-16
CB-9815 Engine name="cordova" should check tools version, not platforms

sgrebnov:
https://issues.apache.org/jira/browse/CB-9815

### DIFF
--- a/cordova-lib/spec-plugman/install.spec.js
+++ b/cordova-lib/spec-plugman/install.spec.js
@@ -255,7 +255,7 @@ describe('install', function() {
                 install('android', project, plugins['com.cordova.engine'])
                 .fail(fail)
                 .fin(function () {
-                    expect(satisfies).toHaveBeenCalledWith('2.5.0','>=2.3.0');
+                    expect(satisfies).toHaveBeenCalledWith('2.5.0','>=1.0.0');
                     done();
                 });
             });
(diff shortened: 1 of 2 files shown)


TimBarham:
This gives us the correct behavior, but are there backwards compatibility issues (is it likely any plugin is using this expecting the current behavior)?

sgrebnov:
@TimBarham I see your point, but all places where I saw that engine name usage all devs thought they check cordova version - see examples below. Taking into account that we will apply this change only in next/future cordova version where we have pinned platforms versions lower than cordova version itself  non of existing plugins will be broken.

Also, I see that this is not covere in docs at all so we may want to contribute some info to Plugin Dev Guide 

InAppBrowser plugin
```
    <engines>
      <engine name="cordova" version=">=3.1.0" /><!-- Needs cordova/urlutil -->
    </engines>
```
phonegap-facebook-plugin
```
    <engines>
        <!-- Requires > 3.3.* because of the custom Framework tag for iOS [CB-5238] -->
        <!-- Requires > 3.5.0 because of the custom Framework tag for Android [CB-6698] -->
        <engine name="cordova" version=">=3.5.0" />
    </engines>
```

TimBarham:
Yeah, sounds reasonable. :+1:

nikhilkh:
LGTM

PR #327, opened 2015-10-20
On Windows, verify browsers installed before launching.

TimBarham:
Do this to avoid dialog that pops up on Windows if browser is not installed, and provide more meaningful error messages.

### DIFF
--- a/cordova-serve/src/browser.js
+++ b/cordova-serve/src/browser.js
@@ -17,8 +17,14 @@
  under the License.
  */
 
-var exec = require('./exec'),
+var child_process = require('child_process'),
+    exec = require('./exec'),
+    fs = require('fs'),
+    path = require('path'),
     Q = require('q');
+
+var NOT_INSTALLED = 'The browser target is not installed: %target%';
+var NOT_SUPPORTED = 'The browser target is not supported: %target%';
 
 /**
  * Launches the specified browser with the given URL.
@@ -33,6 +39,7 @@ module.exports = function (opts) {
     var target = opts.target || 'chrome';
     var url = opts.url || '';
 
+    target = target.toLowerCase();
     return getBrowser(target, opts.dataDir).then(function (browser) {
         var args;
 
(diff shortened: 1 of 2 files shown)


cordova-serve/src/browser.js
```diff
@@ -17,9 +17,15 @@
...
  */
 
-var exec = require('./exec'),
+var child_process = require('child_process'),
+    exec = require('./exec'),
+    fs = require('fs'),
+    path = require('path'),
     Q = require('q');
 
+var NOT_INSALLED = 'The browser target is not installed: %target%';
```
vladimir-kotikov: typo: `NOT_INSTALLED`

cordova-serve/src/browser.js
```diff
@@ -72,7 +79,10 @@ module.exports = function (opts) {
             args.push(url);
         }
         var command = args.join(' ');
-        return exec(command);
+        return exec(command).catch(function (error) {
+            // Assume any error means that the browser is not installed and display that as a more friendly error.
+            throw new Error(NOT_INSALLED.replace('%target%', target));
```
vladimir-kotikov: typo again

cordova-serve/src/browser.js
TimBarham: Thanks. Oops. Fixed :smile:

cordova-serve/src/browser.js
```diff
@@ -95,16 +105,93 @@ function getBrowser(target, dataDir) {
...
+    var promise = target === 'edge' ? edgeSupported() : browserInstalled(browser);
+    return promise.catch(function (error) {
+        return Q.reject((error && error.toString() || NOT_INSTALLED).replace('%target%', target));
+    });
+}
+
+function edgeSupported() {
+    var d = Q.defer();
+
+    child_process.exec('ver', function (err, stdout, stderr) {
```
vladimir-kotikov: Is there any chance to reuse `./exec` module here. Looks like it contains almost the same logic.

vladimir-kotikov: BTW, `os.version()` gives me exactly the same version as `ver`, it might better to use it

TimBarham: I originally used the exec module and the logic actuallly ended up simpler this way.

TimBarham: os.release() doesn't recognize Win10 in fairly common versions of node.

vladimir-kotikov:
Looks good.

TimBarham:
Thanks for taking a look, @vladimir-kotikov.

PR #328, opened 2015-10-21
CB-9736 - Extra main activity generated when an android package name specified

gorkem:
Removes the old package name after the new one is created. If the package
name is different


stevengill:
This functionality is moving to the platforms themselves. For this fix to work in android@5.0+, you should look to add it to https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/lib/prepare.js#L156

gorkem:
@stevengill Yes, created CB-9844 to follow up. This one is still valid for fixing older android engines though.

stevengill:
Definitely. Thanks!

PR #332, opened 2015-10-28
CB-9872 Fixed save.spec.11 failure

alsorokin:
This is a fix for https://issues.apache.org/jira/browse/CB-9872

### DIFF
--- a/cordova-lib/spec-cordova/save.spec.js
+++ b/cordova-lib/spec-cordova/save.spec.js
@@ -30,34 +30,49 @@ describe('(save flag)', function () {
         PlatformApi = require('../src/platforms/PlatformApiPoly'),
         platform    = rewire('../src/cordova/platform');
 
-    var appName             = 'testApp',
-        tempPath            = path.join(__dirname, 'temp'),
-        appPath             = path.join(tempPath, appName),
-        platformName        = helpers.testPlatform,
-        platformVersionOld  = '4.0.0',
-        platformVersionNew  = '4.0.1',
-        platformLocalPath   = path.join(__dirname, 'cordova-' + platformName),
-        platformGitUrl      = 'https://github.com/apache/cordova-' + platformName,
-        platformGitRef      = '4.0.x',
-        platformTgzUrl      = 'https://git-wip-us.apache.org/repos/asf?p=cordova-' + platformName + '.git;a=snapshot;h=' + platformVersionNew + ';sf=tgz',
-        otherPlatformName   = 'windows',
-        otherPlatformSpec   = '4.0.0',
-        pluginName          = 'cordova-plugin-console',
-        pluginVersion       = '1.0.0',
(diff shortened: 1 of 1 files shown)


TimBarham:
So Alex - what was the underlying problem?

cordova-lib/spec-cordova/save.spec.js
```diff
@@ -244,19 +307,26 @@ describe('(save flag)', function () {
...
+            mockDownloadPlatform(platformLocalPathOld, platformVersionOld);
+
+            platform('add', platformName + '@' + platformVersionOld)
             .then(function () {
+                revertDownloadPlatform();
                 var fsExistsSync = fs.existsSync.bind(fs);
                 spyOn(fs, 'existsSync').andCallFake(function (somePath) {
                     return (somePath === path.join(appPath, 'platforms', platformName)) || fsExistsSync(somePath);
                 });
+                mockDownloadPlatform(platformLocalPathNew, platformVersionNew);
```
TimBarham: Using edge for this test has caught 2 `cordova-android` issues in the last week or so. While ideally we don't want `cordova-lib` tests failing because of `cordova-android` issues, should we continue using edge in these tests until we've added `update` tests to `cordova-android`?

alsorokin:
We were using master version of cordova-android platform and after recent changes which introduce new Api, mocks in these tests became insufficient. So basically, we now use android version 4.0.0, 4.0.1 and 4.1.0 for these tests instead of the master version.

cordova-lib/spec-cordova/save.spec.js
alsorokin: I believe I'll be able to catch platform update issues running my usual manual test cases which I do on every platform release.

stevengill:
I don't like the idea of testing 4.1.0 instead of master. Is this something we will have to update everytime we release a newer version? 

What is it about master that broke the tests?

TimBarham:
@stevengill  - ultimately, the point of these tests is to test the `--save` feature, so going forward I don't think we want them to be at risk of being broken by platform changes in master, as long as we add tests to the platforms that cover the e2e scenarios that these tests were "accidentally" covering.

cordova-lib/spec-cordova/save.spec.js
TimBarham: I think we want to have tests for these scenarios so we can verify things aren't being broken at PR/commit time rather than waiting until we're about to do a release.

stevengill:
Sounds good.

vladimir-kotikov:
We've scheduled adding e2e `update` tests for android and refactoring these tests to abstract from real platforms as two next high-pri tasks, so for now I propose to merge this in as temporary solution to unblock release and make tests passing.

Thoughts?

TimBarham:
:+1:

stevengill:
Merge! That way I can do the tools release.

PR #333, opened 2015-10-28
Dummy PR to trigger a test run - don't merge :)

TimBarham:


vladimir-kotikov:
@timbarham, I think we need to merge https://github.com/apache/cordova-lib/pull/332 first to get green badges - apart from android issue there are some problems with tests.

TimBarham:
Yep, saw that PR just after I sent this one :smile:

PR #334, opened 2015-10-29
CB-9901 cordova plugin search opens in a browser

eymorale:


nikhilkh:
LGTM this is a good one to fix. There is a jshint failure though that you might want to fix.

eymorale:
Thanks @nikhilkh! I'll take a look at the jshint failure and fix it

omefire:
What about people who use Cordova-lib from gulp instead of the cli ?
Does it make sense to pop open a browser window ?

eymorale:
@omefire I've started a discussion thread for this on the dev list

TimBarham:
Are people likely to be doing a plugin search from a gulp script?

csantanapr:
I think is a nice feature to open the browser to a good spot on the wen to search for plugins applicable  for cli use cases where people are running by hand when they get started to use the CLI. not applicable to unattended. The current form is just a mess of a UI/UX.

For scripting or unattended like gulp scripts I guess they can use npmsearh same way we are doing for the website plugin search.

stevengill:
Lets merge this in after the 5.4.0 release

eymorale:
sounds good. thanks @stevengill

PR #335, opened 2015-10-29
CB-9902 - cordova run --list broken

shazron:


stevengill:
LGTM

spawn or targets hasn't changed so maybe options coming into targets has changed since cordova-common. I'll look at the history to see whats up but we should get this in so I can release.

PR #337, opened 2015-10-29
Fix create argument prerequisite check

ZombieProtectionAgency:
Store argument.length from create before return call
Update returns prerequisite checks to use the previously stored arg length


omefire:
LGTM!

PR #340, opened 2015-11-09
CB-9981 path.parse only available on node 0.12+

apla:


vladimir-kotikov:
Merged. Thanks, @apla

PR #341, opened 2015-11-11
use plist@1.2 to be able to use cordova with node@0.12

rubensayshi:
`plist@1.1` had `xmlbuilder@2.2` as dependency, which had `engine@<=0.10`.
with `plist@1.2` it now depends on `xmlbuilder@4.4`, which no longer has that restriction.

test needed to be updated and version of plist set to minimum of 1 because `xmlbuilder@4.4` does empty tags as '<string />' instead of '<string></string>'.

alternatively if for w/e reason you don't want this change you're going to have to change the `plist` dependency for `cordova-common` to `~1.1.0` to avoid `xmlbuilder@4.4` because that will make your test fail!
it's also an option to just ask `xmlbuilder` to release another `2.2.x` without the engine restriction, because it seems to work fine, but it upgrading seems like a better idea to me.


vladimir-kotikov:
@rubensayshi, thanks for reporting and providing a fix

PR #343, opened 2015-11-14
add missing package_suffix function on amazon-fireos platform for plu…

paulhovey:
This came about while trying to install https://github.com/katzer/cordova-plugin-badge to an amazon-fireos platform.  I was getting the error that module.exports.package_suffix was undefined, so I tracked it down to the plugman amazon-fireos.js file and added the same logic as in android.js .


omefire:
I think the reason why that function is needed is because of this: https://github.com/apache/cordova-lib/blob/cc1ab409f64e16a06b5b495dd7b5bdfac37acda7/cordova-lib/src/plugman/platforms/android.js#L40

omefire:
LGTM!

PR #349, opened 2015-12-02
Cb 9590 - Ubuntu support for the new plugin naming convention

david-barth-canonical:
Re-opening the pull request with hopefully a better fix that can handle all situations.


cordova-lib/src/plugman/platforms/ubuntu.js
```diff
@@ -29,6 +29,54 @@ function toCamelCase(str) {
...
+    } else {
+	match = plugin_id.match(/cordova\-plugin\-([\w\-]+)$/);
+        if (match && match.length > 0)
+	    class_name = match[0].substr(15);
+	else
+            class_name = toCamelCase(class_name);
+    }
+
+    return class_name;
+}
```
TimBarham: Thanks @david-barth-canonical! Some comments:

* Generally the Cordova codebase uses curly braces even around single line blocks.
* There are some tabs in there that should be converted to spaces.
* The `match` variable needs a `var`.
* Should be calling `toCamelCase()` for the scenario where we match `cordova-plugin-...`
* Where we don't match `cordova-plugin-...`, `class_name` is not defined (should be using `plugin_id`).
* Since we will always end up calling `toCamelCase()`, the logic for the second half of this method can be simplified to:

```js
    // fallback to guess work, based on the plugin package name
    if (plugin_id.match(/\.[^.]+$/)) {
        // old-style plugin name
        class_name = plugin_id.match(/\.[^.]+$/)[0].substr(1);
    } else {
        var match = plugin_id.match(/cordova\-plugin\-([\w\-]+)$/);
        if (match && match.length > 0) {
            class_name = match[0].substr(15);
        } else {
            class_name = plugin_id;
        }
    }

    return toCamelCase(class_name);
```

Also, since the original PR has been merged, can you recreate this as a new PR which will just include the changes in your last commit (and changes in response to feedback, of course).

david-barth-canonical: There is a new PR at https://github.com/apache/cordova-lib/pull/349 with
your suggested fixes. Thanks for the review.

TimBarham:
Thanks - LGTM!

PR #351, opened 2015-12-03
CB-10125: Android build fails on read-only image files

jasongin:
Ensure image files are writeable before trying to delete them.

I considered an alternative: ensuring files are writeable after copying them. But that seemed less robust, and other parts of the build are already able to overwrite read-only files due to the use of shell.cp('-f', ...). The only reason this code dealing with image files had a problem was because it deleted the files before re-copying them.


omefire:
Where is this method called from ? cordova-android ?

jasongin:
@omefire It is called during the prepare phase when building android. See more info including stack trace in the CB-10125 JIRA issue.

omefire:
Thanks @jasongin, found the references. I was using github to perform search, I now realize it was a bad idea.

omefire:
LGTM!

PR #353, opened 2015-12-08
CB-8455 added --nohooks option

bso-intel:
implemented --nohooks noption as described in CB-8455

### DIFF
--- a/cordova-lib/src/hooks/HooksRunner.js
+++ b/cordova-lib/src/hooks/HooksRunner.js
@@ -44,6 +44,12 @@ function HooksRunner(projectRoot) {
  * Returns a promise.
  */
 HooksRunner.prototype.fire = function fire(hook, opts) {
+    if (isHookDisabled(opts, hook)) {
+        /* jshint -W064 */
+        return Q();
+        /* jshint +W064 */
+    }
+
     // args check
     if (!hook) {
         throw new Error('hook type is not specified');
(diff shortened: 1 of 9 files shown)


purplecabbage:
This feature has much deeper and far reaching requirements, as the CI has pointed out.
The individual platforms are all keeping their own parsers, and because this affects the platform-api ALL platforms must be updated for this to work.

bso-intel:
Hi Jesse,

The CI failures actually were caused by not checking the existence of the opts parameter.
In the spec unit tests, update_project() was invoked without the opts parameter that I added in windows_parser.js and wp8_parser.js.
I have fixed the issues, and now all tests passed.

vladimir-kotikov:
Isn't hooks is a CLI/LIB feature? As far as I remember, hooks fired by LIB except for `pre_package` hook for wp8/windows, so probably there is no need to pass the option down to platforms.

bso-intel:
Hi Vladmir,

Exactly.
The pre-package hooks on for the windows* platforms run when LIB prepares for the source.
To suppress this hook, HookRunner needs the option "--nohooks pre_package" that is propagated from the command line.

bso-intel:
HI Jesse and Vladimir,
If there is no more comment, could you pull my branch CB-8455 so the next Cordova release contains this feature?
Thanks.

cordova-lib/src/hooks/HooksRunner.js
```diff
@@ -87,6 +93,12 @@ module.exports = HooksRunner;
  */
 module.exports.fire = globalFire;
 function globalFire(hook, opts) {
+    if (isHookDisabled(opts, hook)) {
+        /* jshint -W064 */
```
vladimir-kotikov: Just a minor nit: do we need for this ignores? We don't have neither 'strict mode' nor JSHint `newcap` option enabled, so this is probably just a no-op.

vladimir-kotikov:
Looks good.

bso-intel:
Hi Vladimir,

Thank you very much for reviewing.
I addressed your feedback and added several unit tests in LIB and CLI.

bso-intel:
Thank you very much for merging, Vladimir.

PR #356, opened 2015-12-21
CB-10193 Add deprecation notice about pre_package removal

vladimir-kotikov:
This PR adds an temporary polyfill for windows 'pre_package' event to keep compatibility with plugins, which uses this hook. It also adds a notice about 'pre_package' hook removal in next windows platform versions.

For more details on this see [corresponding mailing list discussion](http://apache.markmail.org/thread/enzm34cktoeqh4ss)


sgrebnov:
:+1: lgtm

PR #357, opened 2015-12-21
CB-10235 Added clearer error message for info command

ktop:
When using the cordova info command, if Android isn't set up properly, then the error message printed out isn't too helpful. I added to the error message to tell users to make sure they have the path of Android SDK tools and platform-tools in their PATH environment variable.


ktop:
I will redo this pull request. I was having some repository problems so I'll just start over.

PR #358, opened 2015-12-23
Add cli_variables to plugman before_plugin_install hook params

mattdsteele:
Currently the context provided to the `before_plugin_install` hook does not include plugin variables/parameters. This is problematic for my use case, where I bundle a hook with the plugin that performs string manipulation based on plugin variables a user defines.

It's possible to get default values (if they are defined) in a hook using the `context.opts.plugin.pluginInfo.getPreferences()`, but this does not account for `cli_options`, or those defined in an plugin through `config.xml` in the `<variable>` element.

This change adds a `variables` key to the hook's context, which contains the CLI (or default) plugin variables. A hook could access them through the `context.opts.plugin.variables` object.

### DIFF
--- a/cordova-lib/spec-cordova/HooksRunner.spec.js
+++ b/cordova-lib/spec-cordova/HooksRunner.spec.js
@@ -450,6 +450,14 @@ describe('HooksRunner', function() {
                                 delete call.args[1].plugin.pluginInfo._et;
                             }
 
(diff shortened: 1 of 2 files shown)


stevengill:
In the upcoming cordova@7 release, plugin variables will be saved in your project's package.json. You should be able to require package.json in your hook and have access to them.

PR #359, opened 2015-12-29
CB-10274 Honor --www parameter in plugman.

bradleyjames:


cordova-lib/src/plugman/install.js
```diff
@@ -596,11 +596,14 @@ function handleInstall(actions, pluginInfo, platform, project_dir, plugins_dir,
     events.emit('verbose', 'Install start for "' + pluginInfo.id + '" on ' + platform + '.');
 
     options.variables = filtered_variables;
-    // Set up platform to install asset files/js modules to <platform>/platform_www dir
-    // instead of <platform>/www. This is required since on each prepare platform's www dir is changed
-    // and files from 'platform_www' merged into 'www'. Thus we need to persist these
-    // files platform_www directory, so they'll be applied to www on each prepare.
-    options.usePlatformWww = true;
+    if (!('www_dir' in options) || options.www_dir === undefined) {
```
vladimir-kotikov: There is some inconsistency in how this option should be specified: the [plugman's README](https://github.com/apache/cordova-plugman/blob/master/README.md#plugin-management) states that there is `--www` option for plugman cli, [comments in cordova-lib code](https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/plugman/install.js#L68) metntion `www_dir` option, but there is no code, reponsible for conversion `--www` -> `www_dir`

vladimir-kotikov:
IMO, better fix this by moving 'usePlatformWww' option setup to cordova-cli area of responsibility - somewhere around [src/cordova/plugin.js#L180](https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/cordova/plugin.js#L180) - this way you won't need to specify any options to plugman to get desired result, but cordova-cli will continue to install plugins' JS files to `platform_www`

cordova-lib/src/plugman/install.js
bradleyjames: The conversation from --www -> www_dir happens here:

https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/plugman/plugman.js#L107

vladimir-kotikov: Oops, missed that...

bradleyjames:
Seems like that will change the default install location for plugman (when not called from the cordova CLI) from platform_www to www.  Is that what you're wanting?

vladimir-kotikov:
IMO, when you're developing an app for one platform and _following the non-CLI workflow_ with plugman, 'www' folder is the expected target location for plugin's JS files, not 'platform_www'. Since you're not using CLI, there is no need in intermediate 'platform_www' directory.

PR #360, opened 2015-12-30
[CB-10274] Make www directory the default for plugman.

bradleyjames:
New patch that addresses comments made here: https://github.com/apache/cordova-lib/pull/359.

### DIFF
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -186,6 +186,14 @@ module.exports = function plugin(command, targets, opts) {
                                 pluginInfoProvider: pluginInfoProvider
                             };
 
+                            if (!('www_dir' in options) || options.www_dir === undefined) {
+                                // Set up platform to install asset files/js modules to <platform>/platform_www dir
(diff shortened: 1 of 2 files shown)


bradleyjames:
Noticed that uninstall suffers from the same issue.  I'll address that at some point in the future (if I find the time and need).

cordova-lib/src/cordova/plugin.js
```diff
@@ -186,6 +186,14 @@ module.exports = function plugin(command, targets, opts) {
                                 pluginInfoProvider: pluginInfoProvider
                             };
 
+                            if (!('www_dir' in options) || options.www_dir === undefined) {
```
vladimir-kotikov: IMO it is safe to remove this condition and always set `options.usePlatformWww = true` because this code won't be called by plugman and CLI doesn't know anything about `www_dir` option.

vladimir-kotikov:
@stevengill, what do you think about this?

stevengill:
I agree that plugman shouldn't be installing to platform_www so this logic belongs more to cli workflow

LGTM

vladimir-kotikov:
Thanks @stevengill, @bradleyjames. Merged.

PR #363, opened 2016-01-23
New plugin version selection implementation

riknoll:
@stevengill @dblotsky please review. This is an implementation for the plugin version selection scheme that Dmitry and I proposed in [this discuss PR](https://github.com/cordova/cordova-discuss/pull/30). I'd appreciate some feedback! I have some docs written up for this too, I'm just holding off on that PR just yet.

### DIFF
--- a/cordova-lib/spec-cordova/plugin_fetch.spec.js
+++ b/cordova-lib/spec-cordova/plugin_fetch.spec.js
@@ -79,7 +79,7 @@ function checkUnmetRequirements(requirements) {
 
     warnings.forEach(function(warning) {
         var extracted = UNMET_REQ_REGEX.exec(warning);
-        if(extracted) {
+        if (extracted) {
             reqWarnings.push({
                 dependency: extracted[1],
                 installed: extracted[2],
@@ -161,7 +161,12 @@ describe('plugin fetching version selection', function(done) {
         // Jasmine will print it out if the matcher fails
         this.addMatchers({
             toContain: function(check, expected) {
-                return this.actual.find(check);
+                for(var i = 0; i < this.actual.length; i++) {
+                    if (check(this.actual[i])) {
+                        return true;
+                    }
+                }
+                return false;
             }
         });
 
--- a/cordova-lib/spec-cordova/plugin_fetch.spec.js
+++ b/cordova-lib/spec-cordova/plugin_fetch.spec.js
@@ -115,12 +115,12 @@ describe('plugin fetching version selection', function(done) {
         testEngineWithProject(done, testEngine, null);
     });
 
-    it('should ignore upperbounds if no version constraints are given', function(done) {
+    it('should handle upperbounds if no single version constraints are given', function(done) {
         var testEngine = {
             '<1.0.0': { 'cordova-android': '<2.0.0' }
         };
 
-        testEngineWithProject(done, testEngine, null);
+        testEngineWithProject(done, testEngine, '2.3.0');
     });
 
     it('should apply upper bounds greater than highest version', function(done) {
--- a/cordova-lib/spec-cordova/plugin_fetch.spec.js
+++ b/cordova-lib/spec-cordova/plugin_fetch.spec.js
@@ -61,6 +61,8 @@ events.on('warn', function(warning) {
 function testEngineWithProject(done, testEngine, testResult) {
     plugin.getFetchVersion(project,
         {
+            'version': '2.3.0',
+            'name': 'test-plugin',
             'engines': { 'cordovaDependencies': testEngine },
             'versions': testPluginVersions
         }, cordovaVersion)
@@ -359,7 +361,11 @@ describe('plugin fetching version selection', function(done) {
     });
 
     it('should not fail if there is no engine in the npm info', function(done) {
-        plugin.getFetchVersion(project, { versions: testPluginVersions }, cordovaVersion)
+        plugin.getFetchVersion(project, {
+                version: '2.3.0',
+                name: 'test-plugin',
+                versions: testPluginVersions
+            }, cordovaVersion)
         .then(function(toFetch) {
             expect(toFetch).toBe(null);
         })
@@ -370,6 +376,8 @@ describe('plugin fetching version selection', function(done) {
         var after = getWarningCheckCallback(done, []);
 
         plugin.getFetchVersion(project, {
+                version: '2.3.0',
+                name: 'test-plugin',
                 versions: testPluginVersions,
                 engines: {
                     'node': '>7.0.0',
--- a/cordova-lib/spec-cordova/plugin_fetch.spec.js
+++ b/cordova-lib/spec-cordova/plugin_fetch.spec.js
@@ -20,10 +20,10 @@
 var plugin  = require('../src/cordova/plugin'),
     helpers = require('./helpers'),
     path    = require('path'),
+    events = require('cordova-common').events,
     shell   = require('shelljs');
 
 var testPluginVersions = [
-    '0.0.0',
     '0.0.2',
     '0.7.0',
     '1.0.0',
@@ -44,7 +44,20 @@ var tempDir = helpers.tmpDir('plugin_fetch_spec');
 var project = path.join(tempDir, 'project');
 
 var getVersionErrorCallback;
+var warnings = [];
 
+// Used to extract the constraint, the installed version, and the required
+// semver range from a warning message
+var UNMET_REQ_REGEX = /\s+([^\s]+)[^\d]+(\d+\.\d+\.\d+) installed, (.+) required\)/;
+
+// We generate warnings when we don't fetch latest. Collect them to make sure we
+// are making the correct warnings
+events.on('warn', function(warning) {
+    warnings.push(warning);
+});
+
+// Tests a sample engine against the installed platforms/plugins in our test
+// project
 function testEngineWithProject(done, testEngine, testResult) {
     plugin.getFetchVersion(project,
         {
@@ -56,6 +69,69 @@ function testEngineWithProject(done, testEngine, testResult) {
     })
     .fail(getVersionErrorCallback)
     .fin(done);
+}
+
+// Checks the warnings that were printed by the CLI to ensure that the code is
+// listing the correct reasons for failure. Checks against the global warnings
+// object which is reset before each test
+function checkUnmetRequirements(requirements) {
+    var reqWarnings = [];
+
+    warnings.forEach(function(warning) {
+        var extracted = UNMET_REQ_REGEX.exec(warning);
+        if(extracted) {
+            reqWarnings.push({
+                dependency: extracted[1],
+                installed: extracted[2],
+                required: extracted[3]
+            });
+        }
+    });
+
+    expect(reqWarnings.length).toEqual(requirements.length);
+
+    requirements.forEach(function(requirement) {
+        expect(reqWarnings).toContain(function(extractedWarning) {
+            return  extractedWarning.dependency === requirement.dependency.trim() &&
+                    extractedWarning.installed  === requirement.installed.trim() &&
+                    extractedWarning.required   === requirement.required.trim();
+        }, requirement);
+    });
+}
+
+// Helper functions for creating the requirements objects taken by
+// checkUnmetRequirements()
+function getPlatformRequirement(requirement) {
+    return {
+        dependency: 'cordova-android',
+        installed: '3.1.0',
+        required: requirement
+    };
+}
+
+function getCordovaRequirement(requirement) {
+    return {
+        dependency: 'cordova',
+        installed: cordovaVersion,
+        required: requirement
+    };
+}
+
+function getPluginRequirement(requirement) {
+    return {
+        dependency: 'ca.filmaj.AndroidPlugin',
+        installed: '4.2.0',
+        required: requirement
+    };
+}
+
+// Generates a callback that checks warning messages after the test is complete
+function getWarningCheckCallback(done, requirements) {
+    return function() {
+        checkUnmetRequirements(requirements);
+        expect(getVersionErrorCallback).not.toHaveBeenCalled();
+        done();
+    };
 }
 
 function createTestProject() {
@@ -80,14 +156,20 @@ describe('plugin fetching version selection', function(done) {
     createTestProject();
 
     beforeEach(function() {
+        // Adding a matcher for checking the array of warning messages so that
+        // we can have meanigful error messages. Expected is passed because
+        // Jasmine will print it out if the matcher fails
+        this.addMatchers({
+            toContain: function(check, expected) {
+                return this.actual.find(check);
+            }
+        });
+
+        warnings = [];
         getVersionErrorCallback = jasmine.createSpy('unexpectedPluginFetchErrorCallback');
     });
 
-    afterEach(function() {
-        expect(getVersionErrorCallback).not.toHaveBeenCalled();
-    });
-
-    it('should properly handle a mix of upper bounds and single versions', function() {
+    it('should handle a mix of upper bounds and single versions', function(done) {
         var testEngine = {
             '0.0.0' : { 'cordova-android': '1.0.0' },
             '0.0.2' : { 'cordova-android': '>1.0.0' },
@@ -98,10 +180,14 @@ describe('plugin fetching version selection', function(done) {
             '2.3.0' : { 'cordova-android': '6.0.0' }
         };
 
-        testEngineWithProject(done, testEngine, '1.3.0');
+        var after = getWarningCheckCallback(done, [
+            getPlatformRequirement('6.0.0')
+        ]);
+
+        testEngineWithProject(after, testEngine, '1.3.0');
     });
 
-    it('should properly apply upper bound engine constraints', function(done) {
+    it('should apply upper bound engine constraints when there are no unspecified constraints above the upper bound', function(done) {
         var testEngine = {
             '1.0.0' : { 'cordova-android': '>2.0.0' },
             '1.7.0' : { 'cordova-android': '>4.0.0' },
@@ -112,7 +198,49 @@ describe('plugin fetching version selection', function(done) {
             '2.3.0' : { 'cordova-android': '6.0.0' }
         };
 
-        testEngineWithProject(done, testEngine, null);
+        var after = getWarningCheckCallback(done, [
+            getPlatformRequirement('6.0.0')
+        ]);
+
+        testEngineWithProject(after, testEngine, null);
+    });
+
+    it('should apply upper bound engine constraints when there are unspecified constraints above the upper bound', function(done) {
+        var testEngine = {
+            '0.0.0' : {},
+            '2.0.0' : { 'cordova-android': '~5.0.0' },
+            '<1.0.0': { 'cordova-android': '>5.0.0' }
+        };
+
+        var after = getWarningCheckCallback(done, [
+            getPlatformRequirement('~5.0.0')
+        ]);
+
+        testEngineWithProject(after, testEngine, '1.7.1');
+    });
+
+    it('should handle the case where there are no constraints for earliest releases', function(done) {
+        var testEngine = {
+            '1.0.0' : { 'cordova-android': '~5.0.0' }
+        };
+
+        var after = getWarningCheckCallback(done, [
+            getPlatformRequirement('~5.0.0')
+        ]);
+
+        testEngineWithProject(after, testEngine, '0.7.0');
+    });
+
+    it('should handle the case where the lowest version is unsatisfied', function(done) {
+        var testEngine = {
+            '0.0.2' : { 'cordova-android': '~5.0.0' }
+        };
+
+        var after = getWarningCheckCallback(done, [
+            getPlatformRequirement('~5.0.0')
+        ]);
+
+        testEngineWithProject(after, testEngine, null);
     });
 
     it('should handle upperbounds if no single version constraints are given', function(done) {
@@ -120,7 +248,9 @@ describe('plugin fetching version selection', function(done) {
             '<1.0.0': { 'cordova-android': '<2.0.0' }
         };
 
-        testEngineWithProject(done, testEngine, '2.3.0');
+        var after = getWarningCheckCallback(done, []);
+
+        testEngineWithProject(after, testEngine, '2.3.0');
     });
 
     it('should apply upper bounds greater than highest version', function(done) {
@@ -129,7 +259,11 @@ describe('plugin fetching version selection', function(done) {
             '<5.0.0': { 'cordova-android': '<2.0.0' }
         };
 
-        testEngineWithProject(done, testEngine, null);
+        var after = getWarningCheckCallback(done, [
+            getPlatformRequirement('<2.0.0')
+        ]);
+
+        testEngineWithProject(after, testEngine, null);
     });
 
     it('should treat empty constraints as satisfied', function(done) {
@@ -138,21 +272,29 @@ describe('plugin fetching version selection', function(done) {
             '1.1.0' : { 'cordova-android': '>5.0.0' }
         };
 
-        testEngineWithProject(done, testEngine, '1.0.0');
+        var after = getWarningCheckCallback(done, [
+            getPlatformRequirement('>5.0.0')
+        ]);
+
+        testEngineWithProject(after, testEngine, '1.0.0');
     });
 
-    it('should treat an empty engine as not satisfied', function(done) {
+    it('should treat an empty cordovaDependencies as satisfied', function(done) {
         var testEngine = {};
 
-        testEngineWithProject(done, testEngine, null);
+        var after = getWarningCheckCallback(done, []);
+
+        testEngineWithProject(after, testEngine, '2.3.0');
     });
 
-    it('should treat a badly formatted semver range as not satisfied', function(done) {
+    it('should ignore a badly formatted semver range', function(done) {
         var testEngine = {
             '1.1.3' : { 'cordova-android': 'badSemverRange' }
         };
 
-        testEngineWithProject(done, testEngine, null);
+        var after = getWarningCheckCallback(done, []);
+
+        testEngineWithProject(after, testEngine, '2.3.0');
     });
 
     it('should respect unreleased versions in constraints', function(done) {
@@ -162,7 +304,11 @@ describe('plugin fetching version selection', function(done) {
             '1.3.0' : { 'cordova-android': '6.0.0' }
         };
 
-        testEngineWithProject(done, testEngine, '1.1.0');
+        var after = getWarningCheckCallback(done, [
+            getPlatformRequirement('6.0.0')
+        ]);
+
+        testEngineWithProject(after, testEngine, '1.1.0');
     });
 
     it('should respect plugin constraints', function(done) {
@@ -172,7 +318,11 @@ describe('plugin fetching version selection', function(done) {
             '2.3.0' : { 'ca.filmaj.AndroidPlugin': '6.0.0' }
         };
 
-        testEngineWithProject(done, testEngine, '2.0.0');
+        var after = getWarningCheckCallback(done, [
+            getPluginRequirement('6.0.0')
+        ]);
+
+        testEngineWithProject(after, testEngine, '2.0.0');
     });
 
     it('should respect cordova constraints', function(done) {
@@ -182,7 +332,11 @@ describe('plugin fetching version selection', function(done) {
             '2.3.0' : { 'cordova': '6.0.0' }
         };
 
-        testEngineWithProject(done, testEngine, '1.1.0');
+        var after = getWarningCheckCallback(done, [
+            getCordovaRequirement('6.0.0')
+        ]);
+
+        testEngineWithProject(after, testEngine, '1.1.0');
     });
 
     it('should not include pre-release versions', function(done) {
@@ -191,8 +345,12 @@ describe('plugin fetching version selection', function(done) {
             '2.0.0' : { 'cordova-android': '>5.0.0' }
         };
 
+        var after = getWarningCheckCallback(done, [
+            getPlatformRequirement('>5.0.0')
+        ]);
+
         // Should not return 2.0.0-rc.2
-        testEngineWithProject(done, testEngine, '1.7.1');
+        testEngineWithProject(after, testEngine, '1.7.1');
     });
 
     it('should not fail if there is no engine in the npm info', function(done) {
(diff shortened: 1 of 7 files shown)


stevengill:
Hey @riknoll,

Thanks for doing this! I'll review it after the cordova 6 release.

cordova-lib/src/cordova/plugin.js
```diff
@@ -122,25 +123,6 @@ module.exports = function plugin(command, targets, opts) {
...
-                                } else {
-                                    //fetch pinned version from cordova-lib
-                                    var pinnedVer = pkgJson.cordovaPlugins[id];
-                                    target = pinnedVer ? (id + '@' + pinnedVer) : target;
-                                }
-                            }
-                        }
-
                         // Fetch the plugin first.
                         events.emit('verbose', 'Calling plugman.fetch on plugin "' + target + '"');
```
dblotsky: Might `target` be undefined here?

cordova-lib/src/cordova/plugin.js
```diff
@@ -440,6 +451,13 @@ function list(projectRoot, hooksRunner, opts) {
     });
 }
 
+function getInstalledPlugins(projectRoot) {
+    var pluginsDir = path.join(projectRoot, 'plugins');
+    // TODO: This should list based off of platform.json, not directories within plugins/
```
dblotsky: Is there a JIRA for this TODO?

cordova-lib/src/cordova/plugin.js
```diff
@@ -507,3 +525,117 @@ function versionString(version) {
...
+                return allVersions[allVersions.length - 1];
+            }
+        }
+    }
+
+    // No constraints were satisfied
+    return null;
+}
+
+function checkProjectRequirements(reqs, pluginMap, platformMap, cordovaVersion) {
```
dblotsky: Nitpick: maybe rename to something that sounds more like a question.

cordova-lib/src/cordova/plugin.js
riknoll: Not my TODO. Moved from old code

cordova-lib/src/cordova/plugin.js
riknoll: Yup, this line should have been moved down with the rest

riknoll:
@stevengill just checking in now that Cordova 6.0.0 is released. Let me know if you have any feedback.

riknoll:
Rebased to master and responded to PR feedback. @TimBarham @vladimir-kotikov can you take a look at this as well?

cordova-lib/src/cordova/platform.js
```diff
@@ -492,17 +492,13 @@ function list(hooksRunner, projectRoot, opts) {
     var platforms_on_fs = cordova_util.listPlatforms(projectRoot);
     return hooksRunner.fire('before_platform_ls', opts)
```
TimBarham: You don't need `platforms_on_fs` anymore, since you get equivalent information from your call to `getInstalledPlatformsWithVersions()` (so you can use `platformMap` below instead of `platforms_on_fs`).

cordova-lib/src/cordova/util.js
```diff
@@ -185,6 +187,31 @@ function listPlatforms(project_dir) {
...
+            };
+        });
+    })).then(function(platformVersions) {
+        var result = {};
+        for(var i = 0; i < platformVersions.length; i++) {
+            result[platformVersions[i].platform] = platformVersions[i].version;
+        }
+        return result;
+    });
+}
```
TimBarham: The above method would be simplified if you just accumulated the `result` object as you go... As in... initialize `result` before the `Q.all()` call, then instead of, for example, `return { platform: p, version: v ? v : null };` you would just do `result[p] = v || null;`. In the final then, you would ignore the parameter and just return `result`.

cordova-lib/src/cordova/util.js
```diff
@@ -185,6 +187,31 @@ function listPlatforms(project_dir) {
...
 
+function getInstalledPlatformsWithVersions(project_dir) {
+    var platforms_on_fs = listPlatforms(project_dir);
+
+    return Q.all(platforms_on_fs.map(function(p) {
+        return superspawn.maybeSpawn(path.join(project_dir, 'platforms', p, 'cordova', 'version'), [], { chmod: true })
+        .then(function(v) {
+            return {
+                platform: p,
+                version: v ? v : null
```
TimBarham: Minor nit, but `v ? v : null` can be simplified to `v || null`.

cordova-lib/src/cordova/plugin.js
```diff
@@ -154,6 +135,38 @@ module.exports = function plugin(command, targets, opts) {
                             is_top_level: true
                         };
 
+                        // If no version is specified, retrieve the version (or source) from config.xml
+                        if (!version && !cordova_util.isUrl(id) && !cordova_util.isDirectory(id)) {
+                            events.emit('verbose', 'no version specified, retrieving version from config.xml');
+                            var ver = getVersionFromConfigFile(id, cfg);
```
TimBarham: I know this is just moved code, but while you're here could you make that sentence start with an uppercase "N" (as in "No version specified...")? :smile:

cordova-lib/src/cordova/plugin.js
```diff
@@ -154,6 +135,38 @@ module.exports = function plugin(command, targets, opts) {
...
+                                    // Fallback to pinned version if available
+                                    fetchVersion = fetchVersion ? fetchVersion : pkgJson.cordovaPlugins[id];
+                                    target = fetchVersion ? (id + '@' + fetchVersion) : target;
+
+                                    events.emit('verbose', 'Calling plugman.fetch on plugin "' + target + '"');
+                                    return plugman.raw.fetch(target, pluginPath, fetchOptions);
+                                })
+                                .then(function (directory) {
+                                    return pluginInfoProvider.get(directory);
+                                });
```
TimBarham: We don't want this code duplicating the code below. It's only a couple of lines now, but it leaves us open to problems later on if changes are made to what we do here. Instead, I'd suggest moving this whole `if` block (that is, `if (!version && !cordova_util.isUrl(id) && !cordova_util.isDirectory(id)) { ... }`) to a separate method that returns a promise for a version. Then you can have a single `plugman.raw.fetch()` and `pluginInfoProvider.get()` regardless of whether the operation to determine the version is sync or async.

cordova-lib/src/cordova/plugin.js
```diff
@@ -154,6 +135,38 @@ module.exports = function plugin(command, targets, opts) {
...
+                            } else {
+                                // If no version is given at all, We need to decide what version to
+                                // fetch based on the current project
+                                return registry.info([id])
+                                .then(function(pluginInfo) {
+                                    return getFetchVersion(projectRoot, pluginInfo, pkgJson.version);
+                                })
+                                .then(function(fetchVersion) {
+                                    // Fallback to pinned version if available
+                                    fetchVersion = fetchVersion ? fetchVersion : pkgJson.cordovaPlugins[id];
```
TimBarham: Simplify to `fetchVersion = fetchVersion || pkgJson.cordovaPlugins[id];` :smile:

cordova-lib/src/cordova/plugin.js
```diff
@@ -154,6 +135,38 @@ module.exports = function plugin(command, targets, opts) {
...
+                        // If no version is specified, retrieve the version (or source) from config.xml
+                        if (!version && !cordova_util.isUrl(id) && !cordova_util.isDirectory(id)) {
+                            events.emit('verbose', 'no version specified, retrieving version from config.xml');
+                            var ver = getVersionFromConfigFile(id, cfg);
+
+                            if (cordova_util.isUrl(ver) || cordova_util.isDirectory(ver)) {
+                                target = ver;
+                            } else if (ver) {
+                                // If version exists in config.xml, use that
+                                target = ver ? (id + '@' + ver) : target;
```
TimBarham: I know this isn't your code - but would be good to clean this up while we're here... We already know `ver` has a value (since that is the condition of the `if` statement), so this should just be `target = id + '@' + ver`.

cordova-lib/src/cordova/plugin.js
```diff
@@ -507,3 +527,117 @@ function versionString(version) {
...
+ */
+function getFetchVersion(projectRoot, pluginInfo, cordovaVersion) {
+    // Figure out the project requirements
+    if (pluginInfo.engines && pluginInfo.engines.cordovaDependencies) {
+        var pluginList = getInstalledPlugins(projectRoot);
+        var pluginMap = {};
+
+        for(var i = 0; i < pluginList.length; i++) {
+            pluginMap[pluginList[i].id] = pluginList[i].version;
+        }
```
TimBarham: Minor nit, but this would be cleaner if you used `pluginList.forEach()` (or even `pluginList.reduce()`).

cordova-lib/src/cordova/plugin.js
```diff
@@ -507,3 +527,117 @@ function versionString(version) {
...
+                pluginInfo.engines.cordovaDependencies,
+                pluginMap,
+                platformVersions,
+                cordovaVersion);
+        });
+    } else {
+        // If we have no engine, we want to fall back to the default behavior
+        return Q.fcall(function() {
+            return null;
+        });
```
TimBarham: I think here you just need `return Q(null);` (returns a promise fulfilled with `null`).

cordova-lib/src/cordova/plugin.js
```diff
@@ -507,3 +527,117 @@ function versionString(version) {
...
+ *                                  result of calling `registry.info()`)
+ * @param {string}  cordovaVersion  The semver version of cordova-lib
+ *
+ * @return {Promise}                A promise that will resolve to either a string
+ *                                  if there is a version of the plugin that this
+ *                                  project satisfies or null if there is not
+ */
+function getFetchVersion(projectRoot, pluginInfo, cordovaVersion) {
+    // Figure out the project requirements
+    if (pluginInfo.engines && pluginInfo.engines.cordovaDependencies) {
```
TimBarham: Could you document in the comments what we expect `cordovaDependencies` to look like? It would make this code a lot easier to understand.

cordova-lib/src/cordova/plugin.js
```diff
@@ -507,3 +527,117 @@ function versionString(version) {
...
+    for(var version in engine) {
+        if(version.indexOf('<') === 0 && semver.valid(version.substring(1))) {
+            // We only care about upper bounds that our project does not support
+            if(!satisfiesProjectRequirements(engine[version], pluginMap, platformMap, cordovaVersion)) {
+                upperBounds.push(version);
+            }
+        } else if(semver.valid(version) && allVersions.indexOf(version) !== -1) {
+            versions.push(version);
+        }
+    }
```
TimBarham: So do we only support version ranges in the form of `<x.y.z` and specific versions? I was thinking it might be handy for a plugin dev to specify other ranges. For example, `^1.0.0` to identify requirements for all 1.x releases. That might actually be very easy to handle - `semver.maxSatisfying(versions, range)` will take an array of versions (which we have) and a range, and return the highest version from the list that matches the range. If you do this to all ranges (including `<x.y.z` type ranges), could you avoid the special case `upperBounds` handling? Any range could be converted to a single version (the maximum actual version that matches the range) - could that then be handled like any other entry?

cordova-lib/src/cordova/plugin.js
```diff
@@ -507,3 +527,117 @@ function versionString(version) {
...
+        }
+
+        // Check the version constraint
+        if(satisfiesProjectRequirements(engine[versions[i]], pluginMap, platformMap, cordovaVersion)) {
+            // Because we sorted in descending order, we can stop searching once
+            // we hit a satisfied constraint
+            var index = versions.indexOf(versions[i]);
+            if(index > 0) {
+                // Return the highest plugin version below the next highest
+                // constrainted version
```
TimBarham: Typo ("constrained").

cordova-lib/src/cordova/plugin.js
```diff
@@ -507,3 +527,117 @@ function versionString(version) {
...
+                return allVersions[allVersions.indexOf(nextHighest) - 1];
+            } else {
+                // Return the latest plugin version
+                return allVersions[allVersions.length - 1];
+            }
+        }
+    }
+
+    // No constraints were satisfied
+    return null;
```
TimBarham: Weren't we going to display a warning in this scenario?

cordova-lib/src/cordova/plugin.js
TimBarham: `getFetchVersion()` could return null in two scenarios:
* The plugin didn't specify any constraints.
* The plugin specifies constraints and we fail to meet them.

It seems we really should be able to differentiate between these two scenarios. For example:
* Should we only fall back on pinned version if the plugin doesn't define constraints?
* If the plugin does define constraints and we fail to meet them, should we display a warning that we're picking a plugin version that might not work?

cordova-lib/src/cordova/plugin.js
```diff
@@ -507,3 +527,117 @@ function versionString(version) {
...
+        } else if(req.indexOf('cordova-') === 0) {
+            // Might be a platform constraint
+            var platform = req.substring(8);
+            if(platformMap[platform]) {
+                okay = semver.satisfies(platformMap[platform], reqs[req]);
+            }
+        }
+
+        if(!okay) {
+            return false;
```
TimBarham: If we reject a version, it might be useful to output a message describing the version we rejected an why (with a log level of `verbose` of course).

cordova-lib/src/cordova/plugin.js
```diff
@@ -507,3 +527,117 @@ function versionString(version) {
...
+function satisfiesProjectRequirements(reqs, pluginMap, platformMap, cordovaVersion) {
+    for (var req in reqs) {
+        var okay = true;
+
+        if(pluginMap[req]) {
+            okay = semver.satisfies(pluginMap[req], reqs[req]);
+        } else if(req === 'cordova') {
+            okay = semver.satisfies(cordovaVersion, reqs[req]);
+        } else if(req.indexOf('cordova-') === 0) {
+            // Might be a platform constraint
```
TimBarham: Should we also allow platform names without `cordova-`?

riknoll: I originally had it that way, but decided on requiring the `cordova-` prefix to avoid ambiguity with the actual platforms. That is, Android OS version versus cordova-android version.

riknoll:
@TimBarham thanks for the review; I'll update the PR in a bit!

riknoll:
Responded to most the feedback. Also added warnings and verbose logging which I completely forgot in my original PR. I will rebase this branch soon

riknoll:
Rebased to master (that was fun)

cordova-lib/spec-cordova/plugin_fetch.spec.js
```diff
@@ -0,0 +1,203 @@
...
+        .fail(function(err) {
+            console.log(err);
+            expect(true).toBe(false);
+        }).fin(done);
+    });
+
+    it('clean up after plugin fetch spec', function() {
+        removeTestProject();
+    });
+});
```
dblotsky: This is really pretty code.

riknoll: Thanks!

cordova-lib/src/cordova/plugin.js
TimBarham: Sure, that's reasonable. And now that I think about it, consistent with the `engines` tag in `plugin.xml`.

riknoll:
@TimBarham addressed your other feedback

cordova-lib/src/cordova/util.js
```diff
@@ -185,6 +187,22 @@ function listPlatforms(project_dir) {
...
+function getInstalledPlatformsWithVersions(project_dir) {
+    var result = {};
+    var platforms_on_fs = listPlatforms(project_dir);
+
+    return Q.all(platforms_on_fs.map(function(p) {
+        return superspawn.maybeSpawn(path.join(project_dir, 'platforms', p, 'cordova', 'version'), [], { chmod: true })
+        .then(function(v) {
+            result[p] = v || null;
+        }, function(v) {
+            result[p] = v || 'broken';
```
TimBarham: Is this what you intended? The parameter `v` here will actually be the error. In the previous version, `v` was simply ignored (so "version" was always set to "broken").

TimBarham:
Other than one small remaining question, looks great! Thanks @riknoll!

cordova-lib/src/cordova/util.js
riknoll: Good catch! Fixed

riknoll:
Fixed jasmine tests and created a JIRA for this ([CB-10679](https://issues.apache.org/jira/browse/CB-10679))

riknoll:
After conversation with @nikhilkh, I reworked the code a bit so that the warnings it prints are more actionable. They now list what dependencies failed for the latest version of the plugin.

riknoll:
@TimBarham could you take a quick look at the updates when you get a chance?

cordova-lib/src/cordova/plugin.js
```diff
@@ -512,3 +535,175 @@ function versionString(version) {
...
+}
+
+function determinePluginVersionToFetch(allVersions, engine, pluginMap, platformMap, cordovaVersion) {
+    var versions = [];
+    var latest = allVersions[allVersions.length - 1];
+    var upperBound = null;
+    var upperBoundRange = null;
+
+    for(var version in engine) {
+        if(version.indexOf('<') === 0 && semver.valid(version.substring(1))) {
```
nikhilkh: It might be a good idea to be more robust with whitespaces here.

nikhilkh:
@vladimir-kotikov to also take a look.

riknoll:
I have significant changes to push in response to some feedback, so don't review yet. I'll comment when they're in.

riknoll:
Alright, this should be ready for review. The changes I just pushed included a few fixes to edge cases, better handling of malformed input/whitespace, and a lot more tests.

nikhilkh:
@TimBarham @vladimir-kotikov Can you please take a look in the next couple of days? It will be good to get this committed and do a release.

cordova-lib/src/cordova/plugin.js
```diff
@@ -305,6 +289,50 @@ module.exports = function plugin(command, targets, opts) {
     });
 };
 
+function determinePluginTarget(projectRoot, cfg, target, fetchOptions) {
+    var parts = target.split('@');
+    var id = parts[0];
+    var version = parts[1];
+
+    // If no version is specified, retrieve the version (or source) from config.xml
+    if (!version && !cordova_util.isUrl(id) && !cordova_util.isDirectory(id)) {
```
vladimir-kotikov: I'd rather reverse `if` condition here to return `target` immediately if there is no need for any processing. This is just a matter of taste of course but IMO it would help to improve code readability a bit. The same for L302 and L304

cordova-lib/src/cordova/plugin.js
```diff
@@ -305,6 +289,50 @@ module.exports = function plugin(command, targets, opts) {
...
+            var shouldUseNpmInfo = !fetchOptions.searchpath && !fetchOptions.noregistry;
+
+            if(shouldUseNpmInfo) {
+                events.emit('verbose', 'No version given in config.xml, attempting to use plugin engine info');
+            }
+
+            return (shouldUseNpmInfo ? registry.info([id]) : Q({}))
+            .then(function(pluginInfo) {
+                return getFetchVersion(projectRoot, pluginInfo, pkgJson.version);
+            }, function(error) {
```
vladimir-kotikov: Should we catch the here if the only purpose is to rethrow?

vladimir-kotikov:
LGTM apart from a couple of nitpicks.

cordova-lib/src/cordova/plugin.js
riknoll: Sure, I can change it

cordova-lib/src/cordova/plugin.js
riknoll: Right, I just refactored this code, but I should have caught that. Changing now.

cordova-lib/src/cordova/plugin.js
```diff
@@ -512,3 +543,219 @@ function versionString(version) {
...
+                }
+            }
+
+            // There is no req to merge it with
+            latestFailedReqs.push(failedReq);
+        });
+    }
+
+    listUnmetRequirements(latestFailedReqs);
+    events.emit('warn', 'Current project does not satisfy the engine requirements specified by any version of this plugin. Fetching latest or pinned version of plugin anyway (may be incompatible)');
```
stevengill: Glad we fetch anyways!

stevengill:
LGTM! Great work @riknoll! Very clean code and is easy to follow. Looking forward to switching over to this.

cordova-lib/src/cordova/plugin.js
```diff
@@ -305,6 +289,49 @@ module.exports = function plugin(command, targets, opts) {
...
 
+function determinePluginTarget(projectRoot, cfg, target, fetchOptions) {
+    var parts = target.split('@');
+    var id = parts[0];
+    var version = parts[1];
+
+    // If no version is specified, retrieve the version (or source) from config.xml
+    if (version || cordova_util.isUrl(id) || cordova_util.isDirectory(id)) {
+        return Q(target);
+    } else {
```
TimBarham: Minor nit: superfluous `else` - remove it, and decrease the indentation level of the rest of the function.

cordova-lib/src/cordova/plugin.js
```diff
@@ -305,6 +289,49 @@ module.exports = function plugin(command, targets, opts) {
...
+        events.emit('verbose', 'No version specified, retrieving version from config.xml');
+        var ver = getVersionFromConfigFile(id, cfg);
+
+        if (cordova_util.isUrl(ver) || cordova_util.isDirectory(ver)) {
+            return Q(ver);
+        } else if (ver) {
```
TimBarham: Same deal - superfluous `else`.

cordova-lib/src/cordova/plugin.js
```diff
@@ -305,6 +289,49 @@ module.exports = function plugin(command, targets, opts) {
...
+            // If version exists in config.xml, use that
+            return Q(id + '@' + ver);
+        } else {
```
TimBarham: And again :smile:

cordova-lib/src/cordova/plugin.js
```diff
@@ -305,6 +289,49 @@ module.exports = function plugin(command, targets, opts) {
     });
 };
 
+function determinePluginTarget(projectRoot, cfg, target, fetchOptions) {
+    var parts = target.split('@');
+    var id = parts[0];
+    var version = parts[1];
+
+    // If no version is specified, retrieve the version (or source) from config.xml
```
TimBarham: This comment should be moved down (just above the `events.emit` line, perhaps?)

cordova-lib/src/cordova/plugin.js
```diff
@@ -305,6 +289,49 @@ module.exports = function plugin(command, targets, opts) {
...
+            // If no version is given at all and we are fetching from npm, we
+            // can attempt to use the Cordova dependencies the plugin lists in
+            // their package.json
+            var shouldUseNpmInfo = !fetchOptions.searchpath && !fetchOptions.noregistry;
+
+            if(shouldUseNpmInfo) {
+                events.emit('verbose', 'No version given in config.xml, attempting to use plugin engine info');
+            }
+
+            return (shouldUseNpmInfo ? registry.info([id]) : Q({}))
```
TimBarham: This feels a little hokey to me. Perhaps it would be cleaner to pass the plugin `id` and `shouldUseNpmInfo` to `getFetchVersion()`, and let it call `registry.info()` if appropriate. Of course, then you'd have to mock `registry` to test this. So maybe not worth the extra effort since we want to get this change in :smile:. I'll leave it up to you.

cordova-lib/src/cordova/plugin.js
```diff
@@ -512,3 +543,219 @@ function versionString(version) {
...
+                    required: reqs[req].trim()
+                });
+            }
+        }
+    }
+
+    return failed;
+}
+
+function listUnmetRequirements(failedRequirments) {
```
TimBarham: Typo: 'failedRequir**e**ments'

cordova-lib/src/cordova/plugin.js
```diff
@@ -512,3 +543,219 @@ function versionString(version) {
...
+    events.emit('warn', 'Unmet project requirements for latest version of plugin:');
```
TimBarham: Will there always be enough context in our console output to be clear what plugin we're referring to here (for example where we are installing multiple plugins at once - restoring saved plugins or installing dependencies)?

cordova-lib/src/cordova/plugin.js
```diff
@@ -512,3 +543,219 @@ function versionString(version) {
...
+                    var maxMatchingUpperBound = cleanedRange.substring(1);
+                    if (maxMatchingUpperBound && (!upperBound || semver.gt(maxMatchingUpperBound, upperBound))) {
+                        upperBound = maxMatchingUpperBound;
+                        upperBoundRange = version;
+                    }
+                }
+            }
+        }
+    }
+
```
TimBarham: At this point, if `versions` is empty and we have no `upperBound`, should we perhaps just immediately `return null` (since that means there were no valid requirements specified)?

cordova-lib/src/cordova/plugin.js
```diff
@@ -512,3 +543,219 @@ function versionString(version) {
...
+    // Filters out pre-release versions
+    var latest = semver.maxSatisfying(allVersions, '>=0.0.0');
+
+    var versions = [];
+    var upperBound = null;
+    var upperBoundRange = null;
+
+    for(var version in engine) {
+        if(semver.valid(semver.clean(version)) && !semver.gt(version, latest)) {
+            versions.push(version);
```
TimBarham: It would be useful to output a message if we encounter an invalid version (with log level `verbose`) - this would be helpful particularly for plugin authors debugging their engines entries :smile:.

cordova-lib/src/cordova/plugin.js
```diff
@@ -512,3 +543,219 @@ function versionString(version) {
...
+            if(cleanedRange && UPPER_BOUND_REGEX.exec(cleanedRange)) {
+                // We only care about the highest upper bound that our project does not support
+                if(getFailedRequirements(engine[version], pluginMap, platformMap, cordovaVersion).length !== 0) {
+                    var maxMatchingUpperBound = cleanedRange.substring(1);
+                    if (maxMatchingUpperBound && (!upperBound || semver.gt(maxMatchingUpperBound, upperBound))) {
+                        upperBound = maxMatchingUpperBound;
+                        upperBoundRange = version;
+                    }
+                }
+            }
```
TimBarham: It would be useful to output a message (with log level `verbose`) if we are ignoring an entry because it either isn't a valid version or range, or is a valid range (according to semver), but something other than an "upper bound". This would be helpful particularly for plugin authors debugging their `engines` entries :smile:.

cordova-lib/src/cordova/plugin.js
```diff
@@ -512,3 +543,219 @@ function versionString(version) {
...
+
+    // If the upper bound is greater than latest, we need to combine its engine
+    // requirements with latest to print out in the warning
+    if(upperBound && semver.satisfies(latest, upperBoundRange)) {
+        var upperFailedReqs = getFailedRequirements(engine[upperBoundRange], pluginMap, platformMap, cordovaVersion);
+        upperFailedReqs.forEach(function(failedReq) {
+            for(var i = 0; i < latestFailedReqs.length; i++) {
+                if(latestFailedReqs[i].dependency === failedReq.dependency) {
+                    // Not going to overcomplicate things and actually merge the ranges
+                    latestFailedReqs[i].required = latestFailedReqs[i].required + ' AND ' + failedReq.required;
```
TimBarham: `latestFailedReqs[i].required += ' AND ' + failedReq.required;`

TimBarham:
This is looking great **@rikroll**! I have a few comments, but mostly pretty minor stuff.

cordova-lib/src/cordova/plugin.js
riknoll: Actually, it should return `latest` in that case. We return `null` from this function only if every version of the plugin is constrained in some way and we fail all of them (the test cases were updated to reflect that behavior). The old behavior was weird because it didn't handle versions that are below the lowest constraint, but I think it is more intuitive to assume that if a version is not constrained then it is satisfied.

cordova-lib/src/cordova/plugin.js
riknoll: I hadn't considered the case of plugins being added in bulk. I might need to do a little testing on that, so I'll investigate. In any case, it is easy enough to add the id to this message so I will do so. Nice catch!

cordova-lib/src/cordova/plugin.js
riknoll: Sure!

cordova-lib/src/cordova/plugin.js
riknoll: After looking into this, dependent plugins are not subject to any version selection code because of how cordova-lib is written. Latest will always be fetched  (pinned versions won't be taken into account either). Plugins restored with `cordova prepare` are, so we should indeed print out the id in the message. The dependent plugin thing is bad, but I'm going to leave that as a separate issue since it is not a regression. For now, plugin authors should be advised to give versions for their plugin dependencies.

TimBarham: "I'm going to leave that as a separate issue" - agreed.

cordova-lib/src/cordova/plugin.js
riknoll: Right! I wrote this before I was aware that jasmine could mock function calls. As you said, I want to get this in today so I'm going to leave it out for now since that would be a big change to the tests..

riknoll:
Okay, I added a lot of verbose logging and responded to some of the refactor stuff @TimBarham mentioned (except where noted). I plan to rebase this down to one commit and merge at end of day today.

cordova-lib/src/cordova/plugin.js
TimBarham: Yep, that's fine.

cordova-lib/src/cordova/plugin.js
TimBarham: Hmmm... In `getFetchVersion()`, if no `cordovaDependencies` are defined, we return null and display the following message (verbose log level):

    No plugin engine info found or not using registry, falling back to latest or pinned version

If we get to this in point in `determinePluginVersionToFetch()` with an empty versions array and no `upperBound`, then that is essentially equivalent to there having been no `cordovaDependencies` in the first place, right?

riknoll: Right, they are essentially equivalent and will fetch the same version in both scenarios (except for pinned plugins, but that is a different matter). The only difference will be in that verbose message, which I am okay with because I think that there is a distinction between giving an empty `cordovaDependencies` and ignoring the feature. One says "I have no dependencies" and the other doesn't really say anything. I think it is unnecessary to return the special case of `null`, because returning latest better matches the expected behavior of this function. I can still make it return here if that helps clarify the code. I just think it should return latest rather than `null`.

TimBarham: "except for pinned plugins, but that is a different matter" - I think that's my point. The current behavior seems inconsistent to me:

* If there's no `cordovaRequirements`, we fall back on pinned.
* If there **is** a `cordovaRequirements`, but we don't match any requirements, we fall back on pinned.
* If there **is** a `cordovaRequirements`, but it is effectively empty, we always get latest.

I don't understand why the latter case should be an exception.

riknoll: The way I envisioned it, the following cases are semantically the same:

```
cordovaDependencies: {}

cordovaDependencies: {
    "0.0.0": {}
}
```

However, I can see where that might be confusing. I can also see the case where we might want to add this object to the package.json template and not have it automatically apply for plugins that have it. This is not an issue for any non-core plugins so I'm fine with changing the behavior.

riknoll:
@TimBarham updated

PR #366, opened 2016-01-26
CB-10432 Adds e2e test to protect against future regressions

vladimir-kotikov:
This PR adds test to cover the regression, reported in [CB-10432](https://issues.apache.org/jira/browse/CB-10432)


nikhilkh:
Thanks, Vlad! LGTM.

PR #367, opened 2016-01-29
CB-10459: cordova platform list should mark amazon-fireos and wp8 as deprecated

sarangan12:
Added depracted information to wp8 and amazon-fireos

@nikhilkh Can you please review and merge this PR?


sarangan12:
I will open a new PR after squashing the commits into a single one

PR #369, opened 2016-02-01
CB-10052 Expose child process' io streams via promise progress notification

vladimir-kotikov:
This implements [CB-10052](https://issues.apache.org/jira/browse/CB-10052).

The `superspawn.spawn` promise now emits notifications, which caller can subscribe to using `progress` method:

```
superspawn.spawn('adb', ['devices'])
.progress(function(data){
    if (data.stderr) console.error('"adb devices" raised an error: ' + data.stderr);
})
.then(function(devices){
    // Do something...
})
```

Background for this feature can be found at https://github.com/apache/cordova-android/pull/240

### DIFF
--- a/cordova-common/src/superspawn.js
+++ b/cordova-common/src/superspawn.js
@@ -52,15 +52,39 @@ function maybeQuote(a) {
     return a;
 }
 
-// opts:
-//   printCommand: Whether to log the command (default: false)
-//   stdio: 'default' is to capture output and returning it as a string to success (same as exec)
-//          'ignore' means don't bother capturing it
-//          'inherit' means pipe the input & output. This is required for anything that prompts.
-//   env: Map of extra environment variables.
-//   cwd: Working directory for the command.
-//   chmod: If truthy, will attempt to set the execute bit before executing on non-Windows platforms.
-// Returns a promise that succeeds only for return code = 0.
+/**
+ * A special implementation for child_process.spawn that handles
+ *   Windows-specific issues with batch files and spaces in paths. Returns a
+ *   promise that succeeds only for return code 0. It is also possible to
+ *   subscribe on spawned process' stdout and stderr streams using progress
+ *   handler for resultant promise.
+ *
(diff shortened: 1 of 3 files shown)


cordova-common/src/superspawn.js
```diff
@@ -113,11 +139,15 @@ exports.spawn = function(cmd, args, opts) {
         child.stdout.setEncoding('utf8');
         child.stdout.on('data', function(data) {
             capturedOut += data;
+            d.notify({'stdout': data});
```
dblotsky: What is `d`?

dblotsky:
Why not return the stdout and stderr objects on which calling code can subscribe to the `"data"` events directly?

cordova-common/src/superspawn.js
vladimir-kotikov: `d` is a `Deferred` object. See https://github.com/apache/cordova-lib/blob/master/cordova-common/src/superspawn.js#L68

vladimir-kotikov:
Returning data instead of raw eventEmitter is IMO just less complex for caller API, because you, as a caller, do not need to create handlers, subscribe, unsubscribe, etc.

dblotsky:
@vladimir-kotikov you still need to subscribe a callback, just using `progress` instead of `on("data")`. However using `progress` creates a *new* API, while using `on` leverages a well-established existing API. I'd recommend just reusing what's already known by users of Node APIs.

vladimir-kotikov:
@dblotsky, it seems that didn't really understand your idea. Could you please clarify, where do you propose to return `stdout` and `stderr` from?

dblotsky:
Ah, it seems like I'm the one that's confused. I didn't know that `progress` was a standard promise function. Returning `child` is what I meant, but that would basically require us to stop using promises. I see how this works now: please disregard my comment.

vladimir-kotikov:
@dblotsky, is this good to go?

dblotsky:
Yep, LGTM.

PR #370, opened 2016-02-01
CB-10430 Allow to modify events source

vladimir-kotikov:
This fixes [CB-10430](https://issues.apache.org/jira/browse/CB-10430).

This change is needed to allow other tools to replace default `EventEmitter` instance (e.g. platform might to replace default `EventEmitter` with another one, provided by caller, in order to pass these events to upstream code)

### DIFF
--- a/cordova-common/cordova-common.js
+++ b/cordova-common/cordova-common.js
@@ -19,9 +19,10 @@
 
 /* jshint node:true */
 
+var events = require('./src/events');
+
 // For now expose plugman and cordova just as they were in the old repos
 exports = module.exports = {
-    events: require('./src/events'),
     superspawn: require('./src/superspawn'),
 
     ActionStack: require('./src/ActionStack'),
@@ -40,3 +41,12 @@ exports = module.exports = {
 
     xmlHelpers: require('./src/util/xml-helpers')
 };
+
+Object.defineProperty(module.exports, 'events', {
+    get: function () {
+        return events.get();
+    },
+    set: function (value) {
+        return events.set(value);
+    }
+});
--- a/cordova-common/spec/events.spec.js
+++ b/cordova-common/spec/events.spec.js
@@ -1,0 +1,76 @@
+/**
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+var EventEmitter = require('events').EventEmitter;
+var common = require('../cordova-common');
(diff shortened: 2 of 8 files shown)


dblotsky:
Overriding an implementation's logging with your own sounds like a roundabout way to get what you need. This sounds like it's exposing internal events for the purposes of external debugging, which is best solved by using a debugger, not with a custom logging mechanism.

purplecabbage:
Yeah, I too am not entirely behind this.
It seems like an anti-pattern here, and a consequence of require('../events') only being executed once per module.  This essentially makes it a global object that may change at runtime.

vladimir-kotikov:
@dblotsky, @purplecabbage, thanks for feedback. Agree, this probably smells fishy. However, the only one option i see apart from this one is to setup a proxy to transmit events from `cordova-common` to another `EventEmitter` instance (which was passed to `PlatformApi` constructor), when necessary.

All other solutions, that came into my mind, break `events` API and hence require some global changes in cordova (as the `events` is used really widely).

However, this PR doesn't change a public API for this and seems to fix the issue with missing events from `cordova-common` internals without affecting any other cases, so it might make sense to get this in as an intermediate fix and then start looking into redesigning cordova `events` mechanics.

dblotsky:
@vladimir-kotikov I guess my point is: why is it necessary for internal `cordova-common` events to reach platform code?

vladimir-kotikov:
Because we've been logging these events all the time before cordova refactoring. Many of these events doesn't have much value though, and could be safely ignored, but, for example, the command and arguments, used by superspawn are pretty important information IMO, and there should be a way to get it without using debugger

dblotsky:
Is this a part of our external API? If not, then why should there be a way to get them without the debugger if it's an internal feature?

vladimir-kotikov:
No, this is not a part of API. However, this might be critical for some external tools to get this information.

@nikhilkh, what do you think about this?

nikhilkh:
I understand that this approach is not an ideal pattern - Anyone has better ideas?

We do need this to work to ensure that --verbose logs that we produce have details of activity in cordova-common. In particular, logging of all the commands that we spawn (with arguments). It really helps in diagnosing failures and bugs.

vladimir-kotikov:
Guys, just a couple of ideas, based on discussion with Nikhil:
1. Instead of rewriting cordo

vladimir-kotikov:
Guys, just a couple of ideas, based on discussion with Nikhil:
1, Instead of overwriting `events` instance in common, platform might set a reference to `events` on `global` object, which `cordova-common` will check at runtime and use if available
2. Platform might create a proxy - just subscribe to and redirect `cordova-common`'s `events` to own instance - this one have an advantage - we would not need for any special logic in `common` for that

nikhilkh:
I like the idea of a global.

purplecabbage:
I am still not seeing the value in this.
When is this an issue? Who does it affect?

nikhilkh:
https://issues.apache.org/jira/browse/CB-10430 has details of one scenario where verbose logs from the cordova-common do not show up. In particular, the processes that we spawn using superspawn as part of a build can fail and we get an error without context of the command that was run. Ideally, --verbose build logs will have details of the external command that was executed with the correct arguments.

purplecabbage:
In this case, why not emit the event that you are calling spawn with arguments a,b,c 

```
function execSpawn(command, args, resultMsg, errorMsg) {
    events.emit( ... )
    return superspawn.spawn(command, args).then(function(result) {
   ...
```

nikhilkh:
Yes, we could fix all callers of it - but ideally, 'log' messages from `cordova-common` can be surfaced and reported.

vladimir-kotikov:
Closing this for now. I'll reopen another PR with updated implementation.

PR #374, opened 2016-02-03
CB-10465 Pass correct options to prepare from compile

vladimir-kotikov:
This fixes incorrect behaviour of `cordova emulate` command, which results in restoring all saved platforms even when only particular platform specified.

See [CB-10465](https://issues.apache.org/jira/browse/CB-10465) for details


vladimir-kotikov:
@sgrebnov, PTAL

sgrebnov:
LGTM

PR #375, opened 2016-02-03
CB-10176 Adds standard logger, based on logger module from cordova-cli

vladimir-kotikov:
This is an implementation for [CB-10176](https://issues.apache.org/jira/browse/CB-10176)
This PR adds a standard logger implementation, that could be reused by cli and platforms.


nikhilkh:
LGTM

TimBarham:
:+1:  Thanks Vladimir.

purplecabbage:
lgtm!
Seems we are slowly migrating from lowercase file names to CamelCase ..

vladimir-kotikov:
@purplecabbage, i'd say to CamelCase for classes and lowerCamelCase for the rest

PR #381, opened 2016-02-09
CB-10567 Bubble up `cordova.raw.run()` error to the caller

vladimir-kotikov:
JIRA [CB-10567](https://issues.apache.org/jira/browse/CB-10567)

This PR fixes the case when `cordova.raw.run()` promise always being resolved, event if run itself fails due to some reason.


nikhilkh:
LGTM

PR #383, opened 2016-02-10
CB-10518 Correct log level and error messages for some cordova errors

vladimir-kotikov:
Issue [CB-10518](https://issues.apache.org/jira/browse/CB-10518)

### DIFF
--- a/cordova-lib/src/cordova/compile.js
+++ b/cordova-lib/src/cordova/compile.js
@@ -19,8 +19,6 @@
 
 var cordova_util = require('./util'),
     HooksRunner  = require('../hooks/HooksRunner'),
-    events       = require('cordova-common').events,
-    Q            = require('q'),
     promiseUtil  = require('../util/promise-util'),
     platform_lib = require('../platforms/platforms');
 
@@ -39,8 +37,5 @@ module.exports = function compile(options) {
         });
     }).then(function() {
         return hooksRunner.fire('after_compile', options);
-    }, function(error) {
-        events.emit('warn', 'ERROR building one of the platforms: ' + error + '\nYou may not have the required environment or OS to build this project');
-        return Q.reject(error);
     });
 };
--- a/cordova-lib/src/cordova/compile.js
+++ b/cordova-lib/src/cordova/compile.js
@@ -40,7 +40,7 @@ module.exports = function compile(options) {
     }).then(function() {
         return hooksRunner.fire('after_compile', options);
     }, function(error) {
-        events.emit('log', 'ERROR building one of the platforms: ' + error + '\nYou may not have the required environment or OS to build this project');
+        events.emit('warn', 'ERROR building one of the platforms: ' + error + '\nYou may not have the required environment or OS to build this project');
         return Q.reject(error);
     });
 };
--- a/cordova-lib/src/cordova/create.js
+++ b/cordova-lib/src/cordova/create.js
@@ -191,8 +191,7 @@ function create(dir, optionalId, optionalName, cfg) {
 
                 return remoteLoad.gitClone(gitURL, branch).fail(
                     function(err) {
-                        events.emit('warn', err.message);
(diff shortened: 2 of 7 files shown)


cordova-lib/src/cordova/create.js
```diff
@@ -191,8 +191,8 @@ function create(dir, optionalId, optionalName, cfg) {
 
                 return remoteLoad.gitClone(gitURL, branch).fail(
                     function(err) {
-                        events.emit('verbose', err);
-                        return Q.reject('Failed to retrieve '+ cfg.lib.www.url + ' using git.');
+                        events.emit('warn', err.message);
+                        return Q.reject(new CordovaError('Failed to retrieve '+ cfg.lib.www.url + ' using git.'));
```
nikhilkh: Should the `err.message` be appended to new CordovaError or the original `err` be passed as inner error?

cordova-lib/src/cordova/compile.js
```diff
@@ -40,7 +40,7 @@ module.exports = function compile(options) {
     }).then(function() {
         return hooksRunner.fire('after_compile', options);
     }, function(error) {
-        events.emit('log', 'ERROR building one of the platforms: ' + error + '\nYou may not have the required environment or OS to build this project');
+        events.emit('warn', 'ERROR building one of the platforms: ' + error + '\nYou may not have the required environment or OS to build this project');
```
nikhilkh: So what's the guidance for error handling? We should emit a warning with detailed error message and then reject the promise with the actual error? I notice below you sometimes create a CordovaError to reject the promise. It will be great if we can come up with consistent set of guidelines for this.

cordova-lib/src/cordova/targets.js
```diff
@@ -25,9 +25,9 @@ var cordova_util = require('./util'),
 
 function handleError(error) {
     if (error.code === 'ENOENT') {
-        events.emit('log', 'Platform does not support ' + this.script);
+        events.emit('warn', 'Platform does not support ' + this.script);
     } else {
-        events.emit('log', 'An unexpected error has occured');
+        events.emit('warn', 'An unexpected error has occured while running ' + this.script);
```
nikhilkh: Should we log the error code? It's often useful to have that.

cordova-lib/src/cordova/compile.js
nikhilkh: Also when should one use `events.emit('error', ...)` For all errors now we are using warning - this is confusing.

vladimir-kotikov: I'm not sure if we should continue emitting message here at all, because it is not consistent with other similar places, e.g. but `build` and `emulate` do not emit this message.  Also the message is not very informative, as `compile` might fail due to a lot of different reasons.

In general for cases like this, when initial error message might not be very descriptive, i'd rather reject a promise with (or throw in case of sync code) a new `CordovaError` with appropriate message and original error object, attached.

cordova-lib/src/cordova/run.js
```diff
@@ -44,7 +44,7 @@ module.exports = function run(options) {
     }).then(function() {
         return hooksRunner.fire('after_run', options);
     }, function(error) {
-        events.emit('log', 'ERROR running one or more of the platforms: ' + error + '\nYou may not have the required environment or OS to run this project');
+        events.emit('warn', 'ERROR running one or more of the platforms: ' + error + '\nYou may not have the required environment or OS to run this project');
```
vladimir-kotikov: As in the first comment I'm not sure if we should emit this here because it is not consistent and not very informative. Maybe rethrow with this message. @nikhilkh, what do you think?

cordova-lib/src/cordova/targets.js
vladimir-kotikov: Agree. I'll update this

cordova-lib/src/cordova/compile.js
nikhilkh: Let's remove the warning if it's not needed and looks like other commands do not log anyway.

PR #384, opened 2016-02-10
CB-10519 Wrap all sync calls inside of `cordova.raw` methods into promises

vladimir-kotikov:
This is a fix for [CB-10519](https://issues.apache.org/jira/browse/CB-10519)

This PR fixes the issue when promise returned from `cordova.raw` API would never be rejected if command is ran outside of cordova project. Instead synchronous exception will be thrown (by `util.preProcessOptions`), which is definitely unexpected for promise based API.

There are many indentation changes in this PR, so it's more convenient to review with [`?w=1`](https://github.com/apache/cordova-lib/pull/384?w=1) option

### DIFF
--- a/cordova-lib/src/cordova/build.js
+++ b/cordova-lib/src/cordova/build.js
@@ -17,22 +17,25 @@
     under the License.
 */
 
-var cordovaUtil      = require('./util'),
-    HooksRunner           = require('../hooks/HooksRunner');
+var Q = require('q'),
+    cordovaUtil = require('./util'),
+    HooksRunner = require('../hooks/HooksRunner');
 
 // Returns a promise.
 module.exports = function build(options) {
-    var projectRoot = cordovaUtil.cdProjectRoot();
-    options = cordovaUtil.preProcessOptions(options);
+    return Q().then(function() {
+        var projectRoot = cordovaUtil.cdProjectRoot();
+        options = cordovaUtil.preProcessOptions(options);
 
-    // fire build hooks
-    var hooksRunner = new HooksRunner(projectRoot);
-    return hooksRunner.fire('before_build', options)
-    .then(function() {
-        return require('./cordova').raw.prepare(options);
-    }).then(function() {
-        return require('./cordova').raw.compile(options);
-    }).then(function() {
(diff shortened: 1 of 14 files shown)


cordova-lib/src/cordova/build.js
```diff
@@ -17,22 +17,26 @@
...
-    }).then(function() {
-        return require('./cordova').raw.compile(options);
-    }).then(function() {
-        return hooksRunner.fire('after_build', options);
+    return Q().then(function() {
+        var opts = cordovaUtil.preProcessOptions(options);
+        var projectRoot = cordovaUtil.cdProjectRoot();
+        return [new HooksRunner(projectRoot), opts];
+    })
+    .spread(function (hooksRunner, options) {
```
TimBarham: Hey @vladimir-kotikov - what's the purpose of using `spread()` here (and elsewhere)? Why can't we just essentially wrap the entire method in `return Q().then(function() { ... };`?

vladimir-kotikov: > Why can't we just essentially wrap the entire method in return `Q().then(function() { ... };`?

We can and basically this would be same. I think it's just a matter of taste..
If you wish, i can update the PR in the manner you proposed

TimBarham: Well, that'd be my preference, as the code ends up much simpler and easy to read, to my mind. Otherwise we're arbitrarily splitting out the synchronous part of the process into its own separate promise, when there's no need to. But ultimately it's up to you :smile:.

vladimir-kotikov:
@TimBarham, updated

TimBarham:
Thanks @vladimir-kotikov. Looks good on a quick look through. Will take a closer look in the morning.

vladimir-kotikov:
@TimBarham, ping

TimBarham:
Oops, sorry :smile: ... LGTM. There's nowhere we call these directly internally and expect an actual value in return?

vladimir-kotikov:
The return value of `Q.then(thePromise)` is the same as for `thePromise()` due to its' (promise) nature, so i think there is nothing changed for caller.

PR #385, opened 2016-02-10
CB-10584 Splashscreen plugin crashes the app on windows 10 when built…

daserge:
… with browserify

[Jira issue](https://issues.apache.org/jira/browse/CB-10584)

### DIFF
--- a/cordova-lib/src/plugman/browserify.js
+++ b/cordova-lib/src/plugman/browserify.js
@@ -106,7 +106,7 @@ module.exports = function doBrowserify (project, platformApi, pluginInfoProvider
         commitId = cId;
         return getPlatformVersion(commitId, platformApi.root);
     }).then(function(platformVersion){
-        var libraryRelease = bundle(platform, false, commitId, platformVersion);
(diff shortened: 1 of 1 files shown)


daserge:
It was taking the old version from `legacy_exec` so `base.js` was not found.

daserge:
@vladimir-kotikov, could you please review?

cordova-lib/src/plugman/browserify.js
```diff
@@ -106,7 +106,7 @@ module.exports = function doBrowserify (project, platformApi, pluginInfoProvider
         commitId = cId;
         return getPlatformVersion(commitId, platformApi.root);
     }).then(function(platformVersion){
-        var libraryRelease = bundle(platform, false, commitId, platformVersion);
+        var libraryRelease = bundle(platform, false, commitId, platformVersion, wwwDir);
```
vladimir-kotikov: Shouldn't we use `platform_www` directory here instead of `<platform>/www`?

daserge:
Added [CB-10616 Cover browserify scenario with tests](https://issues.apache.org/jira/browse/CB-10616).

PR #386, opened 2016-02-11
CB-10592 Don't quote platform specific args values

vladimir-kotikov:
See JIRA [CB-10592](https://issues.apache.org/jira/browse/CB-10592) for details

### DIFF
--- a/cordova-lib/src/cordova/util.js
+++ b/cordova-lib/src/cordova/util.js
@@ -297,7 +297,7 @@ function ensurePlatformOptionsCompatible (platformOptions) {
     }).map(function (arg) {
         return opts[arg] === true ?
             '--' + arg :
-            '--' + arg + '="' + opts[arg].toString() + '"';
+            '--' + arg + '=' + opts[arg].toString();
     });
 
     return opts;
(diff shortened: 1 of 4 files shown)


cordova-lib/src/cordova/util.js
```diff
@@ -297,7 +297,7 @@ function ensurePlatformOptionsCompatible (platformOptions) {
     }).map(function (arg) {
         return opts[arg] === true ?
             '--' + arg :
-            '--' + arg + '="' + opts[arg].toString() + '"';
+            '--' + arg + '=' + opts[arg].toString();
```
dblotsky: Why is this no longer escaped?

vladimir-kotikov: Because these quotes distorts arguments values e.g. `['--appx=uap']` becomes `['--appx="uap"']` and then (after parsing by `nopt` in platform) - `{appx: "\"uap"\"}` - notice that quotes now included as a part of `appx` option value.

dblotsky: Why are we passing string CLI arguments in a programmatic API? If this is because we're using `cordova.raw`, then can we refactor to not use it?

vladimir-kotikov: If you're talking about this usage
```javascript
cordova.raw.build({
    platforms: ["android"],
    options: ["--debug", "--gradleArg=--no-daemon"]
})
```
i think we shouldn't do that but still need to maintain this scenario for compatibility.

dblotsky: Alright, fair enough. Do you think we can deprecate this usage in the future though? It's a really awkward (and also insecure) API.

vladimir-kotikov: I hope so. We're actually already emitting a warning when arguments used this way, so we could try to get rid of this in next major release.

PR #387, opened 2016-02-14
CB-10550 Fixed the issue of plugin id mapper not enforced when a vers…

bso-intel:
…ion is specified

The issue is described in https://issues.apache.org/jira/browse/CB-10550
The fix is to inspect the given plugin_src if it contains the '@'version_number' before it tries to find the mapper database.

### DIFF
--- a/cordova-lib/src/plugman/fetch.js
+++ b/cordova-lib/src/plugman/fetch.js
@@ -132,10 +132,14 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
                 ));
             }
             // If not found in local search path, fetch from the registry.
-            var newID = pluginMapperotn[plugin_src];
+            var splitVersion = plugin_src.split('@');
+            var newID = pluginMapperotn[splitVersion[0]];
             if(newID) {
                 events.emit('warn', 'Notice: ' + plugin_src + ' has been automatically converted to ' + newID + ' to be fetched from npm. This is due to our old plugins registry shutting down.');                
                 plugin_src = newID;
+                if (splitVersion[1]) {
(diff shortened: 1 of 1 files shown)


cordova-lib/src/plugman/fetch.js
```diff
@@ -132,10 +132,14 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
                 ));
             }
             // If not found in local search path, fetch from the registry.
-            var newID = pluginMapperotn[plugin_src];
+            var splitVersion = plugin_src.split('@');
+            var newID = pluginMapperotn[splitVersion[0]];
             if(newID) {
                 events.emit('warn', 'Notice: ' + plugin_src + ' has been automatically converted to ' + newID + ' to be fetched from npm. This is due to our old plugins registry shutting down.');
```
TimBarham: Looks great, thanks @bso-intel!

One small nit - use `splitVersion[0]` in the message instead of `plugin_src` (so, for example, we get `org.apache.cordova.device has been automatically converted to cordova-plugin-device to be fetched from npm` instead of `org.apache.cordova.device@1.1.1 has been automatically converted to cordova-plugin-device to be fetched from npm`).

bso-intel:
Hi @TimBarham 
Thanks for reviewing this pull request.
I updated the comment as you requested.

TimBarham:
Thanks @bso-intel. I've merged this in https://github.com/apache/cordova-lib/commit/87d8e171, but I rebased on forgot to add "This closes #387" to the comment, so if you could close this PR manually that'd be great!

PR #388, opened 2016-02-15
CB-10611 fixed the before_plugin_install hook not disabled with --noh…

bso-intel:
…ooks

The fix is to pass the command line option to the plugman when adding a plugin.


TimBarham:
LGTM.

bso-intel:
Thank you very much, Tim.

PR #390, opened 2016-02-15
CB-10621 fix for amazon-fireos

vldmrrr:
For android specific code include test for amazon-fireos platform as well.


stevengill:
Amazon-fireos has actually been deprecated. It is recommended to target the android platform instead when releasing for amazon

PR #393, opened 2016-02-18
CB-10641 Run prepare _after_ plugins were installed

vladimir-kotikov:
This is a fix for [CB-10641](https://issues.apache.org/jira/browse/CB-10641) and [CB-10586](https://issues.apache.org/jira/browse/CB-10586)
 
This PR changes order of operations in `cordova.raw.platform('add', ...)` method so `cordova.raw.prepare` is called after all plugins installed into platform. Also apart from resolving the issue above, the new flow (`platform_add -> install_plugins -> prepare`) looks semantically more correct.

### DIFF
--- a/cordova-lib/spec-cordova/platform.spec.js
+++ b/cordova-lib/spec-cordova/platform.spec.js
@@ -26,6 +26,7 @@ var helpers = require('./helpers'),
     Q = require('q'),
     events = require('cordova-common').events,
     cordova = require('../src/cordova/cordova'),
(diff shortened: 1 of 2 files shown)


vladimir-kotikov:
@TimBarham, could you please take a look

TimBarham:
Yep, that looks right to me.

nikhilkh:
Is this fixing a regression in behavior from previous versions? Let's make sure we label the associated JIRAs as such.

vladimir-kotikov:
@nikhilkh, will do that

PR #394, opened 2016-02-18
CB-10644 Adds deprecation message about old platforms support removal

vladimir-kotikov:
JIRA: [CB-10644](https://issues.apache.org/jira/browse/CB-10644)

### DIFF
--- a/cordova-lib/src/platforms/platforms.js
+++ b/cordova-lib/src/platforms/platforms.js
@@ -48,11 +48,17 @@ function getPlatformApi(platform, platformRootDir) {
     var PlatformApi;
     try {
         // First we need to find whether platform exposes its' API via js module
-        // If it has, then we have to require it and extend BasePlatformApi
-        // with platform's API.
+        // If it does, then we require and instantiate it.
         var platformApiModule = path.join(platformRootDir, 'cordova', 'Api.js');
         PlatformApi = require(platformApiModule);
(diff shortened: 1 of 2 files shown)


cordova-lib/src/platforms/platforms.js
```diff
@@ -48,11 +48,17 @@ function getPlatformApi(platform, platformRootDir) {
...
         // First we need to find whether platform exposes its' API via js module
-        // If it has, then we have to require it and extend BasePlatformApi
-        // with platform's API.
+        // If it does, then we require and instantiate it.
         var platformApiModule = path.join(platformRootDir, 'cordova', 'Api.js');
         PlatformApi = require(platformApiModule);
     } catch (err) {
+        // Check if platform already compatible w/ PlatformApi and show deprecation warning
+        if (err && err.code === 'MODULE_NOT_FOUND' && platforms[platform].apiCompatibleSince) {
+            events.emit('warn', 'The support for cordova-' + platform + ' via PlatformApi polyfill ' +
```
nikhilkh: instead of leaking internal implementation detail we should consider re-phrasing this as:
"Using this version of cordova CLI with older version of cordova-<platform> is being deprecated. Consider upgrading to..."

vladimir-kotikov:
@nikhilkh, updated

nikhilkh:
LGTM

PR #395, opened 2016-02-19
CB-10654 make hooks fire when platforms added from repo or dir

tony--:
JIRA [CB-10654](https://issues.apache.org/jira/browse/CB-10654)
move `hooksRunner.fire('before_platform_'...)` till after platform has been resolved, pass resolved platform instead of target.
move `hooksRunner.fire('after_platform_'...)` into platform loop, pass resolved platform instead of target.

### DIFF
--- a/cordova-lib/src/cordova/platform.js
+++ b/cordova-lib/src/cordova/platform.js
@@ -79,148 +79,157 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
     var platformsDir = path.join(projectRoot, 'platforms');
     shell.mkdir('-p', platformsDir);
 
-    return hooksRunner.fire('before_platform_' + cmd, opts)
-    .then(function() {
-        return promiseutil.Q_chainmap(targets, function(target) {
-            // For each platform, download it and call its helper script.
-            var parts = target.split('@');
-            var platform = parts[0];
-            var spec = parts[1];
+    return promiseutil.Q_chainmap(targets, function(target) {
+        // For each platform, download it and call its helper script.
+        var parts = target.split('@');
+        var platform = parts[0];
+        var spec = parts[1];
+        var platDetails = "";
 
-            return Q.when().then(function() {
-                if (!(platform in platforms)) {
-                    spec = platform;
-                    platform = null;
+        return Q.when().then(function() {
+            if (!(platform in platforms)) {
+                spec = platform;
+                platform = null;
+            }
+
+            if(platform === 'amazon-fireos') {
+                events.emit('warn', 'amazon-fireos has been deprecated. Please use android instead.');
+            }
+            if(platform === 'wp8') {
+                events.emit('warn', 'wp8 has been deprecated. Please use windows instead.');
+            }
+            if (platform && !spec && cmd == 'add') {
+                events.emit('verbose', 'No version supplied. Retrieving version from config.xml...');
+                spec = getVersionFromConfigFile(platform, cfg);
+            }
+
+            // If --save/autosave on && no version specified, use the pinned version
+            // e.g: 'cordova platform add android --save', 'cordova platform update android --save'
+            if( (opts.save || autosave) && !spec ){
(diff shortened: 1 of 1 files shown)


tony--:
I'd like to see this get fixed, so I thought I'd submit a PR, but I'm a node noob - this definitely needs to be reviewed before being merged in.

tony--:
@bso-intel will you take a look, please?

cordova-lib/src/cordova/platform.js
```diff
@@ -79,148 +79,156 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
...
+                if (cordova_util.isDirectory(maybeDir)) {
+                    return getPlatformDetailsFromDir(maybeDir, platform);
+                }
+            }
+            return downloadPlatform(projectRoot, platform, spec, opts);
+        }).then(function(platformDetails) {
+            platDetails = platformDetails;
+            var hookOpts = {
+                platforms :[platDetails.platform],
+                nohooks :[opts.nohooks]
```
bso-intel: opts.nohooks is already an array.
So you can change this line to the following
nohooks: opts.nohooks

cordova-lib/src/cordova/platform.js
```diff
@@ -79,148 +79,156 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
...
+                    // Save target into config.xml, overriding already existing settings
+                    events.emit('log', '--save flag or autosave detected');
+                    events.emit('log', 'Saving ' + platform + '@' + spec + ' into config.xml file ...');
+                    cfg.removeEngine(platform);
+                    cfg.addEngine(platform, spec);
+                    cfg.write();
+                }
+                var hookOpts = {
+                    platforms :[platDetails.platform],
+                    nohooks :[opts.nohooks]
```
bso-intel: The same thing. Please remove bracket [ ] here.

cordova-lib/src/cordova/platform.js
```diff
@@ -79,148 +79,156 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
...
+                }
+            }
+            return downloadPlatform(projectRoot, platform, spec, opts);
+        }).then(function(platformDetails) {
+            platDetails = platformDetails;
+            var hookOpts = {
+                platforms :[platDetails.platform],
+                nohooks :[opts.nohooks]
+            };
+            return hooksRunner.fire('before_platform_' + cmd, hookOpts);
```
bso-intel: Now, the behavior is a little different than before.
Previously, the before_platform_add hook ran only once for all target platforms.
This hook could be platform agnostic.
However, by moving the hookRunner inside per platform, this hook will run every time each platform is added.
Not sure whether this is an issue or not, though.

bso-intel:
It looks great assuming that before_platform_add and after_platform_add hooks could be platform specific.
My understanding about the current convention is a little different, though.
Thanks, Tony.

cordova-lib/src/cordova/platform.js
tony--: Thanks for this feedback - that's a good point - I was focused on the platform-specific case.
Instead of firing the hook once per platform, I could retain the once per operation behavior.
I would move the hooksRunner.fire back out of the loop (`promiseutil.Q_chainmap`), but exit the first loop after the targets are resolved and place `hooksRunner.fire` for `before_platform_`.
Then I would start a second loop (`promiseutil.Q_chainmap`) that actually adds the platforms.
I'll make this change ASAP and update the review.

nikhilkh: Do we end up losing some of the original `opts` - as I do not see us copying them over.

tony--: I was only providing the opts that are used by HooksRunner and scriptFinder, but I was planning to switch to cloning opts and replacing the platform in my copy.  Seems less brittle than my current approach.

PR #396, opened 2016-02-21
CB-10314 avoid fetching plugins when oldId is already fetched

bso-intel:
implement the fix to the issue in CB-10314.
Before start fetching the plugin, first check if the plugin with alias id is already fetched in the project.
Otherwise, it is a waste of time fetching the plugin with newId and then decide not to install.
This pull request also includes a fix to the issue of reporting the bogus mismatched id error which is caused by fetching the new-Id when old-Id is renamed to fetch. (checkID function in fetch.js)
After the fix, the plugin add is changed as follows:
```
[CB-10314] cordova plugin
cordova-plugin-whitelist 1.2.1 "Whitelist"
org.apache.cordova.device 0.3.0 "Device"
[CB-10314] cordova plugin add org.apache.cordova.device
Plugin "org.apache.cordova.device" already installed on ios.
```

### DIFF
--- a/cordova-lib/src/plugman/fetch.js
+++ b/cordova-lib/src/plugman/fetch.js
@@ -150,6 +150,7 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
                 // if the plugin alias has already been fetched, use it.
                 var alias = pluginMappernto[splitVersion[0]] || newID;
                 if (alias && fs.existsSync(path.join(plugins_dir, alias))) {
+                    events.emit('warn', 'Found '+alias+' is already fetched. Skipped fetching '+splitVersion[0]);
                     P = Q(path.join(plugins_dir, alias));
                     skipCopyingPlugin = true;
                 } else {
--- a/cordova-lib/src/plugman/fetch.js
+++ b/cordova-lib/src/plugman/fetch.js
@@ -135,13 +135,32 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
             var splitVersion = plugin_src.split('@');
             var newID = pluginMapperotn[splitVersion[0]];
             if(newID) {
-                events.emit('warn', 'Notice: ' + splitVersion[0] + ' has been automatically converted to ' + newID + ' to be fetched from npm. This is due to our old plugins registry shutting down.');
                 plugin_src = newID;
                 if (splitVersion[1]) {
                     plugin_src += '@'+splitVersion[1];
                 }
-            } 
-            return registry.fetch([plugin_src])
+            }
+            var P, skipCopyingPlugin;
+            plugin_dir = path.join(plugins_dir, splitVersion[0]);
+            // if the plugin has already been fetched, use it.
+            if (fs.existsSync(plugin_dir)) {
+                P = Q(plugin_dir);
+                skipCopyingPlugin = true;
+            } else {
+                // if the plugin alias has already been fetched, use it.
+                var alias = pluginMappernto[splitVersion[0]] || newID;
+                if (alias && fs.existsSync(path.join(plugins_dir, alias))) {
+                    P = Q(path.join(plugins_dir, alias));
+                    skipCopyingPlugin = true;
+                } else {
+                    if (newID) {
+                        events.emit('warn', 'Notice: ' + splitVersion[0] + ' has been automatically converted to ' + newID + ' to be fetched from npm. This is due to our old plugins registry shutting down.');
+                    }
+                    P = registry.fetch([plugin_src]);
+                    skipCopyingPlugin = false;
+                }
+            }
+            return P
             .fail(function (error) {
                 var message = 'Failed to fetch plugin ' + plugin_src + ' via registry.' +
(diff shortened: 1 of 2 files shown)


bso-intel:
@vladimir-kotikov 
Could you review this?
Thanks.

cordova-lib/src/plugman/install.js
```diff
@@ -79,21 +79,36 @@ module.exports = function installPlugin(platform, project_dir, id, plugins_dir,
...
-        events.emit('warn', 'Notice: ' + id + ' has been automatically converted to ' + newId + ' and fetched from npm. This is due to our old plugins registry shutting down.');
         if(splitVersion[1]) {
             id = newId +'@'+splitVersion[1];
         } else {
             id = newId;
         }
-     }
-    return possiblyFetch(id, plugins_dir, options)
+    }
+    var P;
```
vladimir-kotikov: I'd move the chunk below to `possiblyFetch` function. This way you could just `return Q(plugin_dir)` instead of using intermediate variable. Also `possiblyFetch` already contains logic to check existence of plugin directory.

cordova-lib/src/plugman/install.js
```diff
@@ -79,21 +79,36 @@ module.exports = function installPlugin(platform, project_dir, id, plugins_dir,
...
+    var plugin_dir = path.join(plugins_dir, splitVersion[0]);
+    // if the plugin has already been fetched, use it.
+    if (fs.existsSync(plugin_dir)) {
+        P = Q(plugin_dir);
+    } else {
+        var alias = pluginMapper.newToOld[splitVersion[0]] || newId;
+        // if the plugin alias has already been fetched, use it.
+        if (alias && fs.existsSync(path.join(plugins_dir, alias))) {
+            P = Q(path.join(plugins_dir, alias));
```
vladimir-kotikov: Maybe we should show a message when we're going to use already fetched plugin with old id despite that user tries to install the one with new id?

cordova-lib/src/plugman/fetch.js
```diff
@@ -135,13 +135,32 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
...
-                events.emit('warn', 'Notice: ' + splitVersion[0] + ' has been automatically converted to ' + newID + ' to be fetched from npm. This is due to our old plugins registry shutting down.');
                 plugin_src = newID;
                 if (splitVersion[1]) {
                     plugin_src += '@'+splitVersion[1];
                 }
-            } 
-            return registry.fetch([plugin_src])
+            }
+            var P, skipCopyingPlugin;
+            plugin_dir = path.join(plugins_dir, splitVersion[0]);
```
vladimir-kotikov: It looks like these checks were already done in `plugman/install.js`. If we get here then neither `plugins_dir/<oldId>` nor `plugins_dir/<newId>`shouldn't exist, or i'm missing something?

bso-intel: In fact, cordova-plugin.js invokes plugman-fetch first and then plugman-install.
So, we need to check existence of the plugin here.
The same checking in plugman-install is still needed because plugman-install can be invoked directly from plugin CLI.

cordova-lib/src/plugman/install.js
bso-intel: Yes, that's a good idea.
I revised the code as you suggested.

cordova-lib/src/plugman/install.js
bso-intel: I actually added a warning message in the fetch.js file.

bso-intel:
Hi Vladimir,
I updated the code to accommodate your comments.
Thanks.

cordova-lib/src/plugman/fetch.js
vladimir-kotikov: > In fact, cordova-plugin.js invokes plugman-fetch first and then plugman-install.

This is probably excess and could be reworked to just call `plugman/install` as it calls `tryFetch` anyway.

> because plugman-install can be invoked directly from plugin CLI

I guess you meant `plugman/fetch`, right? This might be a problem, but since this is user command, could we shift the responsibility for this to the user?

vladimir-kotikov:
@bso-intel, thanks. This looks good apart from the logic duplication in `plugman/install` and `plugman/fetch`. However, this looks like is out of the scope of this PR and could be addressed later.

Do you want me to merge this?

bso-intel:
Yes, please merge this for me.
I don't have committer privilege, which I wish to get. :)
Thanks, Vladimir.

cordova-lib/src/plugman/fetch.js
bso-intel: I agree.

bso-intel:
@vladimir-kotikov ,
Just a friendly reminder. This is still not merged.

vladimir-kotikov:
Sorry for delay, @bso-intel. Merging

bso-intel:
Thank you for merging, @vladimir-kotikov

PR #398, opened 2016-02-24
CB-10662 Use project's config.xml as a fallback for package name

vladimir-kotikov:
JIRA: [CB-10662](https://issues.apache.org/jira/browse/CB-10662)

This change is needed because some platforms (`browser` in particular)  doesn't contain own `config.xml` and rely on cordova, which generates one from defaults during `prepare`. However, prepare now called after plugin installation (after b04db5feaf5516836c8a157cc3331b75d6dca0eb) and hence fails due to abscence of config.

### DIFF
--- a/cordova-lib/src/plugman/platforms/browser.js
+++ b/cordova-lib/src/plugman/platforms/browser.js
@@ -20,10 +20,8 @@
 /* jshint laxcomma:true, sub:true */
 
 var path = require('path')
-    , fs = require('fs')
     , common = require('./common')
     , events = require('cordova-common').events
-    , xml_helpers = require('cordova-common').xmlHelpers
     ;
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


vladimir-kotikov:
@TimBarham, could you take a look please?

vladimir-kotikov:
Looking into failures...

Another possible fix here - update `browser` platform rather than introducing additional complexity to platform adding process. We just need to generate valid `config.xml` in browser's `bin/create` script according to passed parameters.

TimBarham:
A simpler fix might be: since all we're looking for is the project's `id`, why not fall back on the main Cordova project's `config.xml` if we don't find a platform specific one? We could change this in `cordova-lib\src\plugman\platforms\browser.js`. Or, since this code is fairly common across several platforms, we could provide a single implementation (in `common.js`, perhaps?) for those that need it.

vladimir-kotikov:
Thanks for advice, @TimBarham. Updated

cordova-lib/src/plugman/platforms/common.js
```diff
@@ -22,7 +22,39 @@ var shell = require('shelljs'),
...
+            // If no config there try next path
+            if (!fs.existsSync(configPath)) continue;
+
+            var widget_doc = xmlHelpers.parseElementtreeSync(configPath);
+            return widget_doc._root.attrib.id;
+        }
+
+        // No configs found - fail with meaningful error message
+        throw new CordovaError('Unable to find project\'s config in none of ' +
+            'the following directories:\n\t' + configPaths.join('\n\t'));
```
TimBarham: I would suggest "any" instead of "none".

cordova-lib/src/plugman/platforms/browser.js
```diff
@@ -31,25 +29,7 @@ module.exports = {
...
-                events.emit('verbose', 'unable to find '+config_path);
-            } else {
-                config_path = old_config_path;
-            }
-        } else {
-            config_path = preferred_path;
-        }
-        var widget_doc = xml_helpers.parseElementtreeSync(config_path);
-        return widget_doc._root.attrib['id'];
+        return common.package_name(project_dir);
```
TimBarham: Perhaps should make the same change for `firefoxos` and `webos` (and maybe `blackberry10`) too?

TimBarham:
Thanks @vladimir-kotikov - looks good. Just a couple of suggestions.

cordova-lib/src/plugman/platforms/browser.js
vladimir-kotikov: Agree about `firefoxos` and `webos` (I'd add here `tizen` as well)

As for `blackberry10` - it does not rely on configs other than the one at the platform `www` dir, so using common `package_name` might lead to unexpected results. Also seems that it doesn't affected by this bug as it always copying own `config.xml` at `create`

PR #400, opened 2016-02-25
CB-10583 Protect plugin preferences from adding extra Array properties

vladimir-kotikov:
JIRA [CB-10583](https://issues.apache.org/jira/browse/CB-10583)


vladimir-kotikov:
@sgrebnov, could you take a look please?

sgrebnov:
:+1:

PR #401, opened 2016-02-26
CB-10462 Get rid of npmconf in favor of npm

vladimir-kotikov:
JIRA [CB-10462](https://issues.apache.org/jira/browse/CB-10462)


vladimir-kotikov:
@nikhilkh, could you please take a look?

nikhilkh:
LGTM.

PR #402, opened 2016-02-28
CB-10328 set top-level property of plugins when adding new platforms

bso-intel:
The problem is caused by not setting the top-level property of plugins from cordova/platform.js-installPluginsForNewPlatform() function and by always setting is_top_level to true in plugman/install.js-installPlugin.
This caused all plugins that exist in the plugins directory to be set top-level in plugins/<platform>.json files.
The uninstallation process relies on this metadata file to see if there is any dependency plugin was installed when removing a top-level plugin.
By setting the top-level correctly in <platform>.json files, this uninstallation process worked correctly.

After this fix is applied, here is how the behavior is changed:
```
[CB-10328] cordova plugin
cordova-plugin-file 4.1.1 "File"
cordova-plugin-media 2.1.0 "Media"
cordova-plugin-whitelist 1.2.1 "Whitelist"
[CB-10328] cordova plugin rm cordova-plugin-media
Uninstalling 1 dependent plugins.
Uninstalling cordova-plugin-file from android
Uninstalling cordova-plugin-media from android
Uninstalling 1 dependent plugins.
Uninstalling cordova-plugin-file from ios
Uninstalling cordova-plugin-media from ios
Removing "cordova-plugin-media"
[CB-10328] cordova plugin
cordova-plugin-whitelist 1.2.1 "Whitelist"
```

### DIFF
--- a/cordova-lib/src/plugman/install.js
+++ b/cordova-lib/src/plugman/install.js
@@ -71,7 +71,8 @@ module.exports = function installPlugin(platform, project_dir, id, plugins_dir, 
     project_dir = cordovaUtil.convertToRealPathSafe(project_dir);
     plugins_dir = cordovaUtil.convertToRealPathSafe(plugins_dir);
     options = options || {};
-    options.is_top_level = true;
+    if (options.is_top_level === undefined)
+        options.is_top_level = true;
     plugins_dir = plugins_dir || path.join(project_dir, 'cordova', 'plugins');
 
     if (!platform_modules[platform]) {
--- a/cordova-lib/src/plugman/install.js
+++ b/cordova-lib/src/plugman/install.js
@@ -71,7 +71,7 @@ module.exports = function installPlugin(platform, project_dir, id, plugins_dir, 
     project_dir = cordovaUtil.convertToRealPathSafe(project_dir);
     plugins_dir = cordovaUtil.convertToRealPathSafe(plugins_dir);
     options = options || {};
-    if (options.is_top_level === undefined)
+  if (options.hasOwnProperty('is_top_level') === false)
         options.is_top_level = true;
     plugins_dir = plugins_dir || path.join(project_dir, 'cordova', 'plugins');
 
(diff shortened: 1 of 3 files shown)


bso-intel:
@vladimir-kotikov 
Could you review this pull request?
Thanks.

cordova-lib/src/plugman/install.js
```diff
@@ -71,7 +71,8 @@ module.exports = function installPlugin(platform, project_dir, id, plugins_dir,
     project_dir = cordovaUtil.convertToRealPathSafe(project_dir);
     plugins_dir = cordovaUtil.convertToRealPathSafe(plugins_dir);
     options = options || {};
-    options.is_top_level = true;
+    if (options.is_top_level === undefined)
```
vladimir-kotikov: Maybe use `options.hasOwnProperty`? What if it's set to `null` for example?

vladimir-kotikov:
@bso-intel, LGTM apart from one comment. It also would be great if you add a test for this.

cordova-lib/src/plugman/install.js
bso-intel: I changed as you suggested.
Do you mean you want to set options.is_top_level = true if options.is_top_level is set to null ?
In fact, the is_top_level is a boolean property, so a null value for this property is a bug although null is falsy in Javascript.
When options.is_top_level is set to null for some reason, options.hasOwnProperty('is_top_level') is  true.
That means, the new condition (options.hasOwnProperty('is_top_level') === false) will not change options.is_top_level from the null value to true.
Just wanted clarify.

bso-intel:
@vladimir-kotikov ,
I added an e2e test that is described in CB-10328.
The test fails without my fix, and passes with my fix.
Could you merge this pull request?
Thanks.

cordova-lib/src/plugman/install.js
vladimir-kotikov: I meant that IMO the purpose of this check is to validate id someone already specified this value, and in this case it's beter to use `hasOwnProperty` which does exactly what it means

bso-intel:
Thanks for merging, @vladimir-kotikov

PR #403, opened 2016-02-29
CB-10708 Install/uninstall plugins correctly into CLI project using plugman

vladimir-kotikov:
JIRA: https://issues.apache.org/jira/browse/CB-10708

This forces plugman to install plugin's www files to `platform_www` directory instead of `www` (see also https://github.com/apache/cordova-lib/pull/360 for background) when cordova CLI project detected. This allows to install plugins for each platform separately.


nikhilkh:
LGTM. Ideally, we need test cases for this - but I can understand this might be tricky to test.

PR #404, opened 2016-03-03
CB-10673 added --force-copying-src option

bso-intel:
This option is need to rescue some Cordova users who has trouble installing two conflicting plugins that use the same source-file target.


bso-intel:
@nikhilkh,
Could you review this?
Thanks.

PR #405, opened 2016-03-04
CB-10791 Add adjustLoggerLevel to cordova-common.CordovaLogger

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-10791)

### DIFF
--- a/cordova-common/src/CordovaLogger.js
+++ b/cordova-common/src/CordovaLogger.js
@@ -153,6 +153,23 @@ CordovaLogger.prototype.setLevel = function (logLevel) {
 };
 
 /**
+ * Adjusts the current logger level according to the passed options.
+ *
+ * @param   {Object|Array}  opts  An object or args array with options
+ *
+ * @return  {CordovaLogger}     Current instance, to allow calls chaining.
+ */
+CordovaLogger.prototype.adjustLevel = function (opts) {
+    if (opts.verbose || (Array.isArray(opts) && opts.indexOf('--verbose') !== -1)) {
+        this.setLevel('verbose');
+    } else if (opts.silent || (Array.isArray(opts) && opts.indexOf('--silent') !== -1)) {
+        this.setLevel('error');
+    }
+
+    return this;
+};
+
+/**
  * Attaches logger to EventEmitter instance provided.
  *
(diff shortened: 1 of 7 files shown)


cordova-common/src/CordovaLogger.js
```diff
@@ -153,6 +153,30 @@ CordovaLogger.prototype.setLevel = function (logLevel) {
...
+    if (opts instanceof Array) {
+        opts.silent = opts.indexOf('--silent') !== -1;
+        opts.verbose = opts.indexOf('--verbose') !== -1 || opts.indexOf('-d') !== -1;
+    }
+
+    if (opts.silent) {
+        this.setLevel('error');
+    }
+
+    if (opts.verbose || opts.d) {
```
vladimir-kotikov: I think that `opts.d` is not needed because nopt converts `-d` to `{verbose: true}` if `-d` is a shortcut for `-verbose`

daserge:
Looks like github repo is out of sync with Apache'.

cordova-common/src/CordovaLogger.js
```diff
@@ -153,6 +153,30 @@ CordovaLogger.prototype.setLevel = function (logLevel) {
...
 /**
+ * Adjusts the current logger level according to the passed options.
+ *
+ * @param   {Object|Array}  opts  An object or args array with options
+ *
+ * @return  {CordovaLogger}     Current instance, to allow calls chaining.
+ */
+CordovaLogger.prototype.adjustLoggerLevel = function (opts) {
+    if (opts instanceof Array) {
+        opts.silent = opts.indexOf('--silent') !== -1;
```
purplecabbage: Please do not change the type of the object. Why not go with something concise and easy to understand like : 
```
if(Array.isArray(opts)) {
     if(opts.indexOf('--verbose') > -1) {
        this.setLevel('verbose');
    }
    else if (opts.indexOf('--silent') > -1) {
        this.setLevel('error');
    }
}
return this;
```

Note also that I reversed the order of checks in the if/else to reflect that in the original code --verbose out ranks --silent

daserge: Updated, thank you Jesse!

PR #407, opened 2016-03-09
CB-9858 Cordova Fetch Work

stevengill:
This is not ready to merge yet! Opening the PR so it is easy to manage diffs and take feedback


stevengill:
I decided to write the tests in node-tap. I am now regretting that decision a bit. Tests randomly timeout and fail. Doesn't happen in actual use.

stevengill:
Needs https://github.com/apache/cordova-cli/pull/239.

Discussion at https://github.com/cordova/cordova-discuss/pull/33

cordova-fetch/index.js
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * @return {String||Promise}    Returns string of the absolute path to the installed module.
+ *
+ */
+module.exports = function(target, dest, opts) {
+    var fetchArgs = ['install'];
+    opts = opts || {};
+    var tree1;
+
+    if(!shell.which('npm')) {
```
nikhilkh: Judging from previous experience with our tools not being able to find ANDROID_HOME/JAVA_HOME - this might become an error message that will affect new people. We should consider doing a best effort to find the npm installed on the system if `which` can't find it. There are some well known places on Windows/Mac where this gets installed.

cordova-fetch/index.js
```diff
@@ -0,0 +1,176 @@
...
+        
+        //create dest if it doesn't exist
+        if(!fs.existsSync(dest)) {
+            shell.mkdir('-p', dest);         
+        } 
+
+    } else return Q.reject(new CordovaError('Need to supply a target and destination'));
+
+    //set the directory where npm install will be run
+    opts.cwd = dest;
```
nikhilkh: Are the `opts` for superspawn the same as the `opts` passed to this function?

cordova-fetch/index.js
stevengill: Sounds like a good idea! Do you have any examples of doing this with Java_home I could use as a basis for this? I pretty much mickimed how our current fetch code looks for git.

cordova-fetch/index.js
stevengill: Yup! Pass everything to superspawn that fetch was passed. Add the cwd to opts though.

stevengill:
To-do: add cordova-lib tests for this

cordova-fetch/index.js
nikhilkh: https://github.com/apache/cordova-android/blob/master/bin/lib/check_reqs.js#L97. We end up modifying process.env["PATH"] with probable directories for JAVA_HOME/ANDROID_HOME  before invoking superspawn.

cordova-fetch/index.js
```diff
@@ -0,0 +1,176 @@
...
+
+ Unless required by applicable law or agreed to in writing,
+ software distributed under the License is distributed on an
+ 'AS IS' BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ KIND, either express or implied.  See the License for the
+ specific language governing permissions and limitations
+ under the License.
+ */
+
+var Q = require('q');
```
nikhilkh: I was trying to give it a spin but I get this error:
λ cordova platform add android --verbose
Error: Cannot find module 'q'
    at Function.Module._resolveFilename (module.js:336:15)
    at Function.Module._load (module.js:286:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (D:\cordova\cordova-lib\cordova-fetch\index.js:20:9)

Looks like the package.json for this module needs more dependencies.

stevengill: You are correct. Looks like I had it locally but didn't save it. I'll update this when I land and get to my computer.

stevengill:
Plugman PR adding --fetch flag support https://github.com/apache/cordova-plugman/pull/87

cordova-fetch/index.js
```diff
@@ -0,0 +1,226 @@
...
+    opts.cwd = dest;
+
+    //if user added --save flag, pass it to npm install command
+    if(opts.save) {
+        events.emit('verbose', 'saving');
+        fetchArgs.push('--save'); 
+    } 
+    
+    //check if npm is installed
+    isNpmInstalled();
```
vladimir-kotikov: Probably it makes sense to check this at the very beginning of the method

cordova-fetch/index.js
```diff
@@ -0,0 +1,226 @@
...
+
+    //Grab json object of installed modules before npm install
+    return depls(dest)
+    .then(function(depTree) {
+        tree1 = depTree;
+
+        //install new module
+        return superspawn.spawn('npm', fetchArgs, opts);
```
vladimir-kotikov: Perhaps it is worth to add `--production` flag to avoid installing `devDependencies`

cordova-fetch/index.js
stevengill: True. I'll move it up

cordova-fetch/index.js
stevengill: I don't think it is necessary. It is running `npm install` for modules from registry or git url. `devDependencies` shouldn't be getting installed in this scenario regardless of `--production`

stevengill:
Hey Everyone,

I think the first implementation of this is ready to hit master behind the --fetch flag. It includes platform and plugin fetching + removing. 

Take a look @nikhilkh @purplecabbage @vladimir-kotikov 

Needs: apache/cordova-cli#239 and apache/cordova-plugman#87

I will add template fetching next.

I did run into some issues with our cordova tests. The tests finish before the promise resolves. Probably a issue with how we are handling promises in cordova, but could also be related to us using jasmine 1.3 instead of jasmine 2.x. I'll continue to investigate but not a blocker. My manually testing shows that the uninstall command works fine with cordova.

cordova-fetch tests are written using jasmine 2.0. No issues there.

cordova-fetch/index.js
```diff
@@ -0,0 +1,234 @@
...
+var isUrl = require('is-url');
+
+/* 
+ * A function that npm installs a module from npm or a git url
+ *
+ * @param {String} target   the packageID or git url
+ * @param {String} dest     destination of where to install the module
+ * @param {Object} opts     [opts={save:true}] options to pass to fetch module
+ *
+ * @return {String||Promise}    Returns string of the absolute path to the installed module.
```
vladimir-kotikov: Nit: the JSDoc notation for multiple types is a bit different - single `|` instead `||` (please see http://usejsdoc.org/tags-type.html).

cordova-fetch/index.js
```diff
@@ -0,0 +1,234 @@
...
+    return target;
+}
+
+/* 
+ * Takes the moduleID and destination and returns an absolute path to the module
+ *
+ * @param {String} id       the packageID
+ * @param {String} dest     destination of where to fetch the modules
+ *
+ * @return {String||Error}  Returns the absolute url for the module or throws a error
```
vladimir-kotikov: Same here

cordova-fetch/spec/package.json
```diff
@@ -0,0 +1,11 @@
+{
```
vladimir-kotikov: Do we need a separate `package.json` for specs?

cordova-fetch/index.js
```diff
@@ -0,0 +1,234 @@
...
+ * @return {String||Promise}    Returns string of the absolute path to the installed module.
+ *
+ */
+module.exports = function(target, dest, opts) {
+    var fetchArgs = ['install'];
+    opts = opts || {};
+    var tree1;
+
+    //check if npm is installed
+    isNpmInstalled();
```
vladimir-kotikov: Looks like in case if `npm` is not available this will return a promise rejection which is never propagated to the caller. Probably you need to wrap this call and all the code below into top-level promise (similar to #384). This would also protect the code against [CB-10519](https://issues.apache.org/jira/browse/CB-10519)

cordova-fetch/index.js
```diff
@@ -0,0 +1,234 @@
...
+ *
+ * @return {Promise||Error}    Returns a promise with the npm uninstall output or an error.
+ *
+ */
+module.exports.uninstall = function(target, dest, opts) {
+    var fetchArgs = ['uninstall'];
+    opts = opts || {};
+
+    //check if npm is installed on the system
+    isNpmInstalled();
```
vladimir-kotikov: The same issue with not rejecting the promise in case of missing `npm`

cordova-fetch/index.js
stevengill: Thanks! I'll fix that

cordova-fetch/index.js
stevengill: Your right. I originally has it connected to the promise change but removed it during a refactor. I'll fix this!

cordova-fetch/spec/package.json
stevengill: I just copy it into temporary tests folder so I can test - - save. Any alternative suggestion? Should I store it somewhere else?

vladimir-kotikov: Hmm. Maybe just name it differently. My concern is that presence of another `package.json` might be a bit confusing + break some tools in some scenarios

stevengill:
Made the changes. LMK.

vladimir-kotikov:
LGTM

stevengill:
Merged

PR #408, opened 2016-03-10
CB-10822 Manage plugins/modules metadata using PlatformJson

vladimir-kotikov:
JIRA https://issues.apache.org/jira/browse/CB-10822

This PR adds brings logic for storing and generating plugin metadata into PlatformJson class. The logic is fairly similar for all platforms and relies on PlatformJson as a storage.

### DIFF
--- a/cordova-common/package.json
+++ b/cordova-common/package.json
@@ -33,14 +33,15 @@
     "plist": "^1.2.0",
     "q": "^1.4.1",
     "semver": "^5.0.1",
-    "shelljs": "^0.5.1",
+    "shelljs": "^0.5.3",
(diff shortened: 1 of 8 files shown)


vladimir-kotikov:
https://github.com/apache/cordova-lib/commit/317cfdb374b9dfa215de77c617db0c373dc098c6 is seen here due to ASF git and Github is out of sync.

vladimir-kotikov:
The related changes for cordova android can be found here: https://github.com/apache/cordova-android/compare/master...MSOpenTech:CB-10822?expand=1

vladimir-kotikov:
This has been integrated as a part of #414, closing

PR #410, opened 2016-03-14
CB-10808 CLI Should handle templates with files located in sub directory

csantanapr:
Refer to [CB-10808](https://issues.apache.org/jira/browse/CB-10808) for more info

### DIFF
--- a/cordova-lib/src/cordova/create.js
+++ b/cordova-lib/src/cordova/create.js
@@ -220,8 +220,19 @@ function create(dir, optionalId, optionalName, cfg) {
             }
         }
     })
-    .then(function(import_from_path) {
-
+    .then(function(input_directory) {
+        var import_from_path = input_directory;
+        
+        //handle when input wants to specify sub-directory 
+        try {
+            var templatePkg = require(input_directory);
+            if (templatePkg && templatePkg.dirname){
+                import_from_path = templatePkg.dirname;
+            }
+        } catch (e) {
+            events.emit('verbose', 'Can not load template package.json using directory ' + input_directory); 
+        }
+         
         if (!fs.existsSync(import_from_path)) {
             throw new CordovaError('Could not find directory: ' +
                 import_from_path);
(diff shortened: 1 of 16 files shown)


csantanapr:
@stevengill @vladimir-kotikov  could you review this when you have chance.

cordova-lib/src/cordova/create.js
```diff
@@ -220,8 +220,22 @@ function create(dir, optionalId, optionalName, cfg) {
...
         }
     })
-    .then(function(import_from_path) {
-
+    .then(function(input_directory) {
+        var import_from_path = input_directory;
+        var templatePkg;
+        
+        //handle package.json present with dirname
+        if (fs.existsSync(path.join(input_directory,'package.json'))){
```
vladimir-kotikov: Can we omit this check? If there is no `package.json` in template directory we'll catch this in `catch (e)` below.

cordova-lib/src/cordova/create.js
```diff
@@ -220,8 +220,22 @@ function create(dir, optionalId, optionalName, cfg) {
             }
         }
     })
-    .then(function(import_from_path) {
-
+    .then(function(input_directory) {
+        var import_from_path = input_directory;
+        var templatePkg;
```
vladimir-kotikov: `templatePkg` is only used within `try` block and doesn't need to be defined in outer scope

vladimir-kotikov:
LGTM apart from two minor nitpicks

csantanapr:
@vladimir-kotikov thanks for the review I will check on those

stevengill:
I don't think you should be running npm install carlos. This code doesn't even check if npm exists before trying to run it. 

I recommend you leaving the npm install to the fetch refactor (https://github.com/cordova/cordova-discuss/pull/33).

I'm not sure about this PR. I know you merged it in already. I think it requires more discussion.

csantanapr:
@stevengill oops good catch, that npm install portion was not suppose to be there, I was just experimenting, only the top portion for the sub dir.

I agree npm install should be part of [cordova/cordova-discuss#33](https://github.com/cordova/cordova-discuss/pull/33) where everything is npm fetch and handles npm install

csantanapr:
@stevengill the npm install portion is reverted now, I will take a look at your PR for npm fetch on how is handling the template fetching and npm install.

stevengill:
still working on integrating it into the template work, but will add it
soon to the pr.

PR #416, opened 2016-03-25
CB-10961 Error when adding ios platform and plugin declare in config.xml

csantanapr:
@vladimir-kotikov can you review 
This other commit https://github.com/apache/cordova-lib/commit/3b9face5187857d354d85f3e3f4316cbace100db 
changing the order to run prepare after plugin install cause the problem. I added a prepare before, and left the prepare on after plugin install.
I don't think there is harm on running prepare twice on platform add

### DIFF
--- a/cordova-lib/spec-cordova/platform.spec.js
+++ b/cordova-lib/spec-cordova/platform.spec.js
@@ -145,7 +145,7 @@ describe('platform end-to-end', function () {
         .fin(done);
     });
 
-    it('should call prepare after plugins were installed into platform', function(done) {
+    it('should call prepare before and after plugins were installed into platform', function(done) {
         var order = '';
         var fail = jasmine.createSpy(fail);
         spyOn(plugman.raw, 'install').andCallFake(function() { order += 'I'; });
@@ -157,7 +157,7 @@ describe('platform end-to-end', function () {
         })
         .fail(fail)
         .fin(function() {
-            expect(order).toBe('IP'); // Install first, then prepare
+            expect(order).toBe('PIP'); // Install first, then prepare
             expect(fail).not.toHaveBeenCalled();
             done();
         });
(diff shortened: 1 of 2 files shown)


riknoll:
Does running prepare twice have any impact on the hooks that fire?

csantanapr:
@riknoll I don't think is big deal if prepare hooks twice, I will look to see if there is a way to call first prepare without hooks

csantanapr:
@riknoll @vladimir-kotikov I found a way to skip prepare hooks on first prepare call. made the change I rebase, please review again

vladimir-kotikov:
@csantanapr, this LGTM and i believe it would solve the problem, but I think the real problem is in cordova-ios `prepare` method - when you change app's name this change doesn't reflected to ios `Api`'s properties (see [prepare.js#L189-L191](https://github.com/apache/cordova-ios/blob/master/bin/templates/scripts/cordova/lib/prepare.js#L189-L191)), and when you call `Api` next time, it uses the old project's path.

It looks like https://github.com/apache/cordova-lib/commit/3b9face5187857d354d85f3e3f4316cbace100db that you mentioned just uncovered the real problem.

If you think this is urgent, we could merge this and do 6.1.1 release but i'd like to fix this in iOS instead.

csantanapr:
@vladimir-kotikov  yes lets fix this in cordova-lib to get a fix in CLI 6.1.1, but I agree with you that problem should also be fix in iOS 

The root of the problem is the merge of XML the <name> value from project config.xml doesn't overwrite the <name> in default platform config.xml <name> I open the JIRA [CB-10965](https://issues.apache.org/jira/browse/CB-10965)
The real problem is cordova-common merge util where i think the tag name should be single instance in result of merge https://github.com/apache/cordova-lib/blob/master/cordova-common/src/util/xml-helpers.js#L197
similar as author and description.
it will require a cordova-common release, then an iOS to pick up the change. But I think fixing in CLI 6.1.1  is required to not brake other ios platforms versions that don't have this fix

vladimir-kotikov:
@csantanapr, i merged a bit different change, which uses `preparePlatforms` method instead of `prepare` .It is a subset of `prepare` but doesn't include the actions that are not expected to ran at this moment, like restoring platforms, plugins, fire hooks. Tested it with the case described in JIRA and it works.

PR #417, opened 2016-03-26
CB-10965 xml helper allows multiple instances to be merge in config.xml

csantanapr:
@vladimir-kotikov I think this is the real fix on duplicate names ending up in final config.xml triggering the unwanted rename of the xcode project in ios


vladimir-kotikov:
LGTM. Didn't have a time to test it but making `name` a singleton definitely makes sense.

cordova-common/spec/util/xml-helpers.spec.js
```diff
@@ -207,6 +207,16 @@ describe('xml-helpers', function(){
             expect(testElements[0].attrib.email).toEqual('dev@cordova.apache.org');
             expect(testElements[0].text).toContain('Apache Cordova Team');
         });
+        
+        it('should merge singelton name without clobber', function () {
```
vladimir-kotikov: `singelton` -> `singleton`

PR #418, opened 2016-03-28
CB-10940 Can't add Android platform from path

daserge:
Added a corresponding test for events.forwardEventsTo

[Jira issue](https://issues.apache.org/jira/browse/CB-10940)

### DIFF
--- a/cordova-common/spec/events.spec.js
+++ b/cordova-common/spec/events.spec.js
@@ -1,0 +1,48 @@
+/**
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+var events = require('../src/events');
+
+describe('forwardEventsTo method', function () {
+    afterEach(function() {
+        events.forwardEventsTo(null);
+    });
+    it('should not go to infinite loop when trying to forward to self', function () {
+        expect(function() {
+            events.forwardEventsTo(events);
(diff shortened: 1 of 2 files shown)


cordova-common/src/events.js
```diff
@@ -42,7 +42,11 @@ module.exports.forwardEventsTo = function (eventEmitter) {
     if (!(eventEmitter instanceof EventEmitter))
         throw new Error('Cordova events could be redirected to another EventEmitter instance only');
 
-    EVENTS_RECEIVER = eventEmitter;
+    // Skipping forwarding to self to avoid infinite recursion.
+    // This is the case when the modules are npm-linked.
+    if(this !== eventEmitter) {
```
vladimir-kotikov: nit: please add space after `if`. Also could you please reference JIRA here so it'd be more clear why we're checking this

vladimir-kotikov: Another question - if there is already another recevier assigned, calling `forwardEventsTo` with `this` instance will leave it unchanged. Is this expected? Maybe we should disable fofrwarding in this case?

cordova-common/spec/events.spec.js
```diff
@@ -0,0 +1,30 @@
...
+
+var events = require('../src/events');
+
+describe('forwardEventsTo method', function () {
+    afterEach(function() {
+        events.forwardEventsTo(null);
+    });
+    it('should not go to infinite loop when trying to forward to self', function () {
+        events.forwardEventsTo(events);
+        events.emit('log', 'test message');
```
vladimir-kotikov: Looks like the expectations is missing here. You'd probably need to use `rewire` and check the value of internal `EVENTS_RECEIVER` variable here.

cordova-common/src/events.js
daserge: Yeah, probably you are right, I'll update it.

cordova-common/spec/events.spec.js
daserge: This test will fail if we go into an infinite recursion as the title says. 
I will add another one for the `EVENTS_RECEIVER` internals check - although would it be better not to use rewire but to verify that forwarding has been reset after `events.forwardEventsTo(events)` call (the behavior change we are discussing below)?

vladimir-kotikov: > This test will fail if we go into an infinite recursion as the title says. 

You mean that this will be caught by Jasmine and reported as failure? Then it makes sense, please add a comment clarifying this.

> to verify that forwarding has been reset after events.forwardEventsTo(events) call

Nice. Let's do it.

daserge:
Addressed, thanks @vladimir-kotikov, please take a look when you have a moment.

vladimir-kotikov:
LGTM

PR #419, opened 2016-03-30
CB-10981 Remove cordova-common from bundled dependencies

vladimir-kotikov:
Since cordova-common has been released to NPM we do not need to bundle it into cordova-lib release packages. See [CB-10981](https://issues.apache.org/jira/browse/CB-10981)


stevengill:
+1

PR #420, opened 2016-03-31
CB-10921 Emit warning in case of plugin restoration failure

vladimir-kotikov:
This PR adds a warning if plugin restore failed due to some reason. See [CB-10921](https://issues.apache.org/jira/browse/CB-10921)

### DIFF
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -156,8 +156,8 @@ module.exports = function plugin(command, targets, opts) {
                             });
 
                             if (missingVariables.length) {
-                                shell.rm('-rf', pluginInfo.dir);
                                 events.emit('verbose', 'Removing ' + pluginInfo.dir + ' due to installation failure');
+                                shell.rm('-rf', pluginInfo.dir);
                                 var msg = 'Variable(s) missing (use: --variable ' + missingVariables.join('=value --variable ') + '=value).';
                                 return Q.reject(new CordovaError(msg));
                             }
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -157,6 +157,7 @@ module.exports = function plugin(command, targets, opts) {
 
                             if (missingVariables.length) {
                                 shell.rm('-rf', pluginInfo.dir);
+                                events.emit('verbose', 'Removing ' + pluginInfo.dir + ' due to installation failure');
                                 var msg = 'Variable(s) missing (use: --variable ' + missingVariables.join('=value --variable ') + '=value).';
                                 return Q.reject(new CordovaError(msg));
                             }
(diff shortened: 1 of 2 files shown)


vladimir-kotikov:
@omefire, could you please take a look.
Also, duplicating comment in JIRA, could you please explain, why we don't fail in case of plugin restore failure. Is this expected behavior?

vladimir-kotikov:
@omefire, ping

omefire:
@vladimir-kotikov taking a look.

cordova-lib/src/cordova/plugin.js
```diff
@@ -157,6 +157,7 @@ module.exports = function plugin(command, targets, opts) {
 
                             if (missingVariables.length) {
                                 shell.rm('-rf', pluginInfo.dir);
+                                events.emit('verbose', 'Removing ' + pluginInfo.dir + ' due to installation failure');
                                 var msg = 'Variable(s) missing (use: --variable ' + missingVariables.join('=value --variable ') + '=value).';
```
omefire: might be better to emit this variable instead

omefire:
LGTM! except for a minor comment.

cordova-lib/src/cordova/plugin.js
vladimir-kotikov: Not sure if i understand you correctly. The `events.emit` at L160 is added because the reporter was confused by the abscence of plugin inside of `<project>/plugins` directory so i think it'd be better if we clarify that we've removed that.

The missing variable is a part of error message and is logged by caller code as a part of rejection handling (on particular in [restore-utilL127](https://github.com/apache/cordova-lib/pull/420/files#diff-e9221ac2a69074d78c7fb9a8e6b9d4d6R127))

omefire: What I'm suggesting is to remove the change at L160 and replacing it by ```events.emit(msg)``` because:
- The plugin isn't getting removed since it hasn't been installed in the first place
- It lets the user know the reason why the plugin won't be installed

vladimir-kotikov: > plugin isn't getting removed since it hasn't been installed

Right, but the previous message in console states something like 'Copying plugin ../source/path => plugin/destination path' (see [fetch.js#L352](https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/plugman/fetch.js#L352)) so we just crarifying that we've copied the folder but now removing it due to restore failure

> replacing it by events.emit(msg), with msg defined at L161

This is already emitted at [restore-util#R127](https://github.com/apache/cordova-lib/pull/420/files#diff-e9221ac2a69074d78c7fb9a8e6b9d4d6R127), why we need to duplicate that?

image

omefire: I now see what you mean. Could you please move the ```events.emit('Removing ...')``` to L159 right before the removal actually occurs ? I hadn't notice that and got confused.
I think we should also emit a message about the variables missing.(```events.emit(msg)```)

sgrebnov: I think if we add the change below then we will have exactly the same message to be traced twice. The pic above with log trace sample looks good.

> I think we should also emit a message about the variables missing.(events.emit(msg))

vladimir-kotikov:
@omefire, addressed your notes and merged

omefire:
:+1:

PR #421, opened 2016-04-06
CB-10975 Allow plugin path to be relative to current directory

vladimir-kotikov:
JIRA: [CB-10975](https://issues.apache.org/jira/browse/CB-10975)


nikhilkh:
LGTM. Is this a regression? Another one to add tests for.

vladimir-kotikov:
I think this is not a regression, since this code has been added recently in #363 (see the diff [here](https://github.com/MSOpenTech/cordova-lib/commit/45a235fa73e5b23ed9fa44734449503976b999e3#diff-5a4020497db170abc9410280ac313316R297)). I'll add the test for this scenario.

PR #422, opened 2016-04-07
CB-10770 Remove cache-min when adding platforms

nikhilkh:
When we release platforms it takes a day for people to be able to use them. This is sub-optimal and removing the timeout to keep the cache fresh makes sense.

### DIFF
--- a/cordova-lib/src/cordova/lazy_load.js
+++ b/cordova-lib/src/cordova/lazy_load.js
@@ -147,10 +147,7 @@ function cordova_npm(platform) {
 // Returns a promise that resolves to directory containing the package.
 function npm_cache_add(pkg) {
     var npm_cache_dir = path.join(util.libDirectory, 'npm_cache');
-    // 'cache-min' is the time in seconds npm considers the files fresh and
-    // does not ask the registry if it got a fresher version.
     var platformNpmConfig = {
-        'cache-min': 3600*24,
         cache: npm_cache_dir
     };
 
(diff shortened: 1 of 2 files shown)


cordova-lib/src/cordova/lazy_load.js
```diff
@@ -147,10 +147,7 @@ function cordova_npm(platform) {
 // Returns a promise that resolves to directory containing the package.
 function npm_cache_add(pkg) {
     var npm_cache_dir = path.join(util.libDirectory, 'npm_cache');
-    // 'cache-min' is the time in seconds npm considers the files fresh and
```
omefire: Consider totally removing calls to 'npm cache add'. Also, are we sure that npm doesn't set some kind of default cache-min ?

nikhilkh: How will it work if I remove the caching altogether - sounds like a design change.

It does get a default `cache-min` of 10 seconds which if it is fine for npm, should be fine for us:
https://github.com/npm/npm/blob/2a5977e0c65b244e92d848fcd56f2f80ba8cdf3b/lib/config/defaults.js#L125. Their default scenario is quite similar to ours.

purplecabbage:
I think Steve is addressing this in his changes to fetch. 
Cc: @stevengill

nikhilkh:
Yes, I'm aware he is removing npm cache completely. This fix is still valuable in the interim till Steve completes the fetch work.

cordova-lib/src/cordova/lazy_load.js
omefire: sounds good.

omefire:
LGTM!

purplecabbage:
LGTM!

PR #423, opened 2016-04-08
CB-11022 Improve performance of `cordova plugin add` command

vladimir-kotikov:
This addresses [CB-11022](https://issues.apache.org/jira/browse/CB-11022) and dramaticaly reduces time required for plugin installation (especially for the projects with a lot of www assets) by skipping prepare after each plugin install.

This change also requires a corresponding update for the platforms that implement PlatformApi contract (see https://github.com/apache/cordova-android/pull/289) - they have to take care of installing (or removing) their js and www files into both `www` and `platform_www` directories and return any non-falsy value as a result of `addPlugin`/`removePlugin` methods

Also a side note: this fix is an interim solution and ideally we should not running `prepare` after plugin installation at all, despite of what platform has reported. Unfortunately this might break some usage scenarios and would put application into broken state when platform is not updated in a proper way (as in https://github.com/apache/cordova-android/pull/289) - just as described in [CB-9617](https://issues.apache.org/jira/browse/CB-9617). I think that `prepare` removal is a good candidate for major release, when some breaking changes are expected.

UPDATE: to test this change i used modified version of `vs-tac` to be able to use custom android platform version instead of pinned one. To do this you need to modify `vs-tac/app.js` and change `return cordova.raw.platform('add', 'android');` -> `return cordova.raw.platform('add', 'local/android/platform');`

### DIFF
--- a/cordova-lib/spec-cordova/plugin.spec.js
+++ b/cordova-lib/spec-cordova/plugin.spec.js
@@ -23,6 +23,8 @@ var helpers = require('./helpers'),
     shell = require('shelljs'),
     events = require('cordova-common').events,
     cordova = require('../src/cordova/cordova'),
+    prepare = require('../src/cordova/prepare'),
+    platforms = require('../src/platforms/platforms'),
     plugman = require('../src/plugman/plugman'),
     registry = require('../src/plugman/registry/registry');
 
@@ -99,6 +101,22 @@ function mockPluginFetch(id, dir) {
     });
 }
 
+function setupPlatformApiSpies() {
+    var api = platforms.getPlatformApi(helpers.testPlatform, path.join(project, 'platforms', helpers.testPlatform));
+    var addPluginOrig = api.addPlugin;
+    var removePluginOrig = api.removePlugin;
+
+    spyOn(api, 'addPlugin').andCallFake(function () {
+        return addPluginOrig.apply(api, arguments)
+        .thenResolve(true);
+    });
+
+    spyOn(api, 'removePlugin').andCallFake(function () {
+        return removePluginOrig.apply(api, arguments)
+        .thenResolve(true);
+    });
+}
+
 describe('plugin end-to-end', function() {
     events.on('results', function(res) { results = res; });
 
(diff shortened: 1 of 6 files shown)


vladimir-kotikov:
@jasongin, @nikhilkh, please take a look when you have a chance

jasongin:
This code is pretty hard to follow. I don't mean your changes, but plugin installation overall is more complex than I expected.

How do these changes affect plugins' config file transformations? Those are normally processed during prepare, right? How would we ever fully eliminate prepare after installing plugins. Don't we need to keep at least the config file transformations?

vladimir-kotikov:
The config file transformations are already handled during both plugin add/rm and prepare. Consider the following code: [Api.js#L233-L237](https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/Api.js#L233-L237). It calls `add_plugin_changes` from [ConfigChanges.js#L128](https://github.com/apache/cordova-lib/blob/master/cordova-common/src/ConfigChanges/ConfigChanges.js#L128) which:
  1. saves these transormations into `<platform>.json` for further usage and
  2.  applies these transformations immediately (see [ConfigChanges.js#L160](https://github.com/apache/cordova-lib/blob/master/cordova-common/src/ConfigChanges/ConfigChanges.js#L160)).

So right now the only goal of calling `prepare` right after `plugin add` is to copy www files.

jasongin:
OK, I didn't realize config file transformations were also done already in plugin add. And I think I have a better understanding of the code now.

My understanding is we still need to do prepare for any platforms which don't do the following things:
 1. Implement the PlatformApi
 2. Copy over the plugins' www files in addPlugin()
 3. Return true from addPlugin()

So do we also need changes to iOS and Windows platforms to do 2 and 3, to match what you did for Android in the linked PR?

vladimir-kotikov:
> do we also need changes to iOS and Windows platforms

Exactly. Changes for windows and iOS is on their go

vladimir-kotikov:
@jasongin, i've added tests to cover the new logic. Is this PR ready to go?

jasongin:
Yes, LGTM

PR #424, opened 2016-04-11
CB-11062 Add --recursive option to git clone command

jfrumar:
Support cordova plugins that make use of git submodules.

ref: https://issues.apache.org/jira/browse/CB-11062


stevengill:
I think I'm going to close this. In cordova-7, we don't use gitclone.js and instead use cordova-fetch which runs npm install under the hood.

raphinesse:
As @stevengill already pointed out: this PR is outdated and frankly I don't see Cordova supporting any installation variants that `npm` does not support out of the box.

Thanks for taking the time to prepare this and sorry for the late and negative answer.

PR #425, opened 2016-04-11
CB-10986: Adding support for scoped npm package plugins

riknoll:
Apparently pre 6.1.0 versions of the CLI were able to add scoped npm packages, but I don't think we ever supported that feature (if it ever worked, I'm sure it was very error prone). This adds support for npm scoped packages. Some notes about the implementation:

* Scoped npm packages, once installed, should be identified by the plugin ID and not the whole scoped package spec. That means if you run `cordova plugin add @scope/cordova-plugin-camera`, the entry that appears in `cordova plugin ls` will be `cordova-plugin-camera` without the scope. If you wanted to remove the plugin, you would run `cordova plugin remove cordova-plugin-camera` (again without the scope)
* Scoped npm packages cannot be installed side by side with their un-scoped counterparts (but I'm not sure why you would want to do that anyway)
* Scoped plugin entries in `config.xml` populate the spec field in this format: `@scope/plugin-id@version` whereas un-scoped plugins just give the version

I am afraid this may conflict somewhat with cordova-fetch (but I am unsure).

### DIFF
--- a/cordova-lib/src/cordova/plugin.js
+++ b/cordova-lib/src/cordova/plugin.js
@@ -29,6 +29,7 @@ var cordova_util  = require('./util'),
     PluginInfoProvider = require('cordova-common').PluginInfoProvider,
     plugman       = require('../plugman/plugman'),
     pluginMapper  = require('cordova-registry-mapper').newToOld,
+    pluginSpec    = require('./plugin_spec_parser'),
     events        = require('cordova-common').events,
     metadata      = require('../plugman/util/metadata'),
     registry      = require('../plugman/registry/registry'),
@@ -199,8 +200,9 @@ module.exports = function plugin(command, targets, opts) {
                                 } else {
                                     var ver = '~' + pluginInfo.version;
                                     // Scoped packages need to have the package-spec along with the version
-                                    if (cordova_util.isScopedRegistryPluginSpec(target)) {
-                                        attributes.spec = cordova_util.parseRegistryPluginSpec(target)[0] + '@' + ver;
+                                    var parsedSpec = pluginSpec.parse(target);
+                                    if (parsedSpec.scope) {
+                                        attributes.spec = parsedSpec.package + '@' + ver;
                                     } else {
                                         attributes.spec = ver;
                                     }
@@ -305,12 +307,11 @@ module.exports = function plugin(command, targets, opts) {
 };
 
 function determinePluginTarget(projectRoot, cfg, target, fetchOptions) {
-    var parts = cordova_util.parseRegistryPluginSpec(target);
+    var parsedSpec = pluginSpec.parse(target);
 
-    var id = parts[0] || target;
-    var version = parts[1];
+    var id = parsedSpec.package || target;
 
-    if (version || cordova_util.isUrl(id) || cordova_util.isDirectory(id)) {
+    if (parsedSpec.version || cordova_util.isUrl(id) || cordova_util.isDirectory(id)) {
         return Q(target);
     }
 
@@ -318,7 +319,7 @@ function determinePluginTarget(projectRoot, cfg, target, fetchOptions) {
     events.emit('verbose', 'No version specified, retrieving version from config.xml');
     var ver = getVersionFromConfigFile(id, cfg);
 
-    if (cordova_util.isUrl(ver) || cordova_util.isDirectory(ver) || cordova_util.isScopedRegistryPluginSpec(ver)) {
+    if (cordova_util.isUrl(ver) || cordova_util.isDirectory(ver) || pluginSpec.parse(ver).scope) {
         return Q(ver);
     }
 
@@ -422,14 +423,12 @@ function getPluginVariables(variables){
 }
 
 function getVersionFromConfigFile(plugin, cfg){
-    if (cordova_util.isScopedRegistryPluginSpec(plugin)) {
-        plugin = cordova_util.extractPluginId(plugin);
-    }
-    var pluginEntry = cfg.getPlugin(plugin);
-    if (!pluginEntry) {
+    var parsedSpec = pluginSpec.parse(plugin);
+    var pluginEntry = cfg.getPlugin(parsedSpec.id);
+    if (!pluginEntry && !parsedSpec.scope) {
         // If the provided plugin id is in the new format (e.g. cordova-plugin-camera), it might be stored in config.xml
         // under the old format (e.g. org.apache.cordova.camera), so check for that.
-        var oldStylePluginId = pluginMapper[plugin];
+        var oldStylePluginId = pluginMapper[parsedSpec.id];
         if (oldStylePluginId) {
             pluginEntry = cfg.getPlugin(oldStylePluginId);
         }
@@ -524,14 +523,14 @@ function getSpec(pluginSource, projectRoot, pluginName) {
     var scopedPackage = null;
     if (pluginSource.hasOwnProperty('id')) {
         // Note that currently version is only saved here if it was explicitly specified when the plugin was added.
-        var parts = cordova_util.parseRegistryPluginSpec(pluginSource.id);
-        version = parts[1];
+        var parsedSpec = pluginSpec.parse(pluginSource.id);
+        version = parsedSpec.version;
         if (version) {
             version = versionString(version);
         }
 
-        if (cordova_util.isScopedRegistryPluginSpec(pluginSource.id)) {
-            scopedPackage = parts[0];
+        if (parsedSpec.scope) {
+            scopedPackage = parsedSpec.package;
         }
     }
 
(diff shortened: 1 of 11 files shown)


riknoll:
@vladimir-kotikov please review

cordova-lib/src/cordova/util.js
```diff
@@ -401,3 +407,48 @@ function getAvailableNpmVersions(module_name) {
         });
     });
 }
+
+/**
+ * Extracts the name and spec from an npm style package spec that follows the
+ * format of (@scope/)?package(@spec)?
+ * @param {String} raw  The raw npm style package spec
+ * @returns {String[]}  An array containing the package name with optional scope
```
vladimir-kotikov: IMO returning an array is not very convenient to use. This leads to guessing, what is the 0th item of the result in the expressions like `cordova_util.parseRegistryPluginSpec(target)[0]`. I'd prefer to use structured result. something like `npm-package-arg` returns (https://github.com/npm/npm-package-arg#examples)

cordova-lib/src/cordova/util.js
```diff
@@ -401,3 +407,48 @@ function getAvailableNpmVersions(module_name) {
...
+
+
+/**
+ * Determines if a string is a package spec of the format @scope/package(@spec)?
+ * or not
+ * @param {String} raw  The raw npm style package spec
+ * @returns {boolean}   True if the given string follows the format of an npm style
+ *                      package spec or false otherwise
+ */
+function isScopedRegistryPluginSpec(raw) {
```
vladimir-kotikov: This function might be unnecessary if `parseRegistryPluginSpec` would return a structured result with e.g. `scope` field, so you can just check if `spec.scope` is defined

cordova-lib/src/cordova/plugin.js
```diff
@@ -410,6 +422,9 @@ function getPluginVariables(variables){
 }
 
 function getVersionFromConfigFile(plugin, cfg){
+    if (cordova_util.isScopedRegistryPluginSpec(plugin)) {
```
vladimir-kotikov: Should we chack for this before calling `extractPluginId`? From what i understand, `extractPluginId` will return id without scope even if plugin spec contains scope expression.

cordova-lib/src/plugman/fetch.js
```diff
@@ -132,8 +132,9 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
                 ));
             }
             // If not found in local search path, fetch from the registry.
-            var splitVersion = plugin_src.split('@');
-            var newID = pluginMapperotn[splitVersion[0]];
+            var splitVersion = cordovaUtil.parseRegistryPluginSpec(plugin_src);
+            var oldId = splitVersion[0] || plugin_src;
+            var newID = pluginMapperotn[oldId];
```
vladimir-kotikov: Since `oldId` could be either package id or scope + package id i'm not sure how this will work with plugin mapper. Should we use `extractPluginId` instead of `parseRegistryPluginSpec` in L135?

cordova-lib/src/plugman/fetch.js
```diff
@@ -205,8 +206,9 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
 // Helper function for checking expected plugin IDs against reality.
 function checkID(expectedIdAndVersion, pinfo) {
     if (!expectedIdAndVersion) return;
-    var expectedId = expectedIdAndVersion.split('@')[0];
-    var expectedVersion = expectedIdAndVersion.split('@')[1];
+    var parts = cordovaUtil.parseRegistryPluginSpec(expectedIdAndVersion);
+    var expectedId = parts[0] || expectedIdAndVersion;
+    var expectedVersion = parts[1];
     if (expectedId != pinfo.id) {
```
vladimir-kotikov: Since you use `parseRegistryPluginSpec`, it is possible that `expectedId` would be specified as `@scope/id`, but from what i understand, pluginInfo.id could not contain scope so this comparison will always be truthy in this case

cordova-lib/src/plugman/install.js
```diff
@@ -90,35 +90,37 @@ module.exports = function installPlugin(platform, project_dir, id, plugins_dir,
 // Returns a promise.
 function possiblyFetch(id, plugins_dir, options) {
     // Split @Version from the plugin id if it exists.
-    var splitVersion = id.split('@');
+    var splitVersion = cordovaUtil.parseRegistryPluginSpec(id);
+    var extractedId = splitVersion[0] || id;
     //Check if a mapping exists for the plugin id
     //if it does, convert id to new name id
-    var newId = pluginMapper.oldToNew[splitVersion[0]];
+    var newId = pluginMapper.oldToNew[extractedId];
```
vladimir-kotikov: As above, should we pass plugin id instead of scoped id to plugin mapper?

cordova-lib/src/plugman/uninstall.js
```diff
@@ -145,10 +145,9 @@ module.exports.uninstallPlugin = function(id, plugins_dir, options) {
         var deps = pluginInfo.getDependencies();
         var deps_path;
         deps.forEach(function (d) {
-            var splitVersion = d.id.split('@');
-            deps_path = path.join(plugin_dir, '..', splitVersion[0]);
+            deps_path = path.join(plugin_dir, '..', cordovaUtil.extractPluginId(d.id));
             if (!fs.existsSync(deps_path)) {
-                var newId = pluginMapper[splitVersion[0]];
+                var newId = pluginMapper[cordovaUtil.parseRegistryPluginSpec(d.id)[0] || d.id];
```
vladimir-kotikov: You use `extractPluginId` in L148 but `parseRegistryPluginSpec` here but they will return different results in case if there is a scoped dependency in plugin.xml. Is that intentional?

vladimir-kotikov:
As a general comment, i'd love to see the methods for dealing with plugin spec returning a structured result instead of an array. This IMO would make the code more readable, as it's really hard to follow, what e.g. `parseRegistryPluginSpec(target)[0]` means.

Also, factoring this into a separate class could help to improve readability. Something like `PluginSpec.parse(target).version`is much more easy to understand, at least personally for me :)

cordova-lib/src/cordova/util.js
riknoll: This function was provided to match the existing checks we do for `isDirectory`, `isUrl`, etc in the plugin code. I'm fine with unifying the API though, since three functions is probably unnecessary.

cordova-lib/src/cordova/plugin.js
riknoll: Good point

riknoll:
@vladimir-kotikov To clarify the plugin mapper comments: I left in the scope with the intention that the mapping would fail if the scope was present. I don't think there is a case for remapping old plugin ids that have a scope added, because if you are giving a scope then you know what npm package you are trying to fetch. We should never remap in that scenario and I'll try and clarify that in the code. As for changing the API, I'm fine with the approach you suggested (separate class, one function, etc.). I think I just dug myself into a hole with the initial way I wrote the code and didn't refactor when I should have :)

riknoll:
@vladimir-kotikov updated to address feedback

cordova-lib/src/cordova/plugin_spec_parser.js
```diff
@@ -0,0 +1,61 @@
...
+    /** @member {String|null} The npm package of the plugin (with scope) or null if this is not a spec for an npm package */
+    this.package = (scope ? scope + id : id) || null;
+}
+
+/**
+ * Tries to parse the given string as an npm-style package specification of
+ * the form (@scope/)?package(@version)? and return the various parts.
+ *
+ * @param {String} raw  The string to be parsed
+ * @param {PluginSpec}  The parsed plugin spec
```
vladimir-kotikov: Probably a copy/paste typo: `@param` -> `@return`

cordova-lib/src/cordova/plugin_spec_parser.js
```diff
@@ -0,0 +1,61 @@
...
+ */
+function parse(raw) {
+    var split = NPM_SPEC_REGEX.exec(raw);
+    if (split) {
+        return new PluginSpec(raw, split[1], split[2], split[3]);
+    } else {
```
vladimir-kotikov: Not an issue, but `else` is not necessary here

vladimir-kotikov:
LGTM

cordova-lib/src/cordova/plugin_spec_parser.js
riknoll: Good catch! Will update

dpogue:
:+1: I've tested and this seems to be working properly. Thanks @riknoll!

riknoll:
No problem, and thanks for testing @dpogue!

PR #426, opened 2016-04-12
CB-11042: Add cordova run option to skip prepare

jasongin:
I'm also submitting a PR for the cordova-cli repo to parse and document the --noprepare option.

### DIFF
--- a/cordova-lib/src/cordova/run.js
+++ b/cordova-lib/src/cordova/run.js
@@ -32,8 +32,10 @@ module.exports = function run(options) {
         var hooksRunner = new HooksRunner(projectRoot);
         return hooksRunner.fire('before_run', options)
         .then(function() {
-            // Run a prepare first, then shell out to run
-            return require('./cordova').raw.prepare(options);
+            if (!options.options.noprepare) {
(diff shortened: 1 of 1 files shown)


cordova-lib/src/cordova/run.js
```diff
@@ -32,8 +32,10 @@ module.exports = function run(options) {
         var hooksRunner = new HooksRunner(projectRoot);
         return hooksRunner.fire('before_run', options)
         .then(function() {
-            // Run a prepare first, then shell out to run
-            return require('./cordova').raw.prepare(options);
+            if (!options.options.noprepare) {
+                // Run a prepare first, then shell out to run
+                return require('./cordova').raw.prepare(options);
+            }
```
omefire: Please add ```return Q();``` in the alternative code path

jasongin: Are you sure that's necessary? Any promise libraries I've used did not require returning another promise when the then function completes synchronously. Even for a similar case in this file, that is not done after the if block on line 72.

nikhilkh:
LGTM

PR #428, opened 2016-04-19
CB-10761 Resore plugins saved without spec attribute

vladimir-kotikov:
This PR updates plugin restore logic to restore plugins that were added to plugin.xml without `spec` attribute. See [CB-10761](https://issues.apache.org/jira/browse/CB-10761) for details

### DIFF
--- a/cordova-lib/src/cordova/restore-util.js
+++ b/cordova-lib/src/cordova/restore-util.js
@@ -112,7 +112,11 @@ function installPluginsFromConfigXML(args) {
         // assume it is the location to install from.
         var pluginSpec = pluginEntry.spec;
         pluginName = pluginEntry.name;
-        var installFrom = semver.validRange(pluginSpec, true) ? pluginName + '@' + pluginSpec : pluginSpec;
+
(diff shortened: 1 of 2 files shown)


cordova-lib/src/cordova/restore-util.js
```diff
@@ -112,7 +112,10 @@ function installPluginsFromConfigXML(args) {
         // assume it is the location to install from.
         var pluginSpec = pluginEntry.spec;
         pluginName = pluginEntry.name;
-        var installFrom = semver.validRange(pluginSpec, true) ? pluginName + '@' + pluginSpec : pluginSpec;
+        var installFrom = (pluginSpec && semver.validRange(pluginSpec, true)) ?
```
nikhilkh: This is a complex ternary - consider break into and if else as it will make it simpler to read.

nikhilkh:
LGTM.

riknoll:
LGTM

cordova-lib/src/cordova/restore-util.js
vladimir-kotikov: Updated

PR #429, opened 2016-04-19
CB-11117: Add FileUpdater module to cordova-common

jasongin:
The new FileUpdater module contains a few functions that enable optimized
file copies by comparing timestamps. These functions are primarily
intended to be used by each platform's prepare operation to avoid
redundantly copying files that haven't changed since the last build, thus
greatly speeding up build times after the first build. But the usfulness
isn't necessarily limited to the prepare operation; the functions could
be used for any Cordova file copy operations that benefit from the same
optimization.

In this initial change, the FileUpdater isn't used yet anywhere. Look for
follow-up changes in other repos which make use of the FileUpdater from
platforms' prepare operations, and a CLI change that enables the
timestamp-based optimization via a new --incremental flag.

### DIFF
--- a/cordova-common/src/FileUpdater.js
+++ b/cordova-common/src/FileUpdater.js
@@ -24,6 +24,9 @@ var path = require("path");
 var shell = require("shelljs");
 var minimatch = require("minimatch");
 
+var isWindows = (process.platform === "win32");
+var child_process = (isWindows ? require("child_process") : null);
+
 /**
  * Logging callback used in the FileUpdater methods.
  * @callback loggingCallback
@@ -52,11 +55,12 @@ var minimatch = require("minimatch");
  *     last-modified times are copied. By default, only files with different times are copied.
  * @param {loggingCallback} [log] Optional logging callback that takes a string message
  *     describing any file operations that are performed.
+ * @param {Object} context Context object for tracking file operations.
  * @return {boolean} true if any changes were made, or false if the force flag is not set
  *     and everything was up to date
  */
-function updatePathWithStats(sourcePath, sourceStats, targetPath, targetStats, options, log) {
-    log = log || function(message) { };
+function updatePathWithStats(
+        sourcePath, sourceStats, targetPath, targetStats, options, log, context) {
     var updated = false;
 
     var rootDir = (options && options.rootDir) || "";
@@ -92,7 +96,7 @@ function updatePathWithStats(sourcePath, sourceStats, targetPath, targetStats, o
             } else if (sourceStats.isFile()) {
                 // The target file does not exist, so it should be copied from the source.
                 log("copy  " + sourcePath + " " + targetPath + (copyAll ? "" : " (new file)"));
-                shell.cp("-f", sourceFullPath, targetFullPath);
+                copyFileWithTime(context, sourceFullPath, sourceStats, targetFullPath);
                 updated = true;
             }
         } else if (sourceStats.isFile() && targetStats.isFile()) {
@@ -100,20 +104,26 @@ function updatePathWithStats(sourcePath, sourceStats, targetPath, targetStats, o
             if (copyAll) {
                 // The caller specified all files should be copied.
                 log("copy  " + sourcePath + " " + targetPath);
-                shell.cp("-f", sourceFullPath, targetFullPath);
+                copyFileWithTime(context, sourceFullPath, sourceStats, targetFullPath);
                 updated = true;
-            } else if (copyNewer && sourceStats.mtime > targetStats.mtime) {
-                // The caller specified only newer files should be copied, and the file is newer.
-                log("copy  " + sourcePath + " " + targetPath + " (newer file)");
-                shell.cp("-f", sourceFullPath, targetFullPath);
-                updated = true;
-            } else if (!copyNewer && sourceStats.mtime != targetStats.mtime) {
-                // The caller specified only different files should be copied, and
-                // the file has a different time; report either newer or older.
-                log("copy  " + sourcePath + " " + targetPath +
-                    (sourceStats.mtime > targetStats.mtime ? " (newer file)" : " (older file)"));
-                shell.cp("-f", sourceFullPath, targetFullPath);
-                updated = true;
+            } else {
+                // Copying should depend on comparison of the files' last-modified times.
+                var sourceTime = sourceStats.mtime.getTime();
+                var targetTime = targetStats.mtime.getTime();
+
+                if (copyNewer && sourceTime > targetTime) {
+                    // The caller specified only newer files should be copied, and the file is newer.
+                    log("copy  " + sourcePath + " " + targetPath + " (newer file)");
+                    copyFileWithTime(context, sourceFullPath, sourceStats, targetFullPath);
+                    updated = true;
+                } else if (!copyNewer && sourceTime !== targetTime) {
+                    // The caller specified only different files should be copied, and
+                    // the file has a different time; report either newer or older.
+                    log("copy  " + sourcePath + " " + targetPath +
+                        (sourceTime > targetTime ? " (newer file)" : " (older file)"));
+                    copyFileWithTime(context, sourceFullPath, sourceStats, targetFullPath);
+                    updated = true;
+                }
             }
         }
     } else if (targetStats) {
@@ -129,6 +139,36 @@ function updatePathWithStats(sourcePath, sourceStats, targetPath, targetStats, o
     }
 
     return updated;
+}
+
+/**
+ * Helper for updatePath and updatePaths functions. Queries stats for source and target
+ * and ensures target directory exists before copying a file.
+ */
+function updatePathInternal(sourcePath, targetPath, options, log, context) {
+    var rootDir = (options && options.rootDir) || "";
+    var targetFullPath = path.join(rootDir, targetPath);
+    var targetStats = fs.existsSync(targetFullPath) ? fs.statSync(targetFullPath) : null;
+    var sourceStats = null;
+
+    if (sourcePath) {
+        // A non-null source path was specified. It should exist.
+        var sourceFullPath = path.join(rootDir, sourcePath);
+        if (!fs.existsSync(sourceFullPath)) {
+            throw new Error("Source path does not exist: " + sourcePath);
+        }
+
+        sourceStats = fs.statSync(sourceFullPath);
+
+        // Create the target's parent directory if it doesn't exist.
+        var parentDir = path.dirname(targetFullPath);
+        if (!fs.existsSync(parentDir)) {
+            shell.mkdir("-p", parentDir);
+        }
+    }
+
+    return updatePathWithStats(
+        sourcePath, sourceStats, targetPath, targetStats, options, log, context);
 }
 
 /**
@@ -160,28 +200,12 @@ function updatePath(sourcePath, targetPath, options, log) {
         throw new Error("A target path is required.");
     }
 
-    var rootDir = (options && options.rootDir) || "";
-    var targetFullPath = path.join(rootDir, targetPath);
-    var targetStats = fs.existsSync(targetFullPath) ? fs.statSync(targetFullPath) : null;
-    var sourceStats = null;
+    log = log || function(message) { };
 
-    if (sourcePath) {
-        // A non-null source path was specified. It should exist.
-        var sourceFullPath = path.join(rootDir, sourcePath);
-        if (!fs.existsSync(sourceFullPath)) {
-            throw new Error("Source path does not exist: " + sourcePath);
-        }
-
-        sourceStats = fs.statSync(sourceFullPath);
-
-        // Create the target's parent directory if it doesn't exist.
-        var parentDir = path.dirname(targetFullPath);
-        if (!fs.existsSync(parentDir)) {
-            shell.mkdir("-p", parentDir);
-        }
-    }
-
-    return updatePathWithStats(sourcePath, sourceStats, targetPath, targetStats, options, log);
+    var context = beginFileOperations();
+    var updated = updatePathInternal(sourcePath, targetPath, options, log, context);
+    endFileOperations(context, log);
+    return updated;
 }
 
 /**
@@ -206,13 +230,18 @@ function updatePaths(pathMap, options, log) {
         throw new Error("An object mapping from target paths to source paths is required.");
     }
 
-    var updated = false;
+    log = log || function(message) { };
 
-    Object.keys(pathMap).forEach(function (targetPath) {
+    var updated = false;
+    var context = beginFileOperations();
+
+    // Iterate in sorted order to ensure directories are created before files under them.
+    Object.keys(pathMap).sort().forEach(function (targetPath) {
         var sourcePath = pathMap[targetPath];
-        updated = updatePath(sourcePath, targetPath, options, log) || updated;
+        updated = updatePathInternal(sourcePath, targetPath, options, log, context) || updated;
     });
 
+    endFileOperations(context, log);
     return updated;
 }
 
@@ -256,6 +285,8 @@ function mergeAndUpdateDir(sourceDirs, targetDir, options, log) {
         throw new Error("A target directory path is required.");
     }
 
+    log = log || function(message) { };
+
     var rootDir = (options && options.rootDir) || "";
 
     var include = (options && options.include) || [ "**" ];
@@ -292,6 +323,7 @@ function mergeAndUpdateDir(sourceDirs, targetDir, options, log) {
     var pathMap = mergePathMaps(sourceMaps, targetMap, targetDir);
 
     var updated = false;
+    var context = beginFileOperations();
 
     // Iterate in sorted order to ensure directories are created before files under them.
     Object.keys(pathMap).sort().forEach(function (subPath) {
@@ -302,9 +334,11 @@ function mergeAndUpdateDir(sourceDirs, targetDir, options, log) {
             entry.targetPath,
             entry.targetStats,
             options,
-            log) || updated;
+            log,
+            context) || updated;
     });
 
+    endFileOperations(context, log);
     return updated;
 }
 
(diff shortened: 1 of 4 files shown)


jasongin:
@nikhilkh @vladimir-kotikov please review

jasongin:
I opened CB-11117 to track this work.

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,389 @@
...
+ * @param {boolean} force If target and source are both files, and the force flag is not
+ *     set, then the file will not be copied unless the source is newer than the target.
+ * @param {function} [log] Optional logging callback that takes a string message describing any
+ *     file operations that are performed.
+ * @return {boolean} true if any changes were made, or false if the force flag is not set
+ *     and everything was up to date
+ */
+function updatePathWithStats(
+        rootDir, targetPath, targetStats, sourcePath, sourceStats, force, log) {
+    log = log || function(message) { };
```
omefire: If no 'log' is specified, we should log to the console.

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,389 @@
...
+var path = require("path");
+var shell = require("shelljs");
+var minimatch = require("minimatch");
+
+/**
+ * Updates a target file or directory with a source file or directory. (Directory updates are
+ * not recursive.) Stats for target and source items must be passed in. This is an internal
+ * helper function used by other methods in this module.
+ *
+ * @param {string|null} rootDir Root directory (such as a project) to which target and source
```
vladimir-kotikov: There is a special notation for optional parameters [offered by JSDoc](http://usejsdoc.org/tags-param.html): `@param {string} [rootDir] Root directory...`. Current approach doesn't work e.g. for VSCode - it can't parse parameter type from JSDoc annotation:

image

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,389 @@
...
+
+    log = log || function(message) { };
+
+    var targetFullPath = path.join(rootDir, targetPath);
+    var targetStats = fs.existsSync(targetFullPath) ? fs.statSync(targetFullPath) : null;
+    var sourceStats = null;
+
+    if (sourcePath) {
+        // A non-null source path was specified. It should exist.
+        var sourceFullPath = path.join(rootDir, sourcePath);
```
vladimir-kotikov: Same as above

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,389 @@
...
+        throw new Error("A target path is required.");
+    }
+
+    if (sourcePath && typeof(sourcePath) !== "string") {
+        throw new Error("A source path (or null) is required.");
+    }
+
+    log = log || function(message) { };
+
+    var targetFullPath = path.join(rootDir, targetPath);
```
vladimir-kotikov: IMO `path.resolve` is more appropriate here, as it will handle correctly the case when someone has passed both `rootDir` and absolute `targetPath`. `path.join` will return incorrect path in this case.

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,389 @@
...
+ *     set, then the file will not be copied unless the source is newer than the target.
+ * @param {function} [log] Optional logging callback that takes a string message describing any
+ *     file operations that are performed.
+ * @return {boolean} true if any changes were made, or false if the force flag is not set
+ *     and everything was up to date
+ */
+function updatePath(rootDir, targetPath, sourcePath, force, log) {
+    rootDir = rootDir || "";
+    if (typeof(rootDir) !== "string") {
+        throw new Error("A root directory path is required.");
```
vladimir-kotikov: According to JSDoc It is not required - could you please rephrase

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,389 @@
...
+ * @param {string|null} sourcePath Source file or directory to be used to update the
+ *     destination. If the source is null, then the destination is deleted if it exists.
+ * @param {boolean} force If target and source are both files, and the force flag is not
+ *     set, then the file will not be copied unless the source is newer than the target.
+ * @param {function} [log] Optional logging callback that takes a string message describing any
+ *     file operations that are performed.
+ * @return {boolean} true if any changes were made, or false if the force flag is not set
+ *     and everything was up to date
+ */
+function updatePath(rootDir, targetPath, sourcePath, force, log) {
```
vladimir-kotikov: A few comments on the method signature:

1. I'd personally prefer this method signature to follow standart semantics for copying files: source arg goes first then destination (e.g. `cp --help`):
```
  Usage: cp [OPTION]... [-T] SOURCE DEST
  or:  cp [OPTION]... SOURCE... DIRECTORY
```
  or `copy /?`
```
  COPY [/D] [/V] [/N] [/Y | /-Y] [/Z] [/L] [/A | /B ] source [/A | /B]
     [+ source [/A | /B] [+ ...]] [destination [/A | /B]]
```
2. It is common practice to put optional arguments after required ones, so probably `rootDir` is better to be placed after `sourcePath` and `targetPath`
3. I'd pass all optional arguments (except `sourcePath`) as an options object (like `updatePath('./foo', './bar', null, true)` vs. `updatePath('./foo', './bar', {force:  true})`)

cordova-common/src/FileUpdater.js
vladimir-kotikov: I'm not sure if we should - this would be accessed programmatically and it would be really odd if it would log something to console when it is not supposed to.

vladimir-kotikov: Also, do we need to pass logging callback to every method? According to the example usage at https://github.com/apache/cordova-android/pull/295 the methods of `FileUpdater ` are called several times so it might make sense to make this class constructable ans set up logging via constructor parameter.

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,389 @@
...
+                shell.mkdir("-p", targetFullPath);
+                updated = true;
+            } else if (sourceStats.isFile()) {
+                // The target file does not exist, so it should be copied from the source.
+                log("copy " + sourcePath + " " + targetPath +
+                    (!force ? " (new file)" : ""));
+                shell.cp("-f", sourceFullPath, targetFullPath);
+                updated = true;
+            }
+        } else if (sourceStats.isFile() && targetStats.isFile() &&
```
vladimir-kotikov: The `if` block below is executed when `targetStats` is defined so probably id would be better to place it right after [line 72](https://github.com/apache/cordova-lib/pull/429/files#diff-8b56245d31f26df256f8361f42457c84R72)

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,389 @@
...
+ * @param {object} pathMap A dictionary mapping from target paths to source paths.
+ * @param {boolean} force If target and source are both files, and the force flag is not
+ *     set, then the file will not be copied unless the source is newer than the target.
+ * @param {function} [log] Optional logging callback that takes a string message describing any
+ *     file operations that are performed.
+ * @return {boolean} true if any changes were made, or false if the force flag is not set
+ *     and everything was up to date
+ */
+function updatePaths(rootDir, pathMap, force, log) {
+    if (!pathMap || typeof(pathMap) !== "object") {
```
vladimir-kotikov: it would not throw if pathMap is an array (`typeof [] === 'object'` is truthy)

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,389 @@
...
+        throw new Error("An object mapping from target paths to source paths is required.");
+    }
+
+    var updated = false;
+
+    for (var targetPath in pathMap) {
```
vladimir-kotikov: It is considered a bad practice to iterate through object keys using `for...in`. Please add check for `hasOwnProperty` or use `Object.keys()`

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,389 @@
...
+/**
+ * Updates a target directory with merged files and subdirectories from source directories.
+ *
+ * @param {string|null} rootDir Root directory (such as a project) to which target and source
+ *     path parameters are relative, or null if the paths are absolute. The rootDir is omitted
+ *     from any logged paths, to make the logs easier to read.
+ * @param {string} targetDir Destination directory to be updated. If it does not exist, it will be
+ *     created. If it exists, newer files from source directories will be copied over, and files
+ *     missing in the source directories will be deleted.
+ * @param {string|string[]} sourceDirs Source directory or array of source directories to be
```
vladimir-kotikov: This doesn't look like the valid `@param` tag According to [JSDoc guide](http://usejsdoc.org/tags-param.html) multiple types should be enclosed in parentesis

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,389 @@
...
+            // Stats obtained here (required at least to know where to recurse in directories)
+            // are saved for later, where the modified times may also be used. This minimizes
+            // the number of file I/O operations performed.
+            var fullPath = path.join(rootDir, subDir, relativePath);
+            var stats = fs.statSync(fullPath);
+
+            // Directories are included if either something under them is included or they
+            // match an include glob. Files are included only if they match an include glob.
+            if (stats.isDirectory() ?
+                    (mapSubdirectory(rootDir, subDir, relativePath, include, exclude, dirMap) ||
```
vladimir-kotikov: Please split this ternary into `if...else` to make it readable

cordova-common/src/FileUpdater.js
vladimir-kotikov: nit: parenthesis are not needed in `typeof` expression

cordova-common/src/FileUpdater.js
jasongin: I was aware of that syntax (and used it for the [log] parameter here). But my interpretation was that it was for parameters that could be omitted by the caller, not merely parameters that could be null. The JSDoc documentation isn't clear about that, but everywhere else I've seen the brackets used was where the parameter could be omitted.

---
In reply to: [60384873](https://github.com/apache/cordova-lib/pull/429#discussion_r60384873) [](ancestors = 60384873)

cordova-common/src/FileUpdater.js
jasongin: Most uses of JSDoc that I've seen omit those parentheses. And VS Code correctly parses it without parentheses.

---
In reply to: [60399853](https://github.com/apache/cordova-lib/pull/429#discussion_r60399853) [](ancestors = 60399853)

cordova-common/src/FileUpdater.js
jasongin: Yes, I'll rearrange the parameters.

---
In reply to: [60396904](https://github.com/apache/cordova-lib/pull/429#discussion_r60396904) [](ancestors = 60396904)

cordova-common/src/FileUpdater.js
jasongin: According to [http://usejsdoc.org/tags-type.html,](http://usejsdoc.org/tags-type.html,) the syntax for a nullable parameter is {?type}, which is distinct from an optional parameter.

---
In reply to: [60448617](https://github.com/apache/cordova-lib/pull/429#discussion_r60448617) [](ancestors = 60448617,60384873)

cordova-common/src/FileUpdater.js
jasongin: Yes, there could easily be cases this is called when no logging is desired. And logging to the console is a bad default for any Cordova code, because any logging should go through the events pipeline.

---
In reply to: [60397076](https://github.com/apache/cordova-lib/pull/429#discussion_r60397076) [](ancestors = 60397076)

jasongin: Not a bad suggestion, but I'm leaning toward keeping it this way. I can easily imagine multiple calls to FileUpdater methods using different logging. (And I'll refactor the code in the other PR so the logging callback code isn't duplicated.)

---
In reply to: [60397531](https://github.com/apache/cordova-lib/pull/429#discussion_r60397531) [](ancestors = 60397531)

cordova-common/src/FileUpdater.js
jasongin: No. Actually the code just before line 72 may set targetStats to null. That's why line 75 is not an "else".

---
In reply to: [60398288](https://github.com/apache/cordova-lib/pull/429#discussion_r60398288) [](ancestors = 60398288)

cordova-common/src/FileUpdater.js
jasongin: Yes, I'm fixing these.

---
In reply to: [60399365](https://github.com/apache/cordova-lib/pull/429#discussion_r60399365) [](ancestors = 60399365)

cordova-common/src/FileUpdater.js
jasongin: I wrote it as in if..else originally, but then the two lines inside the block below were duplicated. But I agree this is very hard to read as it is now, so maybe the duplication is not as bad.

---
In reply to: [60406393](https://github.com/apache/cordova-lib/pull/429#discussion_r60406393) [](ancestors = 60406393)

cordova-common/src/FileUpdater.js
vladimir-kotikov: I see. Is there a reason why we need to emphasize that the `rootDir` is nullable parameter, but not optional one? From user perspective it is not very convenient to pass `null` instead of just omitting the param

jasongin: Because it was the first parameter and later parameters are required, it could not be optional. But it will become optional (actually an optional part of an optional options object) after I rearrange the parameters.

---
In reply to: [60481370](https://github.com/apache/cordova-lib/pull/429#discussion_r60481370) [](ancestors = 60481370)

cordova-common/src/FileUpdater.js
jasongin: I see your point about handling an absolute path, but I don't think I want a relative path to be always resolved to an absolute path. As documented, the caller should not specify a rootDir when using absolute paths.

---
In reply to: [60396878](https://github.com/apache/cordova-lib/pull/429#discussion_r60396878) [](ancestors = 60396878)

jasongin:
@vladimir-kotikov I think this is ready to go after the latest update. Please let me know if you have any further comments.

vladimir-kotikov:
LGTM

jasongin:
I need to add a workaround for the nodejs bug in setting file timestamps on Windows: https://github.com/nodejs/node/issues/2069

nikhilkh:
Simultaneously, lets ask libuv to make the fix for this issue.

jasongin:
@nikhilkh As we discussed, this workaround was the best I could find. The command script approach has similar performance to just copying files via the node fs APIs, and it correctly copies the file times, enabling incremental builds to work correctly. My biggest remaining concern is any errors during the copy will not pinpoint a specific file, but that should be an extremely rare issue.

I also tried using a generated PowerShell script to just update the file times after copying in nodejs, but that was too slow, both in PS startup cost and per-file cost.

cordova-common/src/FileUpdater.js
```diff
@@ -0,0 +1,499 @@
...
+
+    var targetFullPath = path.join(rootDir || "", targetPath);
+
+    if (sourceStats) {
+        var sourceFullPath = path.join(rootDir || "", sourcePath);
+
+        if (targetStats) {
+            // The target exists. But if the directory status doesn't match the source, delete it.
+            if (targetStats.isDirectory() && !sourceStats.isDirectory()) {
+                log("rmdir  " + targetPath + " (source is a file)");
```
nikhilkh: Rest of our code uses event emitter - is there a reason why we don't use that here?

jasongin: Yes. Basically I wanted to enable callers to choose whether they want any logging and what level and format. I expect these functions could be useful in more than one scenario, so logging requirements may be different.

Initially I am passing in events.emit.bind('verbose') as the logging callback to these functions in the updates I'm making to each platform's prepare code. But I know we may want to adjust logging levels and styles soon in the platform code, and it will be easier to do that if we don't have to update cordova-common also.

jasongin:
I updated the PR to use target creation times instead of modified-times. It works much better: it doesn't require the messy workaround on Windows and doesn't have any problems with imprecise timestamps (as on OS X).

jasongin:
Well, after further testing I found creation dates don't work reliably due to some strange behavior on Windows: http://serverfault.com/questions/92757/incorrect-file-creation-date-in-windows-xp-vista

So I'm still working on finding the best solution here.

jasongin:
All right, I think this is the best we can do. The latest logic copies the file if the source's last-modified time is >= the target's last-modified time, or if the file sizes are different. At the time the file gets copied, the target's last-modified time gets set to the current time, so it is greater than the source's time until the source gets modified again.

This should work fine with hooks, because a hook that modifies a file is likely to change the size, or if it doesn't change the size then the hook is likely to handle the file not being refreshed. One could contrive an artificial scenario in which the right thing doesn't happen, but in practice I don't think there would be any issues. And still there will be the 'clean' functionality to fall back on.

purplecabbage:
I would not consider that contrived. 
This would be an extremely tough issue to find if it happened.

jasongin:
@purplecabbage If a hook fails because it expected a file would be refreshed but it wasn't, then I'd expect the hook to print an error (or stack trace), so that should be easy to find. Is there another scenario you're concerned about?

jasongin:
BTW, I didn't mention above the reason I had to abandon the approach of checking for exactly-equal modified times. The problem is the OS X filesystem doesn't support sub-second timestamps, so it would be possible for a file to be modified again in the same second while still appearing to have the same timestamp.

purplecabbage:
Is it possible that a change to a file without affecting file size could go undetected?

jasongin:
If you're talking about a source file, then that would change the last-modified timestamp of the source file, causing it to be newer than the target file so it will be copied on the next prepare.

If you save changes to a source file at precisely the wrong time _while a build is in progress_ (and the change doesn't affect the size of the file), then it's possible the change could go undetected if a hook modifies the corresponding target file to make it newer than the changed source file. But even good old 'make' has this issue.

purplecabbage:
Okay, thanks for clarifying. I rescind my objection.

jasongin:
@nikhilkh waiting on you. Thanks.

jasongin:
As discussed, I tested this with symbolic links both at the root and in a subdirectory. It works fine without any special symlink handling in the code.

nikhilkh:
Great! Thanks @jasongin! Merging now.

PR #430, opened 2016-04-20
CB-11036 'args.slice is not a function' when building windows with ot…

daserge:
…her platform

Avoid options obj modification as it's then passed to other platforms

[Jira issue](https://issues.apache.org/jira/browse/CB-11036)


nikhilkh:
Does this cloning need to happen for other commands too?

daserge:
Updated this in the remaining commands, thanks.
Although it was not failing (because of simultaneous async run via `Q.all`) it makes sense to handle them as well to avoid possible race-condition issues or in case they will become sequential later.

nikhilkh:
LGTM

PR #431, opened 2016-04-21
CB-11131 Fix TypeError: message.toUpperCase is not a function in Cord…

daserge:
…ovaLogger

[Jira issue](https://issues.apache.org/jira/browse/CB-11131)

### DIFF
--- a/cordova-common/src/CordovaLogger.js
+++ b/cordova-common/src/CordovaLogger.js
@@ -89,7 +89,7 @@ CordovaLogger.prototype.log = function (logLevel, message) {
     var isVerbose = this.logLevel === 'verbose';
     var cursor = this.stdoutCursor;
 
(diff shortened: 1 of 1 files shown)


cordova-common/src/CordovaLogger.js
```diff
@@ -209,7 +209,7 @@ function formatError(error, isVerbose) {
         message = error;
     }
 
-    if(message.toUpperCase().indexOf('ERROR:') !== 0) {
+    if(message && typeof message.toUpperCase === 'function' && message.toUpperCase().indexOf('ERROR:') !== 0) {
```
jasongin: This is fine, but I would have preferred `if (typeof message === 'string' && message.toUpperCase()...)`

daserge:
Updated, thanks!

jasongin:
👍

PR #433, opened 2016-04-29
CB-11174 Resolve symlinked path before getting PlatformApi instance

vladimir-kotikov:
This PR the issue when `getPlatformApi` method in cordova-lib returns a different instances for two paths pointing to the same destination, This also fixes 'npm test' failures on OS X.

See JIRA [CB-11174](https://issues.apache.org/jira/browse/CB-11174) for background and details.

### DIFF
--- a/cordova-lib/spec-cordova/platforms/platforms.spec.js
+++ b/cordova-lib/spec-cordova/platforms/platforms.spec.js
@@ -17,8 +17,11 @@
     under the License.
 */
 
+var fs = require('fs');
+var os = require('os');
 var path = require('path');
 var rewire = require('rewire');
+var shell = require('shelljs');
 
 var util = require('../../src/cordova/util');
 var platforms = rewire('../../src/platforms/platforms');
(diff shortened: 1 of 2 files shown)


cordova-lib/spec-cordova/platforms/platforms.spec.js
```diff
@@ -17,19 +17,25 @@
...
 var PLATFORM_WITH_API = path.join(CORDOVA_ROOT, 'platforms/windows');
 var PLATFORM_WOUT_API = path.join(CORDOVA_ROOT, 'platforms/android');
+var PLATFORM_SYMLINK = path.join(os.tmpdir(), 'cordova_windows_symlink');
 
-var MockPlatformApi = require(path.join(PLATFORM_WITH_API, 'cordova', 'Api'));
+var MockPlatformApi = require(fs.realpathSync(path.join(PLATFORM_WITH_API, 'cordova/Api.js')));
 var PlatformApiPoly = require('../../src/platforms/PlatformApiPoly');
 
+shell.ln('-sf', PLATFORM_WITH_API, PLATFORM_SYMLINK);
+
```
jasongin: Have you tested on Windows? (I see the bug is just about OS X.)

jasongin:
:+1:

cordova-lib/spec-cordova/platforms/platforms.spec.js
vladimir-kotikov: I've tested this on both OS X and Windows. Tests are passing in both cases.

stevengill:
LGTM

PR #434, opened 2016-05-03
CB-11194 Improve cordova load time

nikhilkh:
I instrumented module load and this change improves our load time by ~50% (300+ ms) for `cordova build android` on a hello world project. Biggest gains are from delay loading only when required:
- browserify
- unorm
- unpack

```diff
--- a/Users/nikhil/Developer/cordova/require.log
+++ b/Users/nikhil/Developer/cordova/require-after.log
@@ -1,116 +1,64 @@
 > Module load underscore: 4
->>> Module load graceful-fs: 3
->>> Module load osenv: 3
->>> Module load write-file-atomic: 4
->> Module load configstore: 17
->>>> Module load color-convert: 4
+>> Module load child_process: 3
+>>> Module load graceful-fs: 6
+>>> Module load osenv: 4
+>>>>> Module load crypto: 4
+>>>> Module load ./rng: 5
+>>> Module load uuid: 6
+>>> Module load write-file-atomic: 3
+>> Module load configstore: 24
+>>>> Module load color-convert: 5
 >>> Module load ansi-styles: 6
->> Module load chalk: 11
->> Module load semver-diff: 3
-> Module load update-notifier: 36
->> Module load ../hooks/HooksRunner: 3
-> Module load ./build: 3
->>> Module load unorm: 24
->> Module load ../platforms/PlatformApiPoly: 30
->>>>>>>>>>>> Module load graceful-fs: 12
->>>>>>>>>>> Module load ./lib/reader.js: 13
->>>>>>>>>>>>>> Module load minimatch: 3
->>>>>>>>>>>>>> Module load inflight: 3
->>>>>>>>>>>>> Module load glob: 9
->>>>>>>>>>>> Module load rimraf: 10
->>>>>>>>>>> Module load ./lib/writer.js: 15
->>>>>>>>>> Module load fstream: 33
->>>>>>>>> Module load ./entry.js: 35
->>>>>>>> Module load ./entry-writer.js: 37
->>>>>>> Module load ./lib/pack.js: 39
->>>>>> Module load tar: 44
->>>>> Module load ../../util/unpack: 46
->>>> Module load ../plugman/registry/registry: 49
->>> Module load ./plugin: 54
->> Module load ./restore-util: 54
->>>>>> Module load acorn: 6
->>>>> Module load falafel: 7
->>>>> Module load ./loadConfig: 3
->>>> Module load browserify-transform-tools: 13
->>> Module load aliasify: 15
->>>>>>> Module load resolve: 3
->>>>>> Module load browser-resolve: 4
->>>>>>> Module load acorn: 20
->>>>>>>> Module load estraverse: 3
->>>>>>>> Module load esutils: 4
->>>>>>> Module load escodegen: 11
->>>>>> Module load detective: 33
->>>>>>>>>> Module load ./_stream_readable: 3
->>>>>>>>> Module load ./_stream_duplex: 9
->>>>>>>> Module load ./lib/_stream_transform.js: 9
->>>>>>> Module load readable-stream/transform: 10
->>>>>> Module load through2: 12
->>>>>> Module load concat-stream: 3
->>>>>>>>>>> Module load ./_stream_readable: 3
->>>>>>>>>> Module load ./_stream_duplex: 7
->>>>>>>>> Module load ./lib/_stream_transform.js: 7
->>>>>>>> Module load readable-stream/transform: 8
->>>>>>> Module load through2: 8
->>>>>> Module load stream-combiner2: 11
->>>>> Module load module-deps: 70
->>>>> Module load deps-sort: 3
->>>>>> Module load JSONStream: 3
->>>>>>>>>> Module load ./_stream_readable: 3
->>>>>>>>> Module load ./_stream_duplex: 6
->>>>>>>> Module load ./lib/_stream_transform.js: 7
->>>>>>> Module load readable-stream/transform: 8
->>>>>> Module load through2: 8
->>>>>>>>> Module load ./source-map/source-map-generator: 10
->>>>>>>>> Module load ./source-map/source-node: 3
->>>>>>>> Module load source-map: 14
->>>>>>> Module load inline-source-map: 15
->>>>>>>>> Module load ./source-map/source-map-generator: 9
->>>>>>>>> Module load ./source-map/source-map-consumer: 3
->>>>>>>> Module load source-map: 15
->>>>>>> Module load ./lib/mappings-from-map: 16
->>>>>> Module load combine-source-map: 33
->>>>> Module load browser-pack: 46
->>>>>>>> Module load acorn: 18
->>>>>>> Module load astw: 18
->>>>>> Module load lexical-scope: 19
->>>>>>>>> Module load ./source-map/source-map-generator: 8
->>>>>>>>> Module load ./source-map/source-map-consumer: 5
->>>>>>>> Module load source-map: 16
->>>>>>> Module load inline-source-map: 17
->>>>>> Module load combine-source-map: 21
->>>>> Module load insert-module-globals: 43
->>>>>> Module load acorn: 25
->>>>> Module load syntax-error: 26
->>>>> Module load ./lib/builtins.js: 9
->>>>>> Module load stream-splicer: 4
->>>>> Module load labeled-stream-splicer: 5
->>>> Module load browserify: 208
->>> Module load cordova-js/tasks/lib/bundle-browserify: 216
->> Module load ../plugman/browserify: 234
-> Module load ./prepare: 319
->>>>>>>> Module load sax: 3
->>>>>>> Module load ./sax: 4
->>>>>> Module load ./parsers/index: 4
->>>>> Module load ./parser: 5
->>>> Module load elementtree: 9
+>> Module load chalk: 12
+>>> Module load semver: 3
+>> Module load semver-diff: 4
+> Module load update-notifier: 46
+> Module load ./src/platforms/platforms: 4
+>>>>>>> Module load underscore: 4
+>>>>>>>>>>> Module load sax: 3
+>>>>>>>>>> Module load ./sax: 3
+>>>>>>>>> Module load ./parsers/index: 4
+>>>>>>>> Module load ./parser: 4
+>>>>>>> Module load elementtree: 8
+>>>>>> Module load ../util/xml-helpers: 14
+>>>>> Module load ./PluginInfo: 15
+>>>> Module load ./src/PluginInfo/PluginInfoProvider.js: 16
+>>> Module load ./scriptsFinder: 18
+>>>> Module load q: 5
+>>> Module load ./src/superspawn: 8
+>> Module load ../hooks/HooksRunner: 29
+> Module load ./build: 30
+>> Module load ./src/PlatformJson: 5
+>>> Module load shelljs: 6
+>>> Module load ../plugman/pla`tforms/common: 3
+>>> Module load ./src/ActionStack: 3
+>>>> Module load semver: 5
+>>> Module load ./src/ConfigChanges/ConfigChanges.js: 11
+>> Module load ../platforms/PlatformApiPoly: 26
+> Module load ./prepare: 36
+>>>>>>>> Module load sax: 4
+>>>>>>> Module load ./sax: 5
+>>>>>> Module load ./parsers/index: 6
+>>>>> Module load ./parser: 6
+>>>> Module load elementtree: 12
 >>>>> Module load underscore: 5
 >>>> Module load ./src/util/xml-helpers: 7
->>> Module load ./AndroidManifest: 17
+>>> Module load ./AndroidManifest: 21
 >>>> Module load shelljs: 6
 >>> Module load ./pluginHandlers: 8
->> Module load ./lib/AndroidProject: 28
+>> Module load ./lib/AndroidProject: 31
 >>> Module load q: 5
->>> Module load ./PlatformJson: 4
+>>> Module load ./PlatformJson: 5
+>>>> Module load semver: 3
 >>> Module load ./ConfigChanges/ConfigChanges: 5
->> Module load ./src/PluginManager: 19
->> Module load ./src/CordovaLogger: 3
-> Module load /Users/nikhil/Developer/apps/perf/platforms/android/cordova/Api.js: 50
-load: 764ms
+>> Module load ./src/PluginManager: 20
+> Module load /Users/nikhil/Developer/apps/perf/platforms/android/cordova/Api.js: 54
+load: 341ms
 > Module load ./lib/prepare: 5
 ANDROID_HOME=/Users/nikhil/Library/Android/sdk
 JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_51.jdk/Contents/Home
->> Module load ./Adb: 3
-> Module load ./lib/build: 7
+> Module load ./lib/build: 6
+> Module load ./GradleBuilder: 4
 :preBuild UP-TO-DATE
 :preDebugBuild UP-TO-DATE
 :checkDebugManifest
@@ -171,6 +119,6 @@ JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_51.jdk/Contents/Home
 
 BUILD SUCCESSFUL
 
-Total time: 1.027 secs
+Total time: 1.244 secs
 Built the following apk(s): 
 	/Users/nikhil/Developer/apps/perf/platforms/android/build/outputs/apk/android-debug.apk
```

### DIFF
--- a/cordova-common/cordova-common.js
+++ b/cordova-common/cordova-common.js
@@ -17,27 +17,30 @@
     under the License.
 */
 
-exports = module.exports = {
-    events: require('./src/events'),
-    superspawn: require('./src/superspawn'),
+var addProperty = require('./src/util/addProperty');
 
-    ActionStack: require('./src/ActionStack'),
-    CordovaError: require('./src/CordovaError/CordovaError'),
-    CordovaLogger: require('./src/CordovaLogger'),
-    CordovaExternalToolErrorContext: require('./src/CordovaError/CordovaExternalToolErrorContext'),
-    PlatformJson: require('./src/PlatformJson'),
-    ConfigParser: require('./src/ConfigParser/ConfigParser.js'),
-    FileUpdater: require('./src/FileUpdater'),
+module.exports = { };
 
-    PluginInfo: require('./src/PluginInfo/PluginInfo.js'),
-    PluginInfoProvider: require('./src/PluginInfo/PluginInfoProvider.js'),
+addProperty(module, 'events', './src/events');
+addProperty(module, 'superspawn', './src/superspawn');
 
-    PluginManager: require('./src/PluginManager'),
+addProperty(module, 'ActionStack', './src/ActionStack');
+addProperty(module, 'CordovaError', './src/CordovaError/CordovaError');
+addProperty(module, 'CordovaLogger', './src/CordovaLogger');
+addProperty(module, 'CordovaExternalToolErrorContext', './src/CordovaError/CordovaExternalToolErrorContext');
+addProperty(module, 'PlatformJson', './src/PlatformJson');
+addProperty(module, 'ConfigParser', './src/ConfigParser/ConfigParser');
+addProperty(module, 'FileUpdater', './src/FileUpdater');
 
-    ConfigChanges: require('./src/ConfigChanges/ConfigChanges.js'),
-    ConfigKeeper: require('./src/ConfigChanges/ConfigKeeper.js'),
-    ConfigFile: require('./src/ConfigChanges/ConfigFile.js'),
-    mungeUtil: require('./src/ConfigChanges/munge-util.js'),
+addProperty(module, 'PluginInfo', './src/PluginInfo/PluginInfo');
+addProperty(module, 'PluginInfoProvider', './src/PluginInfo/PluginInfoProvider');
 
-    xmlHelpers: require('./src/util/xml-helpers')
-};
+addProperty(module, 'PluginManager', './src/PluginManager');
+
+addProperty(module, 'ConfigChanges', './src/ConfigChanges/ConfigChanges');
+addProperty(module, 'ConfigKeeper', './src/ConfigChanges/ConfigKeeper');
+addProperty(module, 'ConfigFile', './src/ConfigChanges/ConfigFile');
+addProperty(module, 'mungeUtil', './src/ConfigChanges/munge-util');
+
+addProperty(module, 'xmlHelpers', './src/util/xml-helpers');
+
--- a/cordova-common/src/ConfigChanges/ConfigFile.js
+++ b/cordova-common/src/ConfigChanges/ConfigFile.js
@@ -17,13 +17,17 @@
 var fs = require('fs');
 var path = require('path');
 
-var bplist = require('bplist-parser');
-var et   = require('elementtree');
-var glob = require('glob');
-var plist = require('plist');
+var modules = {};
+var addProperty = require('../util/addProperty');
 
-var plist_helpers = require('../util/plist-helpers');
-var xml_helpers = require('../util/xml-helpers');
+// Use delay loading to ensure plist and other node modules to not get loaded
+// on Android, Windows platforms
+addProperty(module, 'bplist', 'bplist-parser', modules);
+addProperty(module, 'et', 'elementtree', modules);
+addProperty(module, 'glob', 'glob', modules);
+addProperty(module, 'plist', 'plist', modules);
+addProperty(module, 'plist_helpers', '../util/plist-helpers', modules);
+addProperty(module, 'xml_helpers', '../util/xml-helpers', modules);
 
 /******************************************************************************
 * ConfigFile class
@@ -67,7 +71,7 @@ function ConfigFile_load() {
     // the same in a future release
     if (ext == '.xml' || ext == '.appxmanifest') {
         self.type = 'xml';
-        self.data = xml_helpers.parseElementtreeSync(filepath);
+        self.data = modules.xml_helpers.parseElementtreeSync(filepath);
     } else {
         // plist file
         self.type = 'plist';
@@ -76,8 +80,8 @@ function ConfigFile_load() {
         //       Do we still need to support binary plist?
         //       If yes, use plist.parseStringSync() and read the file once.
         self.data = isBinaryPlist(filepath) ?
-                bplist.parseBuffer(fs.readFileSync(filepath)) :
-                plist.parse(fs.readFileSync(filepath, 'utf8'));
+                modules.bplist.parseBuffer(fs.readFileSync(filepath)) :
+                modules.plist.parse(fs.readFileSync(filepath, 'utf8'));
     }
 }
 
@@ -88,7 +92,7 @@ ConfigFile.prototype.save = function ConfigFile_save() {
     } else {
         // plist
         var regExp = new RegExp('<string>[ \t\r\n]+?</string>', 'g');
-        fs.writeFileSync(self.filepath, plist.build(self.data).replace(regExp, '<string></string>'));
+        fs.writeFileSync(self.filepath, modules.plist.build(self.data).replace(regExp, '<string></string>'));
     }
     self.is_changed = false;
 };
@@ -98,14 +102,14 @@ ConfigFile.prototype.graft_child = function ConfigFile_graft_child(selector, xml
     var filepath = self.filepath;
     var result;
     if (self.type === 'xml') {
-        var xml_to_graft = [et.XML(xml_child.xml)];
-        result = xml_helpers.graftXML(self.data, xml_to_graft, selector, xml_child.after);
+        var xml_to_graft = [modules.et.XML(xml_child.xml)];
+        result = modules.xml_helpers.graftXML(self.data, xml_to_graft, selector, xml_child.after);
         if ( !result) {
             throw new Error('grafting xml at selector "' + selector + '" from "' + filepath + '" during config install went bad :(');
         }
     } else {
         // plist file
-        result = plist_helpers.graftPLIST(self.data, xml_child.xml, selector);
+        result = modules.plist_helpers.graftPLIST(self.data, xml_child.xml, selector);
         if ( !result ) {
             throw new Error('grafting to plist "' + filepath + '" during config install went bad :(');
         }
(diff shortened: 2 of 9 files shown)


cordova-common/cordova-common.js
```diff
@@ -3,7 +3,7 @@
     or more contributor license agreements.  See the NOTICE file
     distributed with this work for additional information
     regarding copyright ownership.  The ASF licenses this file
-    to you under the Apache License, Version 2.0 (the
+    to you under the Apache License } Version 2.0 (the
```
jasongin: fix this

cordova-common/cordova-common.js
```diff
@@ -17,26 +17,40 @@
...
-};
+function addProperty(obj, property, modulePath) {
+    var val = null;
+
+    // Add properties as getter to delay load the modules on first invocation
+    Object.defineProperty(obj, property, {
+        configurable: true,
+        get: function () {
+            val = val || require(modulePath);
+            ob[property] = val;
```
jasongin: Typo, should be 'obj', not 'ob' ?

jasongin: Also, there should be no need to use the 'val' cache if you are overwriting obj[property].

cordova-common/cordova-common.js
```diff
@@ -17,26 +17,40 @@
...
-    PluginManager: require('./src/PluginManager'),
-
-    ConfigChanges: require('./src/ConfigChanges/ConfigChanges.js'),
-    ConfigKeeper: require('./src/ConfigChanges/ConfigKeeper.js'),
-    ConfigFile: require('./src/ConfigChanges/ConfigFile.js'),
-    mungeUtil: require('./src/ConfigChanges/munge-util.js'),
-
-    xmlHelpers: require('./src/util/xml-helpers')
-};
+function addProperty(obj, property, modulePath) {
```
jasongin: Name this something more specific than addProperty, because it only adds a specific kind of required-module property.

cordova-common/src/ConfigChanges/ConfigFile.js
```diff
@@ -17,13 +17,30 @@
 var fs = require('fs');
 var path = require('path');
 
-var bplist = require('bplist-parser');
-var et   = require('elementtree');
-var glob = require('glob');
-var plist = require('plist');
+var modules = {};
+
+function addProperty(obj, property, modulePath) {
```
jasongin: Can this be defined once and then required from other sources?

cordova-common/src/superspawn.js
```diff
@@ -22,7 +22,7 @@ var fs = require('fs');
 var path = require('path');
 var _ = require('underscore');
 var Q = require('q');
-var shell = require('shelljs');
+var which = require('shelljs/src/which');
```
jasongin: It looks like this relies on shelljs internal layout that may change in a future version.

cordova-common/src/ConfigChanges/ConfigFile.js
jasongin: Perhaps add an optional 4th parameter which is the name of a property on the required module.

cordova-lib/src/platforms/PlatformApiPoly.js
```diff
@@ -474,7 +473,12 @@ function getCreateArgs(destinationDir, projectConfig, options) {
     // CB-6992 it is necessary to normalize characters
     // because node and shell scripts handles unicode symbols differently
     // We need to normalize the name to NFD form since iOS uses NFD unicode form
-    args.push(platformName == 'ios' ? unorm.nfd(projectConfig.name()) : projectConfig.name());
+    var name = projectConfig.name();
+    if (platformName == 'ios') {
+        var unorm = require('unorm');
+        name = unorm.nfd(projectConfig.name());
```
jasongin: `name = unorm.nfd(name);`

cordova-common/src/superspawn.js
jasongin: Also, does this really save any time? Lots of sources require shelljs, so I'd expect it to be fully loaded somewhere else anyway.

nikhilkh: this eliminates multiple copies of `shelljs` - since every module gets it's own version. There are two versions of cordova-common loaded and hence two shelljs modules.

If we do update the version of shelljs and this breaks - it would be very obvious to fix. I have a PR in the works to submit to shelljs to delay load on their side.

cordova-common/cordova-common.js
nikhilkh: This was a last minute change - fixed already - yes, I can remove 'val'

cordova-common/src/ConfigChanges/ConfigFile.js
rakatyal: +1

cordova-lib/cordova-lib.js
```diff
@@ -18,19 +18,35 @@
 */
 
 // For now expose plugman and cordova just as they were in the old repos
+
+
+function addProperty(obj, property, modulePath) {
```
vladimir-kotikov: Similar functionality already exists at [`src/cordova/util.js:339`](https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/cordova/util.js#L339). There is a slight difference but it looks like it could be reused.

jasongin: Nice, that should definitely be consolidated. It will have to be moved into cordova-common though.

nikhilkh: util.js has a side effect in global code of invoking a mkrdir. Also, the addModuleProperty is not quite the same and it's used in a number of places. I have already created a shared version in cordova-common.

nikhilkh:
Looking to merge this by EOD today if there are no other concerns.

cordova-common/src/superspawn.js
```diff
@@ -22,20 +22,20 @@ var fs = require('fs');
...
     var winExtensions = ['.exe', '.bat', '.cmd', '.js', '.vbs'];
     function isValidExe(c) {
         return winExtensions.indexOf(path.extname(c)) !== -1 && fs.existsSync(c);
     }
     if (isValidExe(cmd)) {
         return cmd;
     }
-    cmd = shell.which(cmd) || cmd;
+    cmd = which({}, cmd) || cmd;
     if (!isValidExe(cmd)) {
```
jasongin: I think this won't work now. The non-wrapped `which()` returns a `ShellString` object, then `path.extname()` throws if `(typeof path !== 'string')`.
Did you test on Windows?

PR #436, opened 2016-05-06
CB-11134: Removing duplicated warning

riknoll:


omefire:
LGTM!

omefire:
(Though not related to this change) it seems like the ```--force``` flag wouldn't work as advertised ?

riknoll:
@omefire you are correct. That flag seems to be totally broken

omefire:
Ok. I've created an issue to track it here: https://issues.apache.org/jira/browse/CB-11219

PR #440, opened 2016-05-13
CB-11261 Cut out '-nightly' prerelease tag when checking plugin engines

vladimir-kotikov:
See the background and explanation at [CB-11261](https://issues.apache.org/jira/browse/CB-11261)

This PR is required for tests to pass on nightly builds

### DIFF
--- a/cordova-lib/src/plugman/install.js
+++ b/cordova-lib/src/plugman/install.js
@@ -134,15 +134,15 @@ function checkEngines(engines) {
         var engine = engines[i];
 
         // This is a hack to allow plugins with <engine> tag to be installed with
-        // engine with '-dev' suffix. It is required due to new semver range logic,
+        // engine with '-dev' or '-nightly' suffixes. It is required due to new semver range logic,
         // introduced in semver@3.x. For more details see https://github.com/npm/node-semver#prerelease-tags.
         //
         // This may lead to false-positive checks, when engine version with dropped
         // suffix is equal to one of range bounds, for example: 5.1.0-dev >= 5.1.0.
         // However this shouldn't be a problem, because this only should happen in dev workflow.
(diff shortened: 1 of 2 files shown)


vladimir-kotikov:
Looks like tests are failing due to not yet published `cordova-common@1.3.0` dependency for `cordova-fetch`

cordova-lib/src/plugman/install.js
```diff
@@ -134,15 +134,15 @@ function checkEngines(engines) {
...
         // introduced in semver@3.x. For more details see https://github.com/npm/node-semver#prerelease-tags.
         //
         // This may lead to false-positive checks, when engine version with dropped
         // suffix is equal to one of range bounds, for example: 5.1.0-dev >= 5.1.0.
         // However this shouldn't be a problem, because this only should happen in dev workflow.
-        engine.currentVersion = engine.currentVersion && engine.currentVersion.replace(/-dev$/, '');
-        if ( semver.satisfies(engine.currentVersion, engine.minVersion) || engine.currentVersion === null ) {
-            // engine ok!
+        engine.currentVersion = engine.currentVersion && engine.currentVersion.replace(/-dev|-nightly.*$/, '');
+        if ( semver.satisfies(engine.currentVersion, engine.minVersion, /*loose=*/true) || engine.currentVersion === null ) {
```
nikhilkh: This needs to be updated too: https://github.com/apache/cordova-lib/blob/f513412a30be34a1e846d1f81deb2350d435447b/cordova-lib/src/plugman/fetch.js#L295

vladimir-kotikov: That code is replacing `-dev` suffix in plugin's version, specified in `plugin.xml`. Until we're not going to publish nightly versions of plugins this should not affect us.

nikhilkh: Ah I see - makes sense.

PR #446, opened 2016-05-31
CB-11349 added --fetch and cordova fetch to create --template

carynbear:


stevengill:
Looks great! I'll merge this in! Thanks @carynbear!!

PR #448, opened 2016-06-02
CB-11164 Allow forced dependent plugin removal and CB-11252 Fix cordova fetch for scoped packages

carynbear:


carynbear:
split this into two separate PRs, excuse my gitting

PR #453, opened 2016-06-05
Support a display name via config.xml

dpogue:
The W3C Widgets spec allows for a `short` attribute on the `<name>` element in config.xml, which is to be used for a display name:  
https://www.w3.org/TR/widgets/#the-short-attribute

This adds a new `shortName` function to ConfigParser, which will return the value of that attribute, falling back to returning the content of the `<name>` element.
The `setShortName` function will attempt to set the attribute on the `<name>` element to the specified value. If the `<name>` element does not exist, it will be created with both the text content and the `short` attribute having the specified value.

I've updated the legacy/compat project metadata helpers here, and will open pull requests against the  Platform API projects. I'm not entirely sure how to best approach that given the API addition to ConfigParser. Maybe it's best to test for the existence of the `shortName` function before calling it?

### DIFF
--- a/cordova-common/spec/ConfigParser/ConfigParser.spec.js
+++ b/cordova-common/spec/ConfigParser/ConfigParser.spec.js
@@ -83,7 +83,20 @@ describe('config.xml parser', function () {
                 cfg.setName('this.is.bat.country');
                 expect(cfg.name()).toEqual('this.is.bat.country');
             });
+
+            describe('short name', function() {
+                it('should default to the app name', function() {
+                    expect(cfg.shortName()).toEqual('Hello Cordova');
+                });
+
+                it('should allow setting the app short name', function() {
+                    cfg.setShortName('Hi CDV');
+                    expect(cfg.name()).toEqual('Hello Cordova');
+                    expect(cfg.shortName()).toEqual('Hi CDV');
+                });
+            });
         });
+
         describe('preference', function() {
             it('Test 010 : should return the value of a global preference', function() {
                 expect(cfg.getPreference('fullscreen')).toEqual('true');
--- a/cordova-common/src/ConfigParser/ConfigParser.js
+++ b/cordova-common/src/ConfigParser/ConfigParser.js
@@ -116,6 +116,16 @@ ConfigParser.prototype = {
         var el = findOrCreate(this.doc, 'name');
         el.text = name;
     },
+    shortName: function() {
+        return this.doc.find('name').attrib['short'] || this.name();
+    },
+    setShortName: function(shortname) {
+        var el = findOrCreate(this.doc, 'name');
+        if (!el.text) {
+            el.text = shortname;
+        }
+        el.attrib['short'] = shortname;
+    },
     description: function() {
         return getNodeTextSafe(this.doc.find('description'));
     },


purplecabbage:
This is cool, I just worry about how we handle localization of the app display name.
All platforms support having multiple sets of metadata for different markets, and I would like to see a solution that handles this as well. The widget spec recommends an xml:lang attrib which I find a bit awkward. Not sure what other options there are.

dpogue:
The `lang` attribute might not be too bad for implementing this... the main problem comes down to how we handle the API for getting those localized names. We probably need to support localized app names and display names, since both can be used in the UI in different places.

Maybe `ConfigParser.localizedNames()` and `localizedShortNames()` could return a JSON object of all the additional localizations like `{"fr": "Bonjour Cordova", "jp": "今日は コルドバ"}`?

macdonst:
@dpogue take a look at this PR https://github.com/apache/cordova-lib/pull/449. It should allow for the editing of anything in AndroidManifest.xml. Currently it works for plugin.xml but the plan is to add config.xml support as well. 

Better to have a generic way to target things in AndroidManifest.xml and pList rather than adding one off features.

dpogue:
@macdonst I would agree if this were platform-specific, or outside the scope of the W3C Widgets spec, but this is a fairly common request in JIRA and the spec supports it reasonably gracefully. I'd rather provide a simple way of setting a display name rather than forcing people to add a bunch of overrides to their config.xml for each platform.

Even with the ability to edit AndroidManifest.xml from config.xml, that actually wouldn't help in this case because it's the strings XML file that needs updating.

macdonst:
@dpogue yeah, I agree that it is a common use case. One reason I bring it up is you may be able to re-use some code that @ktop is landing into cordova-common. Also, I mis-spoke as Karen's changes can edit strings.xml and other xml files as well.

shazron:
Can we merge this in?

macdonst:
@shazron don't think we should as `edit-config` can add/modify the strings.xml file on Android just as easily as AndroidManifest.xml.

jcesarmobile:
I think we should as shortName is part of the W3C Widget spec

dpogue:
I think we should support this in the simplest way possible with a consistent API, as opposed to requiring people to add several unintuitive blocks of `edit-config` to modify various platform-specific files.

stevengill:
Hmm. `edit-config` still hasn't been implemented for `config.xml` and might not for a while. I can see both sides of the argument. 

One of the main goals for `edit-config` was to have a generic way to add elements to config files and not have to create new config elements for every new preference that pops up. Adding this is counter-intuitive to that goal.

On the other hand, for `edit-config` this would require multiple lines to do. One for Android, one for ios, etc. Most users probably wouldn't use/set shortname in this scenario. If this is a feature we want many users to use, it might be better to make it its own preference. 

What are others peoples thoughts? I'm going to hold off including this in 6.5.0 so we can discuss it more. Thoughts @macdonst @purplecabbage @infil00p @shazron?

Side note: Does windows not support a shortname type attribute? I only see additions to iOS & Android in this PR.

dpogue:
Windows has a `WindowsStoreDisplayName` preference, but I'm not clear if it's the same concept or not.

In any case, the proper place for the platform changes should happen in each of the platforms with PlatformApi. I only updated the Android/iOS ones here because it wasn't as clear at the time.

PR #454, opened 2016-06-05
Fix failing cordova-fetch tests

dpogue:


dpogue:
/cc @stevengill 

One test is still disabled, the "create with Git URL --fetch" fails in Appveyor, and then causes the next test ("create with npm package --fetch") to fail as well with an EWAIT trying to delete something in node modules.

cordova-fetch/index.js
```diff
@@ -152,7 +152,7 @@ function trimID(target) {
     //also support scoped packages
     if(target.indexOf('@') != -1) {
         parts = target.split('@');
-        if (parts.length > 1) {
+        if (parts.length > 1 && parts[0] === '') {
```
stevengill: nice addition!

stevengill:
Thanks @dpogue!

PR #455, opened 2016-06-08
CB-10438 Install correct dependency version, requires force to instal…

carynbear:
…l plugin with wrong dependency version installed


carynbear:
@stevengill bump

PR #458, opened 2016-06-24
CB-11412 template support for www folders

carynbear:
@stevengill yay but also need to merge the cordova-cli!


stevengill:
Failures on windows?

PR #462, opened 2016-07-01
CB-8151 util.js creates directory if require()d

purplecabbage:
Defer creating of libDir folder until something actually requests it.
Related: CB-11194


purplecabbage:
Anyone wanna look? @vladimir-kotikov @stevengill ?

vladimir-kotikov:
LGTM

PR #464, opened 2016-07-15
CB-11589: Fix missing plugin files after restore

dpogue:
This avoid running prepare unnecessarily when restoring platforms as part of a prepare. This solves warnings about plugins already being installed, is a performance improvement, and fixes the weird case where I was getting an empty ios.json file after restoring plugins.

One thing that isn't clear to me is why `options.save` gets passed through the restore process. Restore is, by definition, adding stuff that has already been saved, so there should be no case in which that ever saves new platforms/plugins to config.xml?

/cc @vladimir-kotikov @riknoll @stevengill

### DIFF
--- a/cordova-lib/src/cordova/platform.js
+++ b/cordova-lib/src/cordova/platform.js
@@ -193,22 +193,26 @@ function addHelper(cmd, hooksRunner, projectRoot, targets, opts) {
 
                 return promise()
                 .then(function () {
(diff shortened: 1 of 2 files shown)


vladimir-kotikov:
LGTM! @dpogue, IMO 'save' part of 'prepare' also could be hidden behind `restoring` flag - I don't see much sense in this logic either.

dpogue:
Thanks @vladimir-kotikov. I've updated to avoid passing `save` through when restoring.

PR #465, opened 2016-07-20
CB-11605 Non-destructive platform update

purplecabbage:
backup platformId@oldVersion before updating


csantanapr:
LGTM
Maybe print some message about next steps "you need to manually moved any platform configuration from backup version a.b.c to x.y.z as you see it fit for your project"

purplecabbage:
Let there be tests

shazron:
I ran this PR with node 6.4.0 and got this:
```
Error: ENOENT: no such file or directory, stat '/Users/shazron/Documents/git/apache/cordova-lib/cordova-lib/spec-cordova/temp/testApp/platforms/android'
    at Error (native)
    at Object.fs.statSync (fs.js:987:18)
    at /Users/shazron/Documents/git/apache/cordova-lib/cordova-lib/node_modules/shelljs/src/ls.js:99:68
    at Array.forEach (native)
    at _ls (/Users/shazron/Documents/git/apache/cordova-lib/cordova-lib/node_modules/shelljs/src/ls.js:65:9)
    at /Users/shazron/Documents/git/apache/cordova-lib/cordova-lib/node_modules/shelljs/src/common.js:111:7
    at Array.forEach (native)
    at Object.expand (/Users/shazron/Documents/git/apache/cordova-lib/cordova-lib/node_modules/shelljs/src/common.js:94:8)
    at Object._cp (/Users/shazron/Documents/git/apache/cordova-lib/cordova-lib/node_modules/shelljs/src/cp.js:155:20)
    at Object.cp (/Users/shazron/Documents/git/apache/cordova-lib/cordova-lib/node_modules/shelljs/src/common.js:186:23)
```
The folder is definitely not there:
screen shot 2016-09-15 at 11 08 29 pm

purplecabbage:
Yeah. I think the test attempts to call platforms/android/cordova/update on the old platform while I am expecting a create call. 
Curiously though it passes on Travis for newer node versions, so there may be some npm link differences in our environments.

purplecabbage:
Yeah, I think I'll have to do more than just that, since it's 2 years old.

raphinesse:
@purplecabbage Is updating a platform something that we actually need? If so why? Neither `ios` nor `android` support it at the moment:

https://github.com/apache/cordova-ios/blob/46da4bc34495ec7bfe13c1ec1b4074dd2e82641a/bin/lib/create.js#L241-L253

https://github.com/apache/cordova-android/blob/46a036ef265951d5371936b37d7296fa47fb28ce/bin/lib/create.js#L347-L360

I guess this boils down to the question if we want to continue supporting the platform-centric workflow, or whatever it's called, where platforms are not pure build-artifacts. Or am I mistaken?

dpogue:
@raphinesse Once upon a time, we tried to actually "update" the platform files. It was a disaster, as you can probably imagine.

Jesse's proposal here was to allow `cordova platform update ios` to make a backup of the existing `platforms/ios` folder, and then essentially run a platform add for the latest version.

dpogue:
@raphinesse Once upon a time I think we did try to actually update platform files. That failed spectacularly, as you can probably imagine.

What Jesse is proposing here is to make `cordova platform update ios` do two things:
1. Make a backup of the existing `platforms/ios` folder.
2. Run a platform add with the latest version of the platform.

This is slightly better than our current instructions of telling users to do a `platform rm`/`platform add` which might cause them to lose any changes they've made to the platform native files.

raphinesse:
@dpogue Thanks for the feedback. I understood what the proposal was about, I just wanted to assert that this is still something we want. I would rather kick out update for good. I never had to manually change stuff in `platforms` (at least not after the arrival of `<edit-config>`). So I usually do a `git clean` followed by installing newer stuff. But I can understand that it's desirable to provide an escape hatch for when things go south.

PR #214, opened 2015-05-11
CB-8978 Added new image element tag for config.xml

ktop:

PR #468, opened 2016-07-27
CB-8978 Add resource-file parsing to config.xml

dpogue:
Original JIRA issue https://issues.apache.org/jira/browse/CB-8978
Related to/replaces https://github.com/apache/cordova-lib/pull/214
See also https://github.com/cordova/cordova-discuss/issues/6

This is just the first pass, we'll also need to update the platform API in each platform to actually copy the resource files around on prepare.

This returns the same data structure as the PluginInfo class. To be determined if we think having top-level `resource-file` elements makes sense, it appears that plugin.xml does not support them.

/cc @ktop @jcesarmobile

### DIFF
--- a/cordova-common/spec/ConfigParser/ConfigParser.spec.js
+++ b/cordova-common/spec/ConfigParser/ConfigParser.spec.js
@@ -281,5 +281,21 @@ describe('config.xml parser', function () {
                 expect(cfg.getStaticResources('android', 'icon').getByDensity('mdpi').src).toBe('logo-android.png');
             });
         });
+
+        describe('file resources', function() {
+            var hasSrcPropertyDefined = function (e) { return !!e.src; };
+            var hasTargetPropertyDefined = function (e) { return !!e.target; };
+            var hasArchPropertyDefined = function (e) { return !!e.arch; };
+
+            it('should fetch platform-specific resources', function() {
+                expect(cfg.getFileResources('android').length).toBe(2);
+            });
+
+            it('should parse resources\' attributes', function() {
+                expect(cfg.getFileResources('android').every(hasSrcPropertyDefined)).toBeTruthy();
+                expect(cfg.getFileResources('android').every(hasTargetPropertyDefined)).toBeTruthy();
+                expect(cfg.getFileResources('windows').every(hasArchPropertyDefined)).toBeTruthy();
+            });
+        });
     });
 });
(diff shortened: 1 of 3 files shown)


stevengill:
So this is the first step in adding `resource-file`.  Future updates to platform API's will use `getFileResources` to do the copying of these files `src`attribute to the `target` attribute. 

I'm not sure I like `null` argument for top level `resource-file`. I agree with you that I don't see much value in adding this as top-level. Maybe developers want to copy a top level `json` file? but why not just include it in `www` at that point? Maybe to copy a sensitive file that isn't under version control? I'm reaching a bit here.

dpogue:
I'm happy to leave out support for top-level `resource-file` tags, since I agree it doesn't really make sense

dpogue:
Any other input here? Are we in agreement to remove the top-level file copying?

ktop:
SGTM

jcesarmobile:
I think top level might make sense here if it's a file that the IDE needs but not the app (configuration file like google-services.json/plist)

Can the launch images and icons be outside the www folder?

dpogue:
For clarity, "top-level support" means allowing `resource-file` tags that aren't scoped to a particular platform.

i.e.,
```xml
<widget ...>
    <resource-file src="foo.xml" target="bar.xml" />
</widget>
```
This would copy `./foo.xml` to `./platforms/ios/bar.xml`, `./platforms/windows/bar.xml`, etc.

It's not clear if there's a use case for files that get copied into every platform.

---

What will definitely be supported is platform-scoped `resource-file` tags:
```xml
<widget ...>
    <platform name="android">
        <resource-file src="gservice.json" target="google-services.json" />
        <resource-file src="icons/notification-mdpi.png" target="res/drawable-mdpi/ic_notification.png" />
    </platform>
</widget>
```
This would copy `./gservices.json` to `./platforms/android/google-services.json` and copy `./icons/notification-mdpi.png` to `./platforms/android/res/drawable-mdpi/ic_notification.png`

This does not require files to go under the www folder, the target can be anywhere in the platform directory (same as with plugin `resource-file` tags)

jcesarmobile:
Then looks good to me, I can't think any example right now where the same file should be needed for all platforms

vladimir-kotikov:
LGTM 👍

shazron:
@stevengill Can I assume support for this feature is out in the cordova@6.5.0 release?

stevengill:
It is! https://cordova.apache.org/news/2017/01/23/tools-release.html

PR #469, opened 2016-07-28
CB-8320: We look for a build.gradle to make sure it's Android, this allows plugman to be used with non-standard Cordova projects.

infil00p:


cordova-lib/src/cordova/metadata/android_parser.js
```diff
@@ -31,14 +31,20 @@ var fs            = require('fs'),
...
         throw new CordovaError('The provided path "' + project + '" is not an Android project.');
     }
 
     // Call the base class constructor
     Parser.call(this, 'android', project);
 
     this.path = project;
+
+    //If we have a gradle file but not a manifest, this is an Android Studio project
+    if(!fs.existsSync(path.join(project, "AndroidManifest.xml"))) {
```
dpogue: @infil00p Apparently this fails the jslint step because of double quotes here, which is why the Travis and Appveyor builds failed. It wants all JS strings to use single quotes.

infil00p: This just reminds me of how much I hate lint and Javascript.

infil00p:
This PR is hilariously out of date, and this functionality was added elsewhere.

PR #470, opened 2016-07-28
CB-11636 Handle attributes with quotes correctly

vladimir-kotikov:
This PR changes `mergeXml` inner implementation to avoid using `elementtree.findall()` with complex queries such as `tag[@attrib1="value1"][@attrib2="value2"]` to avoid issues with attributes w/ double quotes inside. Instead we use `findall('tag')` and the filter out unwanted candidates

For more details see [CB-11636](https://issues.apache.org/jira/browse/CB-11636) and [CB-11613](https://issues.apache.org/jira/browse/CB-11613)


vladimir-kotikov:
@stevengill, @purplecabbage, could you please review?

stevengill:
LGTM

PR #472, opened 2016-07-29
CB-11645: Check for getEditConfigs before invoking it on pluginInfo

riknoll:
Adding a guard to prevent a type error when platforms and CLI have mismatching cordova-common versions. The issue here is that the `PluginInfo` object crosses the platform API barrier between the two so when 5fb8dffb2dc92e4d08d286432f23299da0e812 added `getEditConfigs`, it ended up being a breaking change. This just checks for the function and ignores `edit-config` changes if it is not defined. More permanent fix might be to move its functionality elsewhere so that it does not cross the platform API barrier. Plugins that rely on `edit-config` (which should be none at this point) should [specify the platform and cordova cli version they need](http://cordova.apache.org/docs/en/latest/guide/hybrid/plugins/index.html#specifying-cordova-dependencies) in package.json.


riknoll:
@ktop @stevengill @vladimir-kotikov @TimBarham please take a look

riknoll:
Closing in favor of #473

PR #473, opened 2016-07-30
CB-11645 added check to see if getEditConfig exists before trying to …

stevengill:
…use it

Please review @macdonst @vladimir-kotikov @ktop. I am not familiar enough with these new changes. Does more need to be done?


riknoll:
The only other consideration I can think of off the top of my head is that the edit-config stuff also made changes to the android.json (and other equivalent platform files) for edit-config support. They aren't breaking changes as far as I know because I don't think the information that is saved for the old config-file tag was altered (@ktop to confirm).

I believe the `PluginInfo` object is the only thing used in the platform API that has changed. It might be worth someone else taking a look at the APIs that changed and where they get used, though.

To avoid subtle breaks like this in the future, we should probably better delineate what objects are part of the platform API and what are not. The `PluginInfo` object gets used all over and it isn't immediately obvious that altering it could cause effects like this (especially when using it in other files in the same repo).

vladimir-kotikov:
LGTM

stevengill:
Merged

PR #474, opened 2016-08-01
CB-11653 copied findProjectRoot from cordova-lib

carynbear:
… over

### DIFF
--- a/cordova-common/cordova-common.js
+++ b/cordova-common/cordova-common.js
@@ -27,6 +27,7 @@ addProperty(module, 'superspawn', './src/superspawn');
 addProperty(module, 'ActionStack', './src/ActionStack');
 addProperty(module, 'CordovaError', './src/CordovaError/CordovaError');
 addProperty(module, 'CordovaLogger', './src/CordovaLogger');
+addProperty(module, 'CordovaCheck', './src/CordovaCheck');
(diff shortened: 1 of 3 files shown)


stevengill:
Two things:

1) migrate the tests over and add them to this pr
2) add the apache license header comment to the top. You will notice every other file has this. Apache requirement.

carynbear:
Done

stevengill:
LGTM

stevengill:
Can you create a issue to remove the duplicate code + tests from cordova-lib. That can be a separate PR after cordova-common has been released.

stevengill:
@vladimir-kotikov any issues you see with us moving this from cordova-lib to cordova-common?

carynbear:
@stevengill https://issues.apache.org/jira/browse/CB-11654

vladimir-kotikov:
@stevengill, @carynbear, no issues. Just curious where do you plan to use it outside of cordova-lib :)

carynbear:
I would like to use it in Phonegap, but also it seems like a useful method in general for dealing with Cordova projects. In the spirit of modularizing Cordova, it seems that Plugman would be a potential use case if it is broken out.

vladimir-kotikov:
Ah, I see, thanks!

PR #475, opened 2016-08-02
CB-11652 Update run and emulate to skip build

vladimir-kotikov:
This PR moves handling of `--nobuild` option to CLI. CLI previously has delegated this logic to platform's
`run` method but since introducing `before_deploy` hook we call platform's `build` and `run` separately and now need to handle this option here.

JIRA issue: [CB-11652](https://issues.apache.org/jira/browse/CB-11652)

### DIFF
--- a/cordova-lib/src/cordova/emulate.js
+++ b/cordova-lib/src/cordova/emulate.js
@@ -32,6 +32,10 @@ module.exports = function emulate(options) {
         options.options.device = false;
         options.options.emulator = true;
 
+        var optsClone = _.clone(options.options);
+        // This is needed as .build modifies opts
+        optsClone.nobuild = true;
+
         var hooksRunner = new HooksRunner(projectRoot);
         return hooksRunner.fire('before_emulate', options)
         .then(function() {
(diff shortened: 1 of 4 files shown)


vladimir-kotikov:
@daserge. @sgrebnov, PTAL

cordova-lib/src/cordova/emulate.js
```diff
@@ -42,20 +46,17 @@ module.exports = function emulate(options) {
...
-                        return hooksRunner.fire('before_deploy', options);
-                    })
-                    .then(function() {
-                        optsClone.nobuild = true;
-                        return platform_lib
-                            .getPlatformApi(platform)
-                            .run(optsClone);
-                    });
+
+                var emulatePromise = options.options.nobuild ? Q() :
```
daserge: Should this also be called `buildPromise`?

vladimir-kotikov:
Thanks, @daserge, updated. Is this good to go now?

daserge:
Yes, LGTM :+1:

daserge:
You should also check test failures if they occur after squashing.

PR #476, opened 2016-08-08
CB-11679 Speed up save/restore tests

vladimir-kotikov:
JIRA issue [CB-11679](https://issues.apache.org/jira/browse/CB-11679)


vladimir-kotikov:
@alsorokin, please take a look

alsorokin:
Nice performance increase, LGTM :+1:

PR #478, opened 2016-08-11
CB-11607 breakout cordova-create from cordova-lib

carynbear:
Waiting on [cordova-create](https://github.com/apache/cordova-create) release on npm.
Will need to update package.json.


carynbear:
Not passing tests (with plugins? and platforms?)

stevengill:
Looks like this fails because cordova-create removes all listeners on events.

PR #479, opened 2016-08-12
CB-11698 Fix plugin installation when restoring platform

vladimir-kotikov:
This partially reverts #464 and forces cordova to always check for and install plugins, added to the project, to platform that is being restored at the moment.

JIRA link: [CB-11698](https://issues.apache.org/jira/browse/CB-11698)

### DIFF
--- a/cordova-lib/spec-cordova/save.spec.js
+++ b/cordova-lib/spec-cordova/save.spec.js
@@ -676,6 +676,29 @@ describe('(save flag)', function () {
                 done();
             });
         });
+
+        it('spec.25 should install plugins already added to the project into platform when restoring it', function (done) {
+            var fail = jasmine.createSpy('fail').andCallFake(function (err) {
+                console.log(err.message);
+            });
+
+            cordova.raw.plugin('add', localPluginPath)
+            .then(function () {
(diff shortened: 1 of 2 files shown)


vladimir-kotikov:
@stevengill, @dpogue, please take a look

dpogue:
Honestly, this is just further convincing me that `restore` and `prepare` should be two separate commands. It seems weird to me to run prepare for a platform that isn't installed and expect it to do anything.

The original reason I added this is because when running `cordova prepare` on a project with both iOS and Android in config.xml and empty platforms/plugins dirs, I would get a bunch of warnings logged about plugins already being installed. It seemed like it was trying to install the plugins for both platforms multiple times.

vladimir-kotikov:
> Honestly, this is just further convincing me that restore and prepare should be two separate commands. 

I tend to agree, but we can't change this as I believe this is a part of our CLI contract (I didn't find any proofs, though, apart from [Raymond's blog](https://www.raymondcamden.com/2015/04/21/big-updates-to-apache-cordova/) and [cordova@5.0.0 announcement](http://cordova.apache.org/news/2015/04/21/tools-release.html))

@dpogue, @stevengill, I plan to merge this tomorrow if there is no objections.

PR #480, opened 2016-08-15
CB-11687 platform-specific plugin prepare hooks should be run even wh…

daserge:
…en no platforms are specified

DO NOT MERGE YET, discussion is needed

See details in [Jira](https://issues.apache.org/jira/browse/CB-11687)

### DIFF
--- a/cordova-lib/src/cordova/prepare.js
+++ b/cordova-lib/src/cordova/prepare.js
@@ -37,23 +37,25 @@ function prepare(options) {
     return Q().then(function() {
         var projectRoot = cordova_util.cdProjectRoot();
         var config_json = config.read(projectRoot);
-        options = options || { verbose: false, platforms: [], options: {} };
+        options = options || { verbose: false, options: {} };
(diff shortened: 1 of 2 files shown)


cordova-lib/src/cordova/prepare.js
```diff
@@ -37,23 +37,25 @@ function prepare(options) {
     return Q().then(function() {
         var projectRoot = cordova_util.cdProjectRoot();
         var config_json = config.read(projectRoot);
-        options = options || { verbose: false, platforms: [], options: {} };
+        options = options || { verbose: false, options: {} };
+        options.platforms = options.platforms || cordova_util.listPlatforms(projectRoot);
```
vladimir-kotikov: can we just call `cordova_util.preProcessOptions` here?

daserge: No, because preProcessOptions will throw `No platforms added to this project` on restoring ([save.spec](https://github.com/apache/cordova-lib/blob/2a178662acfad9c779114fddd815e484fc1b6f1d/cordova-lib/spec-cordova/save.spec.js#L613)).

PR #484, opened 2016-08-22
[CB-11730] Modify condition of if clause to avoid similar project nam…

ahirun0426:
…e with plugin name

The original if clause accepts 
 c:\work\SimplePlugin_TestBed with c:\work\SimplePlugin

So I add \ to the plugin directory name for only enable the if clause to complete match the directory name.
 c:\work\SimplePlugin\

nodejs.org description explains the path.resolve removes trailing slashes, the code should work with the situation.

https://nodejs.org/dist/latest-v6.x/docs/api/path.html#path_path_resolve_path
'The resulting path is normalized and trailing slashes are removed unless the path is resolved to the root directory.'

### DIFF
--- a/cordova-lib/src/plugman/fetch.js
+++ b/cordova-lib/src/plugman/fetch.js
@@ -330,7 +330,7 @@ function copyPlugin(pinfo, plugins_dir, link) {
 
     shell.rm('-rf', dest);
 
-    if(!link && dest.indexOf(path.resolve(plugin_dir)) === 0) {
+    if(!link && dest.indexOf(path.resolve(plugin_dir)+'\\') === 0) {
 
         if(/^win/.test(process.platform)) {
             /*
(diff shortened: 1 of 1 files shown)


cordova-lib/src/plugman/fetch.js
```diff
@@ -330,7 +330,7 @@ function copyPlugin(pinfo, plugins_dir, link) {
 
     shell.rm('-rf', dest);
 
-    if(!link && dest.indexOf(path.resolve(plugin_dir)) === 0) {
+    if(!link && dest.indexOf(path.resolve(plugin_dir)+'\\') === 0) {
```
stevengill: Wouldn't this fail on macs/unix?

with this change, on windows it would be `c:\work\SimplePlugin\`, but on mac it would be `~/work/SimplePlugin\`

ahirun0426: Thank you pointed out the problem!
Yes, it should be invalid value for macs/unix.

I should change it to node official separator expression.

PR #485, opened 2016-08-26
CB-11771 Deep symlink directories to target project instead of linking the directory itself

mbektchiev:
When installing a plugin with custom library using the --link option the whole directory is symlinked and temporary
files leak into the original plugin directory on build. This leads to broken builds if the same plugin is linked in
2 projects targeting different Cordova versions.

fixes [CB-11771 Installing plugin with symlink option pollutes original directory with intermediate files](https://issues.apache.org/jira/browse/CB-11771)

related to https://github.com/apache/cordova-ios/pull/244 and https://github.com/apache/cordova-android/pull/326


mbektchiev:
Can someone please review this PR? https://github.com/apache/cordova-ios/pull/244 is already merged and in order to fix the issue we need this one and https://github.com/apache/cordova-android/pull/326 as well.

stevengill:
I confirmed the leaking! Your patch solves the problem. Thanks!

PR #487, opened 2016-08-26
CB-11777: Restore plugins before preparing

dpogue:
When running `cordova prepare` to restore platforms and plugins, the platform has prepare called before the plugins are restored. This leads to the top-level config.xml data being copied into the platform at prepare time, and then plugin config being appended when they are later restored.

In my case, this was causing the Crosswalk version defined in my top-level config.xml to be overwritten by an undefined version when the plugin was installed. A workaround is to run `cordova prepare` a second time.

A better fix is probably to restore the platforms, restore the plugins, and then run prepare after everything has been restored.

All the existing tests pass, but this does change the order of operations around the prepare hooks, and I'm not certain if other projects/plugins have dependencies on that ordering.
/cc @vladimir-kotikov

### DIFF
--- a/cordova-lib/src/cordova/prepare.js
+++ b/cordova-lib/src/cordova/prepare.js
@@ -52,6 +52,9 @@ function prepare(options) {
                 return platforms.getPlatformApi(p, platform_path).getPlatformInfo().locations.www;
             });
             options.paths = paths;
+        }).then(function () {
+            options = cordova_util.preProcessOptions(options);
+            return restore.installPluginsFromConfigXML(options);
         }).then(function() {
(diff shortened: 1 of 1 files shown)


stevengill:
Nice investigating @dpogue! 

Good point about the hook order changing. My gut says it should be fine. Would that be a considered a major change is the question.

stevengill:
Guess this won't matter if we rip restore out of prepare!

https://github.com/cordova/cordova-discuss/pull/54

stevengill:
I'm thinking we move forward with this as I'm feeling hesitant removing restore from prepare.

PR #488, opened 2016-08-31
CB-11517 Allow .folder matches

purplecabbage:


stevengill:
LGTM. Any .folders/files we might want to exclude on purpose?

purplecabbage:
.DS_Store .. but that gets tricky/hacky in a hurry

PR #490, opened 2016-09-16
Android studio

imhotep:
Adds Android Studio support
Adds some logging for when patformApi fails for weird reasons.

### DIFF
--- a/cordova-common/src/ConfigChanges/ConfigFile.js
+++ b/cordova-common/src/ConfigChanges/ConfigFile.js
@@ -193,7 +193,8 @@ function resolveConfigFilePath(project_dir, platform, file) {
         return filepath;
     }
 
-    // FIXME android studio
+    // XXX this checks for android studio projects
+    // only if none of the options above are satisfied does this get called
     if(platform === 'android' && !fs.existsSync(filepath)) {
       filepath = path.join(project_dir, 'app', 'src', 'main', 'res', 'xml', 'config.xml');
     }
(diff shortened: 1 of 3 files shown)


cordova-common/src/ConfigChanges/ConfigFile.js
```diff
@@ -193,6 +193,11 @@ function resolveConfigFilePath(project_dir, platform, file) {
         return filepath;
     }
 
+    // FIXME android studio
+    if(platform === 'android' && !fs.existsSync(filepath)) {
```
purplecabbage: Can you put a comment here to describe how this works and why it is needed.
It looks to me like if we are looking for config.xml and platform is android, the `return filepath;` call on ln:193 is going to return before ever getting here.

cordova-lib/src/platforms/platforms.js
```diff
@@ -60,6 +60,8 @@ function getPlatformApi(platform, platformRootDir) {
             events.emit('warn', ' Using this version of Cordova with older version of cordova-' + platform +
                 ' is being deprecated. Consider upgrading to cordova-' + platform + '@' +
                 platforms[platform].apiCompatibleSince + ' or newer.');
+        } else {
+            events.emit('warn', 'Error loading cordova-'+platform);
```
stevengill: Does this error get shown often?

imhotep: Not a lot. However, whenever there is something wrong with the `PlatformApi`, the error that gets displayed to the user is indicating that they have the wrong version of the platform. In reality though, something could be wrong with the `PlatformApi` code which can be **super** misleading, hence the change. This is a very small change that can help immensely with debugging and troubleshooting errors related to `PlatformApi` code.

PR #491, opened 2016-09-19
CB-11776 Add test case for different edit-config targets

ktop:
An error was thrown when adding two plugins with different edit-config targets. The conflict checking code was missing a case, so it errors out and the second plugin cannot be added. 

This PR is just adding the test case.


ktop:
Was part of this PR https://github.com/apache/cordova-lib/pull/486

stevengill:
Thanks @ktop! Merged!

PR #492, opened 2016-09-26
CB-11908 Add edit-config to config.xml

ktop:
Handles edit-config for config.xml. 

Fixed up the conflict checking so that config.xml will take precedence over plugin.xml changes. 
Added a remove mode to config.xml only, explained more below. 

To test out these changes, you'll need changes in your global cordova-common and remember to copy it to the platform's cordova-common too. 
You will also need this cordova-lib commit: https://github.com/apache/cordova-lib/pull/493

A couple things to discuss/clarify:

1. These changes will make config.xml changes take precedence over plugin.xml changes even if --force is used. Is this alright? Or should --force actually force add the plugin anyways? 

2. There's no real way to "undo" edit-config changes in config.xml. As a fix, I've added a "remove" mode that will remove the attributes specified. Do you think this is useful to have? There's no such thing as restoring/reverting the xml elements to what they were before. What was initially the old list of attributes from before the edit-config change will get overwritten on every prepare.

### DIFF
--- a/cordova-common/src/ConfigChanges/ConfigChanges.js
+++ b/cordova-common/src/ConfigChanges/ConfigChanges.js
@@ -148,6 +148,11 @@ function add_plugin_changes(pluginInfo, plugin_vars, is_top_level, should_increm
     }
     else {
         var isConflictingInfo = is_conflicting(edit_config_changes, platform_config.config_munge, self, plugin_force);
+
+        if (isConflictingInfo.conflictWithConfigxml) {
+            throw new Error(pluginInfo.id +
+                ' cannot be added. <edit-config> changes in this plugin conflicts with <edit-config> changes in config.xml. Conflicts must be resolved before plugin can be added.');
+        }
         if (plugin_force) {
             CordovaLogger.get().log(CordovaLogger.WARN, '--force is used. edit-config will overwrite conflicts if any. Conflicting plugins may not work as expected.');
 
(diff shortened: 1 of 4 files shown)


stevengill:
JSHint is failing for you

ktop:
Thanks, I fixed the errors.

stevengill:
@ktop so if I remove an `edit-config` tag from my `config.xml`, it won't actually remove it from `AndroidManifest.xml`. Guessing the solution for users in this case would be to rm and re-add the android platform. That seems fine to me. 

This is only for merging and overwriting. For adding new elements, we should use `config-file` still?

stevengill:
do you have a usecase example when remove might be useful?

stevengill:
`remove` wouldn't be a revert right. Since you have no way to store what got overwritten/merged

stevengill:
I'm going to merge this in for now so I can get the common release out. But it is definitely worth discussing some of these points.

stevengill:
Thanks for sending this pr @ktop! I've merged it in.

andreszs:
@ktop Please let me know if the plugin can be used to add the attribute `android:installLocation="auto"` into the `manifest` node itself. The docs only explain how to add **child nodes**, and I really tried many ways to add only this attribute only without success. Thank you.

dpogue:
@andreszs Something like this should work:

```xml
<edit-config file="AndroidManifest.xml" target="/*" mode="merge">
    <manifest android:installLocation="auto" />
</edit-config>
```

andreszs:
@dpogue That worked perfectly, thank you!! +10

PR #497, opened 2016-10-07
CB-11976 - Add deprecated node version warning for 0.x

shazron:
### Platforms affected
All

### What does this PR do?
Prints out a warning on `prepare` that you are using a deprecated node version, if you are using any node version 0.x

Tested locally.

### Checklist

### DIFF
--- a/cordova-lib/src/platforms/platforms.js
+++ b/cordova-lib/src/platforms/platforms.js
@@ -29,6 +29,12 @@ var cachedApis = {};
 // PlatformProject classes for now.
 function getPlatformApi(platform, platformRootDir) {
 
+    if (/^v0.\d+[.\d+]*/.exec(process.version)) { // matches v0.* 
+        var msg = 'Warning: using node version ' + process.version +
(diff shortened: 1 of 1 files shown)


shazron:
Making a change so it is shown on every command, not just `prepare`

shazron:
We have to print the warning out in the getPlatformAPI method, somewhere before this line: https://github.com/apache/cordova-lib/blob/f8b58c782c71558516adbbd81929ee31ec3ded7f/cordova-lib/src/platforms/platforms.js#L46

shazron:
Closing. Can't have an effective warn in here, moving to cordova-cli.

vladimir-kotikov:
@shazron, we still have `"node": ">=0.9.9"` in [package.json](https://github.com/apache/cordova-lib/blob/master/cordova-lib/package.json#L16). Might want to update it, so the users will see a warning at install time

shazron:
@vladimir-kotikov good idea, I'll update it to >=4.0.0 since engineStrict is not used anymore in npm@3 -- and we should remove engineStrict anyways. This PR has moved to https://github.com/apache/cordova-cli/pull/262

PR #498, opened 2016-10-08
CB-11985 Check if cached platform/plugin exists before 'npm cache'

TimBarham:
### What does this PR do?
Before calling `npm cache add` when installing a platform or plugin, look for an existing `package.tgz` file in the cache directory. This avoids unnecessary `npm` call with long timeout if not connected to the internet.

Background: First time build failures in Cordova tools in Visual Studio are primarily due to `npm` failures. Therefore we are working to avoid any requirement for `npm` to hit the internet in simple build scenarios. This includes pre-installing cached versions of certain platforms and plugins. But we only get the full benefit if Cordova uses these cached versions if it finds them, without calling npm cache add.

A couple of things to note:
* This change uses shared code for platforms and plugins, and means plugins will also be cached to the `.cordova` directory rather than the global npm cache directory. Is that a concern?
* It is likely this change will be meaningless with the proposed changes to installing platforms and plugins in Cordova 7.x. We'll cross the bridge when we come to it 😄.

Local testing.

### Checklist

### DIFF
--- a/cordova-lib/src/plugman/registry/registry.js
+++ b/cordova-lib/src/plugman/registry/registry.js
@@ -20,10 +20,10 @@
 /* jshint laxcomma:true */
 
 var npm = require('npm'),
-    path = require('path'),
     Q = require('q'),
     npmhelper = require('../../util/npm-helper'),
-    events = require('cordova-common').events;
+    events = require('cordova-common').events,
+    pluginSpec = require('../../cordova/plugin_spec_parser');
 
 module.exports = {
     settings: null,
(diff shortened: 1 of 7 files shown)


cordova-lib/src/util/npm-helper.js
```diff
@@ -72,4 +75,53 @@ function restoreSettings() {
...
+ * Invokes "npm cache add," and then returns a promise that resolves to a directory containing the downloaded,
+ * or cached package.
+ * @param packageName - name of an npm package
+ * @param packageVersion - requested version (not a version range)
+ */
+function cachePackage(packageName, packageVersion) {
+    var cacheDir = path.join(util.libDirectory, 'npm_cache');
+
+    // If already cached, use that rather than calling 'npm cache add' again.
+    var packageCacheDir = path.resolve(cacheDir, packageName, packageVersion);
```
vladimir-kotikov: If `packageVersion` is not provided, this will throw the following error: 'TypeError: Path must be a string. Received undefined'. IMO while this is legitimate error, it probably needs to be wrapped into promise, since this function return a promised result.

cordova-lib/src/plugman/registry/registry.js
```diff
@@ -114,19 +113,11 @@ function initThenLoadSettingsWithRestore(promises) {
...
-            var unpack = require('../../util/unpack');
-            var pluginDir = path.resolve(npm.cache, info.name, info.version, 'package');
-            // Unpack the plugin that was added to the cache (CB-8154)
-            var package_tgz = path.resolve(npm.cache, info.name, info.version, 'package.tgz');
-            return unpack.unpackTgz(package_tgz, pluginDir);
-        });
-    });
+    events.emit('log', 'Fetching plugin "' + plugin + '" via npm');
+    var pluginSplit = plugin.split('@');
+    return npmhelper.fetchPackage(pluginSplit[0], pluginSplit[1]);
```
vladimir-kotikov: Looks like this will not work for scoped NPM packages (like `@scope/cordova-plugin-camera@2.0.0`). `pluginSplit[0]` and `pluginSplit[1]` would become `''` and `'scope/cordova-plugin-camera'` in this case.

cordova-lib/src/util/npm-helper.js
TimBarham: Sure I'll wrap the whole method in a promise.

brodybits:
I would personally favor this change to be "opt-in". I like the idea that npm checks for a newer release before using the cached version of a package.

brodybits:
Thanks @TimBarham for the explanation

cordova-lib/src/plugman/registry/registry.js
TimBarham: Hah, look at that. I didn't know @riknoll had added support for scoped packages. I'll switch this to use `plugin-spec-parser`.

vladimir-kotikov:
Looks good!

PR #504, opened 2016-10-18
CB-11979 added deprecation warning for subdirectories

sterlingann:
### Platforms affected
all

### What does this PR do?
warns of deprecation of subdirectories in Cordova@7

added and removed a subdirectory plugin, saw warning

### Checklist
- [x ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [x ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.

### DIFF
--- a/cordova-lib/src/plugman/fetch.js
+++ b/cordova-lib/src/plugman/fetch.js
@@ -69,7 +69,7 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
                 options.subdir = result[2];
             //if --fetch was used, throw error for subdirectories
         
-        if(option.subdir) {
+        if(options.subdir) {
             events.emit('warn', 'support for subdirectories is depreated and will be removed in Cordova@7');
         if (options.fetch) {
             return Q.reject(new CordovaError('--fetch does not support subdirectories'));
--- a/cordova-lib/src/plugman/fetch.js
+++ b/cordova-lib/src/plugman/fetch.js
@@ -70,6 +70,10 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
             //if --fetch was used, throw error for subdirectories
             if (result[2] && options.fetch) {
                 return Q.reject(new CordovaError('--fetch does not support subdirectories'));
+            } else {
+                if(result[2]) {
+                    events.emit('warn', 'support for subdirectories is depreated and will be removed in Cordova@7');
+                }
             }
 
             // Recurse and exit with the new options and truncated URL.
--- a/cordova-lib/src/plugman/fetch.js
+++ b/cordova-lib/src/plugman/fetch.js
@@ -68,13 +68,13 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
             if (result[2])
                 options.subdir = result[2];
             //if --fetch was used, throw error for subdirectories
-        
-        if(options.subdir) {
-            events.emit('warn', 'support for subdirectories is depreated and will be removed in Cordova@7');
-        if (options.fetch) {
-            return Q.reject(new CordovaError('--fetch does not support subdirectories'));
-            } 
-        }
+
+            if(options.subdir) {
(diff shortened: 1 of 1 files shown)


cordova-lib/src/plugman/fetch.js
```diff
@@ -70,6 +70,10 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
             //if --fetch was used, throw error for subdirectories
             if (result[2] && options.fetch) {
                 return Q.reject(new CordovaError('--fetch does not support subdirectories'));
+            } else {
+                if(result[2]) {
+                    events.emit('warn', 'support for subdirectories is depreated and will be removed in Cordova@7');
+                }
```
shazron: if options.subdir *and* options.fetch are both set, this warning will never be printed. This warning should be emitted in the code block here instead: https://github.com/sterlingann/cordova-lib/blob/497a6d57f5c4ebb324ce6a838dd96a5881a21a75/cordova-lib/src/plugman/fetch.js#L69

stevengill: so, if fetch and subdir -> fail, if subdir && no fetch -> do deprecation notice. If we move it up, users who are using --fetch get deprecation warning & failure. That would be acceptable too.

shazron: Without knowing anything, I was just going by what the PR intended to do: "warns of deprecation of subdirectories". Therefore, if that flag is there, it should print out the warning. If the intent was different (in case of other conditionals), that should be made clear.

purplecabbage: Yes, Sterling and I have resolved this and will push in the morning.
We did a few things:
- used options.subdir defined above instead of result[2] for clarity
- removed nested `else { if() { } }` in favor of `else if() { }`

shazron:
Make sure to run `npm test` before re-submitting your changes. CI fail due to jshint error.
`src/plugman/fetch.js: line 72, col 12, 'option' is not defined.`

cordova-lib/src/plugman/fetch.js
```diff
@@ -68,9 +68,13 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
             if (result[2])
                 options.subdir = result[2];
             //if --fetch was used, throw error for subdirectories
-            if (result[2] && options.fetch) {
-                return Q.reject(new CordovaError('--fetch does not support subdirectories'));
-            }
+        
+        if(options.subdir) {
+            events.emit('warn', 'support for subdirectories is depreated and will be removed in Cordova@7');
+        if (options.fetch) {
```
vladimir-kotikov: Probably an Indentation issue? This should has the same level of indentation as line above

cordova-lib/src/plugman/fetch.js
```diff
@@ -68,9 +68,13 @@ function fetchPlugin(plugin_src, plugins_dir, options) {
             if (result[2])
                 options.subdir = result[2];
             //if --fetch was used, throw error for subdirectories
-            if (result[2] && options.fetch) {
-                return Q.reject(new CordovaError('--fetch does not support subdirectories'));
-            }
+        
+        if(options.subdir) {
+            events.emit('warn', 'support for subdirectories is depreated and will be removed in Cordova@7');
```
sterlingann: Thank you Vladimir!  I'm an intern, this is my first pull request.  I appreciate the feedback :) I will fix it now...

stevengill:
This PR can be closed now! It was merged

yyfearth:
This fix has issue now since 
```
options.subdir = options.subdir || '.';
...
if (result[2]) options.subdir = result[2]; // not work if it is empty
...
if(options.subdir) { // it become always true, since subdir will be "."
```

yyfearth:
I have create RP #511 CB-12261 to fix the issue.

PR #515, opened 2017-01-24
CB-12383 Fix cordova_plugins metadata definition

daserge:
Define parent object first
[Jira issue](https://issues.apache.org/jira/browse/CB-12383)

### Platforms affected
Tested on windows and android with this code:
```
    onDeviceReady: function() {
        this.receivedEvent('deviceready');

        navigator.notification.alert(cordova.require("cordova/plugin_list").metadata);
    },
```

### What does this PR do?
Makes `require("cordova/plugin_list").metadata` also available with browserify.

Manual tests, run cordova-lib auto tests

### Checklist

### DIFF
--- a/cordova-lib/src/plugman/browserify.js
+++ b/cordova-lib/src/plugman/browserify.js
@@ -157,8 +157,8 @@ module.exports = function doBrowserify (project, platformApi, pluginInfoProvider
         // instead of generating intermediate file on FS
         var cordova_plugins = new Readable();
         cordova_plugins.push(
-            'module.exports.metadata = ' + JSON.stringify(pluginMetadata, null, 4) + ';\n' +
-            'module.exports = ' + JSON.stringify(modulesMetadata, null, 4) + ';\n', 'utf8');
+            'module.exports = ' + JSON.stringify(modulesMetadata, null, 4) + ';\n' +
+            'module.exports.metadata = ' + JSON.stringify(pluginMetadata, null, 4) + ';\n', 'utf8');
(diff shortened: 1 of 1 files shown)


cordova-lib/src/plugman/browserify.js
```diff
@@ -157,8 +157,8 @@ module.exports = function doBrowserify (project, platformApi, pluginInfoProvider
         // instead of generating intermediate file on FS
         var cordova_plugins = new Readable();
         cordova_plugins.push(
-            'module.exports.metadata = ' + JSON.stringify(pluginMetadata, null, 4) + ';\n' +
-            'module.exports = ' + JSON.stringify(modulesMetadata, null, 4) + ';\n', 'utf8');
+            'module.exports = ' + JSON.stringify(modulesMetadata, null, 4) + ';\n' +
+            'module.exports.metadata = ' + JSON.stringify(pluginMetadata, null, 4) + ';\n', 'utf8');
```
vladimir-kotikov: So, if I understand correctly, the whole `module.exports` object was being rewritten, so `metadata` property was being lost, right?

daserge: Right. And the behavior differs for normal and browserify cases.
I.e. in normal case when we are injecting a script element the above code did not result in `module.exports` being overwritten.

vladimir-kotikov: Cool, thanks for clarification

vladimir-kotikov:
LGTM

PR #519, opened 2017-02-21
CB-11346 - Remove known platforms check for Platform API

shazron:
### Platforms affected

CLI

### What does this PR do?

Removes platforms check. Right now only known platforms are installable.

None. Unit tests are coming.

### Checklist

### DIFF
--- a/cordova-lib/src/platforms/PlatformApiPoly.js
+++ b/cordova-lib/src/platforms/PlatformApiPoly.js
@@ -55,9 +55,6 @@ function PlatformApiPoly(platform, platformRootDir, events) {
     this.platform = platform;
     this.events = events || require('cordova-common').events;
 
-    if (!(platform in knownPlatforms))
-        throw new CordovaError('Unknown platform ' + platform);
-
     var ParserConstructor = require(knownPlatforms[platform].parser_file);
     this._parser = new ParserConstructor(this.root);
     this._handler = require(knownPlatforms[platform].handler_file);


shazron:
re: https://issues.apache.org/jira/browse/CB-11346?focusedCommentId=15875577&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15875577

Looks like there is more to it than the patch. The error occurs here: https://github.com/apache/cordova-lib/blob/a9605411ab007d700849c971043f0c211c06100c/cordova-lib/src/cordova/platform.js#L389

All three conditions have to be satisfied:
1. version (from package.json)
2. platform name (from package.json name, has to be in the form of cordova-XXXX where XXX is the actual platform name)
3. platform name exists in platformConfig.json file

We would have to remove checks 2 and 3, but not sure of the impact since platformConfig.json contains information that is probably needed elsewhere (ideally it would exist in the platform itself!)

purplecabbage:
Yeah, the deeper pinnings are related to the generic platforms work I've
already done.
I think this is part of the larger goal, I'll try to dig up that work.



@purplecabbage
risingj.com

shazron:
The `src/platforms/platforms.js` file touches almost all commands, particularly the use of the `getPlatformAPI` call, so removing this is tricky -- but at least there is only one API call to abstract away.

purplecabbage:
Yeah, it's deep. If a platform does not have a known name, there will be
issues currently.
The only way right now is to fork a blessed platform, but there is no way
to easily create your own with a generic name.



@purplecabbage
risingj.com

purplecabbage:
We should ultimately just be checking that there is a valid platformAPI,
and if so, use it, otherwise fail.
( super simplification )


@purplecabbage
risingj.com

purplecabbage:
I've started a discuss thread on this subject.
https://github.com/cordova/cordova-discuss/issues/64


@purplecabbage
risingj.com

purplecabbage:
This pr does address part of the issue, I think we can move ahead with it.

PR #549, opened 2017-04-26
CB-12705: Pass plugin info to project-level plugin hooks

feichngr:
### Platforms affected
Cordova-cli and maybe plugman

### What does this PR do?
Makes sure plugin information is passed to project-level `before_plugin_uninstall`, `before_plugin_install`, and `after_plugin_install` hooks.

Other than modifying the test spec for `HooksRunner.js`, I created a small test project with `before_plugin_uninstall`, `before_plugin_install`, and `after_plugin_install` js-hooks configured in `config.xml` which print the context to console. They now output plugin information just like they do for plugin hooks.

### Checklist

### DIFF
--- a/cordova-lib/src/hooks/HooksRunner.js
+++ b/cordova-lib/src/hooks/HooksRunner.js
@@ -178,7 +178,9 @@ function runScriptViaModuleLoader(script, context) {
     }
     var scriptFn = require(script.fullPath);
     context.scriptLocation = script.fullPath;
-    context.opts.plugin = script.plugin;
+    if(script.plugin) {
+        context.opts.plugin = script.plugin;
+    }
 
     // We can't run script if it is a plain Node script - it will run its commands when we require it.
     // This is not a desired case as we want to pass context, but added for compatibility.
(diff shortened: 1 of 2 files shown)


cordova-lib/src/hooks/HooksRunner.js
```diff
@@ -178,7 +178,9 @@ function runScriptViaModuleLoader(script, context) {
     }
     var scriptFn = require(script.fullPath);
     context.scriptLocation = script.fullPath;
-    context.opts.plugin = script.plugin;
+    if(script.plugin) {
+        context.opts.plugin = script.plugin;
```
stevengill: Am I correct in understanding that this is already being passed in? You just add a if statement to confirm it exists before assigning?

@feichngr

feichngr: Yes and no. cordova-lib/src/plugman/install.js already populates `opts.plugin` before it is passed to `HooksRunner`.
What currently happens in this code snippet is that the field is overridden with either information on the plugin which provides the script (i.e. the on in which `plugin.xml` it is defined) or `undefined` if the script is part of the project (i.e. defined in config.xml) because `script.plugin` is `undefined`.
My change just prevents the override in the later case. The first case should behave as before so it doesn't break existing scripts.

stevengill:
merged! Thanks

PR #558, opened 2017-05-16
CB-12766 Consistently write JSON with 2 spaces indentation

remcohaszing:
### Platforms affected
None really.

### What does this PR do?
JSON was written with 4 spaces indentation and even a tab at some
point. Most NodeJS tools write their JSON configuration with an
indentation of 2 spaces. Most notably npm does this, which writes to
package.json, as does Cordova. This caused unnecessary changes in
package.json.

None, this is mostly stylistic.

### Checklist

### DIFF
--- a/cordova-common/spec/PlatformJson.spec.js
+++ b/cordova-common/spec/PlatformJson.spec.js
@@ -112,9 +112,9 @@ describe('PlatformJson class', function() {
             it('Test 009 : should generate text metadata containing list of installed modules', function () {
                 var meta = platformJson.addPluginMetadata(fakePlugin).generateMetadata();
                 expect(typeof meta).toBe('string');
-                expect(meta.indexOf(JSON.stringify(platformJson.root.modules, null, 4))).toBeGreaterThan(0);
-                // expect(meta).toMatch(JSON.stringify(platformJson.root.modules, null, 4));
-                expect(meta).toMatch(JSON.stringify(platformJson.root.plugin_metadata, null, 4));
+                expect(meta.indexOf(JSON.stringify(platformJson.root.modules, null, 2))).toBeGreaterThan(0);
+                // expect(meta).toMatch(JSON.stringify(platformJson.root.modules, null, 2));
+                expect(meta).toMatch(JSON.stringify(platformJson.root.plugin_metadata, null, 2));
             });
         });
     });
--- a/cordova-common/src/PlatformJson.js
+++ b/cordova-common/src/PlatformJson.js
@@ -39,7 +39,7 @@ PlatformJson.load = function(plugins_dir, platform) {
 
 PlatformJson.prototype.save = function() {
     shelljs.mkdir('-p', path.dirname(this.filePath));
-    fs.writeFileSync(this.filePath, JSON.stringify(this.root, null, 4), 'utf-8');
+    fs.writeFileSync(this.filePath, JSON.stringify(this.root, null, 2), 'utf-8');
 };
 
 /**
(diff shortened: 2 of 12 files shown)


stevengill:
really good point. We should switch to two like npm.

piotrowski:
Which files will this affect?

remcohaszing:
@piotrowski I did a regex replace to replace all occurrences of `JSON.stringify`, so basically this affects all JSON files that are modified by Cordova.

I think this is a pretty complete list of the files affected:

- *.cordova/config.json*
- *plugins/fetch.json*
- *plugins/[platform].json*
- *package.json*

mraerino:
Is this going to be merged?
The current behavior really messes up my workflow.

purplecabbage:
+1, let's merge this.

stevengill:
merged

uloco:
When adding platforms or plugins, you also perform a config.xml rewrite. Will this merge affect config.xml indentation rewrites, too? I do use 2 space indentation everywhere. It would be nice if you could just get the default indentation for the file and use the present one. Is this possible or worked on anywhere else?

stevengill:
@uloco want to add your comment to this issue: https://issues.apache.org/jira/browse/CB-12810

uloco:
@stevengill Go for it. I don't have an account there...

uloco:
You should also keep the last line if it's empty in both, `package.json` and `config.xml`.
The [POSIX standard](http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) defines a line like this:

> 3.206 Line
> A sequence of zero or more non- <newline> characters plus a terminating <newline> character.

A lot of us are developing on UNIX systems and are following this practice. It would be nice, if you could help us keeping it consistent.

G-Rath:
@stevengill I really agree with @uloco - Not only for the reason of consistent, but because the constant removal of the last line is picked up by VCSs, which can be really annoying. 

While you've attached uloco's comment to CB-12810, I've noticed that it's been marked as fixed. 

Because of the way the new line removal effects VCSs, I'd consider this a bug (even if a minor one).

While I'm happy to (and normally would just) file a bug report, since you've already commented on it, and there is technically an issue already for it, can I leave it in your hands to make sure this gets considered?

Thanks :)

stevengill:
Hey @G-Rath 

CB-12810 is very much still open :). https://issues.apache.org/jira/browse/CB-12810

Thanks for your comment. This is something we will hopefully get to soon.

funkyvisions:
I don't like that my original package.json and config.xml get overwritten.  I manage them myself and have lots of comments in them to detail intricacies.  I also group things together so that they make logical sense for me.  How can I get cordova to not touch these files?  I tried --nosave on prepare and "cordova config set autosave false".  Let me know if there is a better place we can have this discussion.  This is the only thread I found regarding the overwriting of the configuration files.

janpio:
@funkyvisions You might want to create a new issue for that instead of commenting on old ones.

stevengill:
here is an existing issue https://issues.apache.org/jira/browse/CB-12810

jcesarmobile:
There is also https://issues.apache.org/jira/browse/CB-13009, they are similar, not sure if duplicated

alexreg:
Surely the whitespace settings should simply be maintained, like NPM/Yarn does? i.e. if it's already using tabs, keep tabs, and keep the number of spaces being used.

Simpler1:
Why was this pull request closed and not merged?

remcohaszing:
I think it has been cherry-picked manually. Anyway, it got into master. See https://github.com/apache/cordova-lib/commit/d244cf8442004f2780775cb16f0ac5454b056996

Simpler1:
@remcohaszing Perfect.  Thanks.

PR #561, opened 2017-06-06
Cb 12870 - Check that all use cases are caught in Api.js

audreyso:
Not ready for review yet, still updating tests. Thanks!

### What does this PR do?

Check that all use cases are caught in Api.js

### Checklist


filmaj:
Couple things in this one:

1. This will need a rebase with the latest.
2. I wonder if instead of putting more efforts into tweaking the fixtures and integration tests, we put more effort into converting the integration tests into unit tests wherever it makes sense.

Let me know what you think @audreyso.

audreyso:
@filmaj I rebased this for now. If we change the tests to unit tests, should it be in this PR or in a separate one?

filmaj:
Your experience going through the integration tests in here is valuable with respect to understanding what code these tests exercise, so armed with that knowledge, I think you have the right context to add unit tests for that code (and instead, delete the integration tests relevant to that code).

As for whether it's in this PR or a separate one, your call. There will probably end up being extraneous commits in this PR. I would probably issue a new PR.

PR #562, opened 2017-06-12
Removing lazy load, platform command refactor, start of finer-grained unit tests.

filmaj:
### Platforms affected

all / node ?

### What does this PR do?

A lot!

1. Removed lazy loading of modules.

Per cordova/cordova-discuss#70, it was determined that the obfuscation the lazy loading introduced was not worth the time saved on load (~250ms in the worst case). It also opens the door for _much_ easier unit testing - more on that later.

2. Refactored the platform command

Previously, it was implemented as one 770 lines-of-code module. It is now split up into an `index.js` main module under a `platform/` subdirectory, with the high-level commands split out as sub-modules. This kind of split draws clearer abstraction boundaries, and makes unit testing easier to organize as there are clearer responsibilities divided amongst the modules.

3. Start of fine-grained unit tests for the platform command

Once point 2 above was implemented, the combination of removing lazy loading and splitting up the platform command allowed for a simpler layout of tests: one spec file per module, and mock out everything outside of that one module. In this PR, that was done for the new `platform/index.js` file as a first step. The other platform subcommands still need unit tests written for them.

4. Moved a lot of tests from `spec-cordova/` and `spec-plugman` into `integration-tests`

At least all of the platform command specs that I could identify as relying heavily on the filesystem and network, I moved these into the `integration-tests/` folder. The idea is that as new unit tests are written we can eliminate integration tests that test the same functionality.

5. Rewritten/removed some integration tests that have proper unit test coverage elsewhere
 - `ios-add-plugin-with-cocoapod.spec.js` (and the fixtures that that test relied on) have now moved as part of [increased unit test coverage in cordova-ios](https://github.com/apache/cordova-ios/pull/319).

Only unit tests. A lot of the integration tests are failing at this point - I am unsure if that's a showstopper for this PR or not, looking for feedback on that.

Worth noting that moving more tests into `integration-tests/` dropped the code coverage: from 80% to 63%. However, the platform command still needs to have almost all of its subcommands unit tested, so I think that is bound to go up. As I mentioned above, moving forward, we should review the tests for different commands, refactor them as I did here, and try to rewrite a lot of the integration tests as unit tests instead.

While the code coverage right now has gone down, so has test execution time! On my machine, running the unit tests went from 700+ seconds to under a minute. I am confident that if we continue this trend of rewriting integration tests as unit tests, we can have the same code coverage but have the tests execute in a few seconds.

Please review / FYI @stevengill, @dpogue, @shazron, @purplecabbage, @audreyso, @surajpindoria, @kerrishotts.


filmaj:
EDIT: added a new point number 5 to the above list to point out that I'm moving certain integration tests into unit tests where I believe it makes sense.

src/cordova/platform.js
```diff
@@ -1,769 +0,0 @@
-/**
```
stevengill: to confirm, you ripped out this file into multpile smaller files. I like it. It has been replaced by addHelper.js, check.js, getPlatformDetailsFromDir.js, index.js, list.js, remove.js and save.js

I see tests for addHelper. Do the other modules need new unit tests?

spec-cordova/platform.spec.ios.js
```diff
@@ -1,137 +0,0 @@
-/**
```
filmaj: Yes: https://github.com/apache/cordova-ios/pull/319

src/cordova/platform.js
filmaj: Sure do! Basically, each module should have unit tests. A good rule of thumb is to have one spec file per module file.

To really get a sense of what coverage exists and what is missing, run `npm run cover` and get the code coverage report. Note that as of right now, the unit tests still have _some_ integration tests - so the code coverage report from a unit test standpoint is somewhat inaccurate (integration or end to end tests are very coarse, so they probably cover a lot of code unintentionally). I would recommend a) looking through the spec-cordova and spec-plugman tests for heavy i/o, very broad tests that we can classify as integration tests and moving them into integration-tests and b) replacing them with finer-grained unit tests.

This cordova-ios PR is an example of that general pattern: https://github.com/apache/cordova-ios/pull/319

spec-cordova/platform/addHelper.spec.js
```diff
@@ -0,0 +1,515 @@
+/**
```
filmaj: Sort of. The old tests I `xdescribe`d them so they don't run. Instead, based on the code and general purpose of the `addHelper` module, I wrote up four `it()` statements describing tests we should implement to get decent coverage and exercise the spirit of the module. I also added a big mulitline comment roughly sketching out what `addHelper` does - we should convert that comment into more `it()` statements to lay out the overview of unit tests for this module.

Finally, we should go over the old tests I `xdescribe`d at the bottom and figure out a) if we already have the coverage for what those tests do elsewhere, or if not b) write up unit tests to replace them (possibly in a different module - depends on what the test is testing).

spec-cordova/platform/index.spec.js
```diff
@@ -0,0 +1,181 @@
...
+
+    describe('main module function', function () {
+        describe('error/warning conditions', function () {
+            // TODO: what about other commands? update? save?
+            it('should require at least one platform for add and remove commands', function (done) {
+                // targets = empty array
+                platform('add', []).then(function () {
+                    fail('should not succeed without targets');
+                }, function (err) {
+                    expect(err).toMatch(/You need to qualify.* with one or more platforms/gi);
```
kerrishotts: Out of scope for this commit, but I'd love to see the ability to use `expect(err).toBeInstanceOf(CordovaErrors.ExpectedOneOrMorePlatforms)`. That way should the error message ever change, the test still passes. Plus, then we don't have to worry about ignoring dynamic portions of the error.

I have no idea how big a job that might be (I expect it'd be pretty big!), so I'm just throwing it out there as a "nice-to-have", but I wouldn't want it to hold up any progress on this commit, which is super important.

Thanks for your hard work on this, BTW!

filmaj: That makes a lot of sense - certainly would be much less brittle and thus make the tests more useful.

Perhaps file a JIRA issue for it and assign to me? I'd be happy to put together a PR proofing that approach - if it passes muster, we can then chip away at changing that.

purplecabbage: I would suggest we defined error codes before we define static properties, and start by manually looking things up.  But yeah, I'd consider this out of scope.

src/util/alias.js
```diff
@@ -0,0 +1,27 @@
...
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+var cordova_events = require('cordova-common').events;
+
+module.exports = function aliasMethodToRawWithDeprecationNotice(property, targetObj, component) {
+    targetObj.raw[property] = function() {
+        cordova_events.emit('warn', 'Use of `' + component + '.raw.*` methods is deprecated and `' + component + '.raw` will be removed in a future release. Please migrate to using the top-level `' + component + '.*` methods instead.');
```
purplecabbage: *nit: The habit of using backticks in output sort of muddies the water when we eventually move to using proper ES6 backtick javascript string formatting.

purplecabbage:
This all looks great to me, good work.
I would like to move forward with this asap as it is big, and affects a lot of different areas.
Let's keep our merge conflict exposure low and get this in soon.
I am looking into failing tests on CI.

src/util/alias.js
filmaj: Werd, will remove.

filmaj:
Cheers. The CI is failing due to the end to end tests failing (also JSHint, but that's minor, and I think JSHint in particular is on its way out anyways).

I will keep chipping away at rewriting the integration tests as unit tests wherever I can, however that could take some time, which stands at odds with "merge this in sooner rather than later." 😬

stevengill:
Lets merge and keep chipping away at tests in separate prs.

filmaj:
We'd have to be OK with either removing the e2e tests from standard CI until we finish the chip-away, or be OK with CI failing for a while. Both are tough calls.

stevengill:
I think we should get CI passing before merging. I'm looking into e2e failures now.

filmaj:
Oh so it's merged now? OK, gonna focus more on these tests then.

PR #567, opened 2017-06-21
CB-12361: Added unit tests for prepare.js

stevengill:
### What does this PR do?
Add unit tests for prepare.js in prepare.spec.js. 

npm run cover lists prepare.js at 97.4

### Checklist

### DIFF
--- a/spec-cordova/prepare.spec.js
+++ b/spec-cordova/prepare.spec.js
@@ -29,13 +29,18 @@ var platforms = require('../src/platforms/platforms');
 var HooksRunner = require('../src/hooks/HooksRunner');
 var Q = require('q');
 var PlatformJson = require('cordova-common').PlatformJson;
+var PluginInfoProvider = require('cordova-common').PluginInfoProvider;
 
 var project_dir = '/some/path';
 
 describe('cordova/prepare', function () {
-    var platform_api_prepare_mock;
+    let platform_api_prepare_mock;
+    let platform_api_add_mock;
+    let platform_api_remove_mock;
     beforeEach(function () {
         platform_api_prepare_mock = jasmine.createSpy('prepare').and.returnValue(Q());
+        platform_api_add_mock = jasmine.createSpy('add').and.returnValue(Q());
+        platform_api_remove_mock = jasmine.createSpy('remove').and.returnValue(Q());
         spyOn(platforms, 'getPlatformApi').and.callFake(function (platform, rootDir) {
             return {
                 prepare: platform_api_prepare_mock,
@@ -43,19 +48,22 @@ describe('cordova/prepare', function () {
                     locations: {
                         www: path.join(project_dir, 'platforms', platform, 'www')
                     }
-                })
+                }),
+                removePlugin: platform_api_add_mock,
+                addPlugin: platform_api_remove_mock
             };
         });
         spyOn(PlatformJson, 'load');
+        spyOn(HooksRunner.prototype, 'fire').and.returnValue(Q());
+        spyOn(util, 'isCordova').and.returnValue(true);
     });
 
     describe('main method', function () {
         beforeEach(function () {
             spyOn(cordova_config, 'read').and.returnValue({});
-            spyOn(HooksRunner.prototype, 'fire').and.returnValue(Q());
             spyOn(restore, 'installPlatformsFromConfigXML').and.returnValue(Q());
             spyOn(restore, 'installPluginsFromConfigXML').and.returnValue(Q());
-            spyOn(util, 'isCordova').and.returnValue(true);
+            
             spyOn(prepare, 'preparePlatforms').and.returnValue(Q);
         });
         describe('failure', function () {
@@ -83,22 +91,21 @@ describe('cordova/prepare', function () {
             beforeEach(function () {
                 spyOn(util, 'cdProjectRoot').and.returnValue(project_dir);
                 spyOn(util, 'preProcessOptions').and.callFake(function(options) {
-                    let platforms = options.platforms || []
-                    return {'platforms':platforms}
+                    let platforms = options.platforms || [];
+                    return {'platforms':platforms};
                 });
             });
             it('should fire the before_prepare hook and provide platform and path information as arguments', function(done) {
                 prepare({}).then(function() {
                     expect(HooksRunner.prototype.fire.calls.argsFor(0)[0]).toEqual('before_prepare');
                     done();
-                })
+                });
             });
             it('should invoke restore module\'s installPlatformsFromConfigXML method', function(done) {
                 prepare({}).then(function() {
                     expect(restore.installPlatformsFromConfigXML).toHaveBeenCalled();
                     done();
                 }).fail(function(err){
-                    console.log(err);
                     expect(err).toBeUndefined();
                     done();
                 });
@@ -112,25 +119,31 @@ describe('cordova/prepare', function () {
                     expect(platforms.getPlatformApi.calls.argsFor(1)[1]).toBe('/some/path/platforms/ios');
                     expect(platforms.getPlatformApi.calls.argsFor(2)[0]).toBe('android');
                     expect(platforms.getPlatformApi.calls.argsFor(3)[0]).toBe('ios');
-
                     done();
                 }).fail(function(err){
-                    console.log(err);
                     expect(err).toBeUndefined();
                     done();
                 });
             });
             it('should invoke restore module\'s installPluginsFromConfigXML method', function(done) {
                 prepare({platforms:[]}).then(function() {
-                    //expect(true).toHaveBeenCalled();
+                    expect(restore.installPluginsFromConfigXML).toHaveBeenCalled();
                     done();
                 });
             });
-            it('should invoke preparePlatforms method, providing the appropriate platforms');
+            it('should invoke preparePlatforms method, providing the appropriate platforms', function(done) {
+                prepare({platforms:['android']}).then(function() {
+                    expect(prepare.preparePlatforms).toHaveBeenCalled();
+                    expect(prepare.preparePlatforms.calls.argsFor(0)[0][0]).toBe('android');
+                    expect(prepare.preparePlatforms.calls.argsFor(0)[1]).toBe('/some/path');
+                    done();
+                });
+            });
             it('should fire the after_prepare hook and provide platform and path information as arguments', function(done) {
-                prepare({}).then(function() {
+                prepare({platforms:['android']}).then(function() {
                     expect(HooksRunner.prototype.fire.calls.argsFor(1)[0]).toEqual('after_prepare');
-                    expect(HooksRunner.prototype.fire.calls.argsFor(1)[1]).toEqual({ 'platforms': [], 'paths': [], 'searchpath': undefined });
+                    expect(HooksRunner.prototype.fire.calls.argsFor(1)[1]).toEqual({ 'platforms': [ 'android' ], 'paths': [ '/some/path/platforms/android/www' ], 'searchpath': undefined });
+                    expect(HooksRunner.prototype.fire.calls.count()).toBe(2);
                     done();
                 });
             });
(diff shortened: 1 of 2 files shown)


spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
-        fire,
-        parsers = {},
-        find_plugins,
-        cp,
-        mkdir,
-        load, platformApi, getPlatformApi;
+var project_dir = '/some/path';
 
+describe('cordova/prepare', function () {
+    let platform_api_prepare_mock;
```
filmaj: whoa, `let` !? dragging us into the future eh
iuno if I'm cool with mixing `var` and `let` - I say use either one or the other

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+            spyOn(prepare, 'preparePlatforms').and.returnValue(Q);
+        });
+        describe('failure', function () {
+            it('should invoke util.preProcessOptions as preflight task checker, which, if fails, should trigger promise rejection and only fire the before_prepare hook', function(done) {
+                spyOn(util, 'cdProjectRoot').and.returnValue(project_dir);
+                spyOn(util, 'preProcessOptions').and.throwError();
+                prepare({}).fail(function(err) {
+                    expect(err).toBeDefined();
+                    expect(HooksRunner.prototype.fire.calls.count()).toBe(1);
+                    expect(HooksRunner.prototype.fire.calls.argsFor(0)[0]).toEqual('before_prepare');
```
filmaj: I'd prefer we use the recommended jasmine syntax of `toHaveBeenCalled()` if we just want to check whether a spy was called, or `toHaveBeenCalledWith(arg1, arg2)` if we want to set expectations on invocation arguments. Getting deep into the `calls` sub-properties can get ugly and confusing fast.

e.g. `expect(HooksRunner.prototype.fire).toHaveBeenCalledWith('before_prepare', jasmine.any(Object));`

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+        beforeEach(function () {
+            spyOn(cordova_config, 'read').and.returnValue({});
+            spyOn(restore, 'installPlatformsFromConfigXML').and.returnValue(Q());
+            spyOn(restore, 'installPluginsFromConfigXML').and.returnValue(Q());
+            
+            spyOn(prepare, 'preparePlatforms').and.returnValue(Q);
+        });
+        describe('failure', function () {
+            it('should invoke util.preProcessOptions as preflight task checker, which, if fails, should trigger promise rejection and only fire the before_prepare hook', function(done) {
+                spyOn(util, 'cdProjectRoot').and.returnValue(project_dir);
```
filmaj: Don't we want this spy in the top-level `beforeEach`? Then nested `describe` blocks, or even individual tests, can always customize the spy behaviour with a single `util.cdProjectRoot.and.callFake` or `.and.returnValue`. I think it's cleaner, simpler, and ensures that all nested tests have the function properly mocked out.

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                spyOn(util, 'preProcessOptions').and.throwError();
+                prepare({}).fail(function(err) {
+                    expect(err).toBeDefined();
```
filmaj: I think we want to set expectations on the error message, otherwise, if an unexpected or unrelated error gets thrown, this test may still pass.

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                    expect(HooksRunner.prototype.fire.calls.count()).toBe(1);
+                    expect(HooksRunner.prototype.fire.calls.argsFor(0)[0]).toEqual('before_prepare');
+                    done();
+                });
+            });
+            it('should invoke util.cdProjectRoot as a preflight task checker, which, if fails, should trigger a promise rejection and fire no hooks', function(done) {
+                spyOn(util, 'cdProjectRoot').and.throwError();
+                prepare({}).fail(function(err) {
+                    expect(err).toBeDefined();
```
filmaj: Ditto - please set expectations on error contents.

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                    expect(HooksRunner.prototype.fire.calls.any()).toBe(false);
```
filmaj: Ditto - please use `expect(HooksRunner.prototype.fire).not.toHaveBeenCalled();`

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+            beforeEach(function () {
+                spyOn(util, 'cdProjectRoot').and.returnValue(project_dir);
+                spyOn(util, 'preProcessOptions').and.callFake(function(options) {
+                    let platforms = options.platforms || [];
+                    return {'platforms':platforms};
+                });
+            });
+            it('should fire the before_prepare hook and provide platform and path information as arguments', function(done) {
+                prepare({}).then(function() {
+                    expect(HooksRunner.prototype.fire.calls.argsFor(0)[0]).toEqual('before_prepare');
```
filmaj: Please use `toHaveBeenCalledWith()` 🌹

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                    done();
+                });
+            });
+            it('should invoke restore module\'s installPlatformsFromConfigXML method', function(done) {
+                prepare({}).then(function() {
+                    expect(restore.installPlatformsFromConfigXML).toHaveBeenCalled();
+                    done();
```
filmaj: Instead of popping a `done()` into both the success and error handlers, factor it out into a final `.done(done);` promise handler.

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+            });
+            it('should retrieve PlatformApi instances for each platform provided', function(done) {
+                prepare({'platforms':['android', 'ios']}).then(function() {
+                    expect(platforms.getPlatformApi.calls.count()).toBe(4);
+                    expect(platforms.getPlatformApi.calls.argsFor(0)[0]).toBe('android');
+                    expect(platforms.getPlatformApi.calls.argsFor(0)[1]).toBe('/some/path/platforms/android');
+                    expect(platforms.getPlatformApi.calls.argsFor(1)[0]).toBe('ios');
+                    expect(platforms.getPlatformApi.calls.argsFor(1)[1]).toBe('/some/path/platforms/ios');
+                    expect(platforms.getPlatformApi.calls.argsFor(2)[0]).toBe('android');
+                    expect(platforms.getPlatformApi.calls.argsFor(3)[0]).toBe('ios');
```
filmaj: This is where using `toHaveBeenCalledWith()` will really shine, as otherwise, if even the order of calls changes, these expectations will fail. Please use that API :) 🌷

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                    done();
+                }).fail(function(err){
+                    expect(err).toBeUndefined();
+                    done();
```
filmaj: Ditto, factor the dones into a `.done(done);` please 💝

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                });
+            });
+            it('should invoke restore module\'s installPluginsFromConfigXML method', function(done) {
+                prepare({platforms:[]}).then(function() {
+                    expect(restore.installPluginsFromConfigXML).toHaveBeenCalled();
+                    done();
```
filmaj: For these kind of tests, where we want to set expectations on something happening in _either_ the success or failure handler, I like to structure the test like so:

```
prepare().then(function () {
    expect(something).toBe(true);
}).fail(function (e) {
    fail('unexpected failure handler invoked');
    console.error(e);
}).done(done);
```

The reasons I do it this way are:
1. If the handler that I am not expecting gets invoked, the test fails immediately. With your implementation above, we'd have to wait for the timeout to kick in (which is by default 5 seconds IIRC). That can add up if a lot of tests are exhibiting that behaviour.
2. There's always one spot to call `done`, using the final promise resolution handler (either `done` or `fin` - both should be identical).
3. If the wrong handler gets invoked, by logging out the error, the person that sees the error can immediately start debugging, as `console.log` or `console.error` can handle Error object stack traces and print them out too. Maybe there's a way to combine the `fail` + `console.error` calls into one. I know if you do `expect(err).toBeUndefined()`, you will only get the error message and not the stack trace, which makes debugging harder. Perhaps `fail('unexpected fail handler invoked with error: ' + err)` would work too?

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                });
+            });
+            it('should invoke preparePlatforms method, providing the appropriate platforms', function(done) {
+                prepare({platforms:['android']}).then(function() {
+                    expect(prepare.preparePlatforms).toHaveBeenCalled();
+                    expect(prepare.preparePlatforms.calls.argsFor(0)[0][0]).toBe('android');
+                    expect(prepare.preparePlatforms.calls.argsFor(0)[1]).toBe('/some/path');
+                    done();
```
filmaj: Ditto, factor done into `.done(done);` and explicitly fail in the fail handler please :) 🎁

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                });
+            });
+            it('should fire the after_prepare hook and provide platform and path information as arguments', function(done) {
+                prepare({platforms:['android']}).then(function() {
+                    expect(HooksRunner.prototype.fire.calls.argsFor(1)[0]).toEqual('after_prepare');
+                    expect(HooksRunner.prototype.fire.calls.argsFor(1)[1]).toEqual({ 'platforms': [ 'android' ], 'paths': [ '/some/path/platforms/android/www' ], 'searchpath': undefined });
+                    expect(HooksRunner.prototype.fire.calls.count()).toBe(2);
+                    done();
```
filmaj: Ditto, done into .done(done); and explicitly fail in the fail handler please :) 💃

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                prepare({platforms:[]}).then(function() {
+                    expect(restore.installPluginsFromConfigXML).toHaveBeenCalled();
+                    done();
+                });
+            });
+            it('should invoke preparePlatforms method, providing the appropriate platforms', function(done) {
+                prepare({platforms:['android']}).then(function() {
+                    expect(prepare.preparePlatforms).toHaveBeenCalled();
+                    expect(prepare.preparePlatforms.calls.argsFor(0)[0][0]).toBe('android');
+                    expect(prepare.preparePlatforms.calls.argsFor(0)[1]).toBe('/some/path');
```
filmaj: Please use `toHaveBeenCalledWith()` here 🌈

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                    done();
+                });
+            });
+            it('should fire the after_prepare hook and provide platform and path information as arguments', function(done) {
+                prepare({platforms:['android']}).then(function() {
+                    expect(HooksRunner.prototype.fire.calls.argsFor(1)[0]).toEqual('after_prepare');
+                    expect(HooksRunner.prototype.fire.calls.argsFor(1)[1]).toEqual({ 'platforms': [ 'android' ], 'paths': [ '/some/path/platforms/android/www' ], 'searchpath': undefined });
+                    expect(HooksRunner.prototype.fire.calls.count()).toBe(2);
```
filmaj: `toHaveBeenCalledWith()` here please 🌞

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+            platform_munger_revert_mock();
+        });
+        it('should call restoreMissingPluginsForPlatform', function(done) {
+            prepare.preparePlatforms(['android'],project_dir, {}).then(function() {
+                expect(prepare.restoreMissingPluginsForPlatform).toHaveBeenCalled();
+                done();
+            }).fail(function(err) {
                 expect(err).toBeUndefined();
-            }).fin(done);
+                done();
```
filmaj: Ditto, factor dones into `.done(done);` please 🏄

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
-                supported_platforms.forEach(function(p) {
-                    expect(parsers[p]).toHaveBeenCalled();
-                    expect(getPlatformApi).toHaveBeenCalledWith(p);
-                });
-                expect(platformApi.run).toHaveBeenCalled();
-            }, function(err) {
+        it('should retrieve the platform API via getPlatformApi per platform provided, and invoke the prepare method from that API', function(done) {
+            prepare.preparePlatforms(['android'],project_dir, {}).then(function() {
+                expect(platforms.getPlatformApi.calls.count()).toBe(1);
+                expect(platforms.getPlatformApi.calls.argsFor(0)[0]).toBe('android');
```
filmaj: `toHaveBeenCalledWith()` please 🏒

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                done();
+            }).fail(function(err) {
+                expect(err).toBeUndefined();
+                done();
```
filmaj: Factor dones into `.done(done);` please ⚽️

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
         });
-        it('Test 004 : should get PlatformApi instance for each platform and invoke its\' run method', function(done) {
-            prepare().then(function() {
-                supported_platforms.forEach(function(p) {
-                    expect(parsers[p]).toHaveBeenCalled();
-                    expect(getPlatformApi).toHaveBeenCalledWith(p);
-                });
-                expect(platformApi.run).toHaveBeenCalled();
-            }, function(err) {
+        it('should retrieve the platform API via getPlatformApi per platform provided, and invoke the prepare method from that API', function(done) {
```
filmaj: I think in this test we also need to assert that the platform API object's `prepare` method gets invoked.

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+            });
+        });
+        it('should fire a pre_package hook for the windows', function(done) {
+            prepare.preparePlatforms(['windows'],project_dir, {}).then(function() {
+                expect(HooksRunner.prototype.fire.calls.count()).toBe(1);
+                expect(HooksRunner.prototype.fire.calls.argsFor(0)[0]).toBe('pre_package');
+                done();
+            }).fail(function(err) {
+                expect(err).toBeUndefined();
+                done();
```
filmaj: `.done(done);` please 🔸

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+            });
+        });
+        it('should fire a pre_package hook for the windows', function(done) {
+            prepare.preparePlatforms(['windows'],project_dir, {}).then(function() {
+                expect(HooksRunner.prototype.fire.calls.count()).toBe(1);
+                expect(HooksRunner.prototype.fire.calls.argsFor(0)[0]).toBe('pre_package');
```
filmaj: `toHaveBeenCalledWith()` please 🥇

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+        xit('should invoke browserify if the browserify option is provided');
+        it('should handle config changes by invoking add_config_changes and save_all', function(done) {
+            prepare.preparePlatforms(['android'],project_dir, {}).then(function() {
+                expect(platform_munger_mock.prototype.add_config_changes).toHaveBeenCalled();
+                expect(platform_munger_save_mock).toHaveBeenCalled();
+                done();
+            }).fail(function(err) {
                 expect(err).toBeUndefined();
-            }).fin(done);
+                done();
```
filmaj: `done(done);` please 🐤

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
             });
-            it('Test 006 : should fire after hooks through the hooker module, and pass in platforms and paths as data object', function(done) {
-                prepare('android').then(function() {
-                    expect(fire.calls.argsFor(1)).toEqual([ 'after_prepare',{ platforms: [ 'android' ],verbose: false,options: {},paths: [ path.join( '/','some','path', 'platforms', 'android', 'www' ) ],searchpath: undefined } ]);
-                }, function(err) {
-                    expect(err).toBeUndefined('Exception while running `prepare android`:\n' + err.stack);
-                }).fin(done);
+
+            prepare.restoreMissingPluginsForPlatform('android', project_dir, {}).then(function() {
+                expect(platforms.getPlatformApi).not.toHaveBeenCalled();
```
filmaj: If I understand correctly, we are asserting that `getPlatformApi` does not get called as a way of ensuring that `restoreMissingPluginsForPlatform` "short-circuits" out of the heavier logic, in the easy case. If that is true, can we change the test name to reflect that so that it's clear for people looking at these in the future?

Perhaps "should resolve quickly and not invoke getPlatformAPI in the easy case of there being no different between old and new platform.json" ?

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+                done();
+            }).fail(function(err) {
+                expect(err).toBeUndefined();
+                done();
```
filmaj: `.done(done);` please

spec-cordova/prepare.spec.js
```diff
@@ -17,175 +17,265 @@
...
+            });
+            prepare.restoreMissingPluginsForPlatform('android', project_dir, {}).then(function() {
+                expect(platforms.getPlatformApi).toHaveBeenCalled();
+                expect(platform_api_add_mock).toHaveBeenCalled();
+                expect(platform_api_remove_mock).toHaveBeenCalled();
+                expect(PluginInfoProvider.prototype.get).toHaveBeenCalled();
+                done();
+            }).fail(function(err) {
+                expect(err).toBeUndefined();
+                done();
```
filmaj: `done(done);` please

spec-cordova/prepare.spec.js
stevengill: Ahh. I didn't realize i could reassign values. I moved some spies down into its when I needed them to have different values to other its in the same describe block. Seeing that I can just customize values with `util.cdProjectRoot.and.callFake` or `.and.returnValue` makes sense and is cleaner.

spec-cordova/prepare.spec.js
stevengill: What do you recommend I do when I'm force throwing the error and it has no message? See line 72. I throw the error there. So err doesn't have a set message. I could fake add one in during the throwError but that is useless.

spec-cordova/prepare.spec.js
filmaj: I think it's a cleaner pattern, plus, it forces you to mock out everything right from the beginning, enforcing a truer unit testing style. One more benefit is that within the scope of a single `it`, you can see just what mocks are being tweaked to trigger the behaviour the test is looking to reproduce. So I think it makes clearer the situation under which the test executes.

spec-cordova/prepare.spec.js
stevengill: kk. I wasn't sure what style we wanted to go with. I've also seen `.fin(done)`

spec-cordova/prepare.spec.js
filmaj: I actually didn't know you could use `throwError` that way! Looking at the jasmine docs, it looks like you can pass a string argument to `throwError` to customize the error message - maybe try that? If not, in the past I have used `and.callFake(function () { throw new Error('boom!'); });`

spec-cordova/prepare.spec.js
stevengill: Sounds good. Should I add a `.fail( function (e) { ...` to all of the tests? I added it selectively throughout the tests. 

Also, didn't know about `fail('unexpected failure handler invoked');`. Is that a built in jasmine function? Doesn't seem documented on their site.

spec-cordova/prepare.spec.js
stevengill: Gotta start dragging us that way sometime 😄! I know @purplecabbage cabbage has said he doesn't want us to start using let yet, but it is pretty safe to use in our tests to start with. 

I'll update the top vars to lets.

spec-cordova/prepare.spec.js
filmaj: I think all of the tests, yeah, or at least the tests where there is an implicit requirement on which handler is to be invoked (and which not).

Here's documentation on `fail`: https://jasmine.github.io/api/2.6/global.html#fail

stevengill: Thanks! I was looking at 2.0 docs 😅

spec-cordova/prepare.spec.js
stevengill: Yeah i'm pretty sure I can do `throwError('error message')`. It felt a little useless to me at the time but I see now that if something unexpected goes wrong, want the test to fail

filmaj: I try to have fun with it, use dumb jokes for strings, let my imagination run wild a bit ;P

spec-cordova/prepare.spec.js
purplecabbage: I am cool with `let|const|var`, my reservation is around `() =>` syntax as I want to be sure everyone is using it correctly before it gets dumped all over the place.

I kinda think this pr is not the place to start the conversion effort.

filmaj: Agreed.

stevengill: I've converted them back to `var` for now and will bring up es6 on the list

stevengill:
Sent an update with changes based on feedback

spec-cordova/prepare.spec.js
```diff
@@ -63,103 +63,103 @@ describe('cordova/prepare', function () {
...
-                    expect(platforms.getPlatformApi.calls.argsFor(2)[0]).toBe('android');
-                    expect(platforms.getPlatformApi.calls.argsFor(3)[0]).toBe('ios');
-                    done();
-                }).fail(function(err){
-                    expect(err).toBeUndefined();
-                    done();
-                });
+            it('should retrieve PlatformApi instances for each platform provided', function (done) {
+                prepare({'platforms':['android', 'ios']}).then(function () {
+                    expect(platforms.getPlatformApi).toHaveBeenCalledTimes(4);
```
filmaj: O I didn't know this was available!

filmaj:
@stevengill rebase away and give it all another go.

PR #597, opened 2017-10-16
CB-13451 (all platforms) "pkg not defined" exception when running plugman with createpackagejson command-line

OminousWater:
A global js-lint pass on the cordova-lib codebase on the 30/8/2017 caused references to the global 'package' variable in init-default.js to be renamed to 'pkg’, presumably because the cordova-lib source doesn’t declare a global variable called ‘package’ but _does_ declare a global variable called ‘pkg’ (in src/cordova/info.js). However in this case, the ‘package’ variable refers to the one declared in the plugman source, specifically main.js, so it should have stayed as 'package’. To test the fix, run:

	plugman createpackagejson .

Without the fix, this will trigger the following exception: 'pkg is not defined'. Having patched in the fix, running the command line should work as expected, prompting the user with questions and then spitting out a package.json file.

### Platforms affected
All

### What does this PR do?
Fixes an bug where an exception would be thrown when running 'plugman createpackagejson .'

The command in question ('plugman createpackagejson .') has been tested with and without the fix to confirm that the fix fixes the problem.

### Checklist

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,4 +1,1 @@
-spec/cordova/fixtures/**
-spec/plugman/projects/**
-spec/plugman/plugins/**
-spec/cordova/temp/**
+**/init-defaults.js
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,6 @@
-**/init-defaults.js
+spec/cordova/fixtures/*
+spec/plugman/projects/*
+spec/plugman/plugins/*
+spec/cordova/temp/*
+src/plugman/*
+
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,6 +1,6 @@
+src/plugman/init-defaults.js
 spec/cordova/fixtures/*
 spec/plugman/projects/*
 spec/plugman/plugins/*
 spec/cordova/temp/*
-src/plugman/*
 
--- a/src/plugman/init-defaults.js
+++ b/src/plugman/init-defaults.js
@@ -19,7 +19,7 @@
 
 /* global dirname */
 /* global config */
-/* global pkg */
+/* global package */
 /* global basename */
 /* global yes */
 /* global prompt */
@@ -55,17 +55,18 @@ function readDeps (test) {
     };
 }
 
-var name = pkg.name || defaults.id || basename;
+/* eslint-disable */
+var name = package.name || defaults.id || basename;
 exports.name = yes ? name : prompt('name', name);
 
-var version = pkg.version ||
+var version = package.version ||
               defaults.version ||
               config.get('init.version') ||
               config.get('init-version') ||
               '1.0.0';
 exports.version = yes ? version : prompt('version', version);
 
-if (!pkg.description) {
+if (!package.description) {
     if (defaults.description) {
         exports.description = defaults.description;
     } else {
@@ -73,7 +74,7 @@ if (!pkg.description) {
     }
 }
 
-if (!pkg.cordova) {
+if (!package.cordova) {
     exports.cordova = {};
     if (defaults.id) {
         exports.cordova.id = defaults.id;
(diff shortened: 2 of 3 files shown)


OminousWater:
I see the CI has failed, because 'package' is a reserved keyword (this better explains the changes made as part of the js-lint pass a few months back). Perhaps the better fix, then, is to rename the variable in the plugman source; because that, too, would presumably fail a js-lint pass.

zaq962:
how can i resove it, I have the same problem

OminousWater:
@zaq962 if you apply the changes in my pull request (it's just one file) to your local checkout, it should work fine. The only reason it doesn't pass the continuous integration test is that 'package' is a reserved keyword, which eslint doesn't like. I've been experimenting with fixing the problem on the plugman side, by renaming the 'package' keyword there, which would be a better fix arguably. If I get that working then I'll re-open the pull request against the plugman source.

OminousWater:
@zaq962 I think I have the "correct" fix now, am about to open another pull request.

OminousWater:
OK I understand the problem better now. init-package-json (a separate npm module) initialises the global variable 'package' in init-package-json.js. So ideally init-package-json would use a different variable name, but I guess the maintainers of that package might not have as strict a requirement as the cordova-lib maintainers. I've opened a separate pull request with a compromise solution here: https://github.com/npm/init-package-json/pull/74, hopefully they will accept (or have a better solution).

stevengill:
I think turning it back to `package` in plugman is the way to go. We can add an eslint exception for it I hope

OminousWater:
@stevengill I've been trying that as alternative solution, however it seems eslint doesn't support exceptions for parser errors. See their response to the issue I lodged here after a fair bit of fruitless experimentation with eslint exclude rules: https://github.com/eslint/eslint/issues/9483

stevengill:
Thanks for all your work on this @OminousWater!

OminousWater:
@stevengill I have a working fix locally now (on advice from the eslint devs I've added a top-level .eslintignore file with a pattern that filters out any init-defaults.js files). The AppVeyor continuous integration step is still failing, though for a seemingly unrelated reason (to do with the 'qs' package?). I will try to repro locally in case it's related to my change, but I don't _think_ it is (famous last words).

OminousWater:
I have committed my fix now, and think the pull-request is good to go. eslint is still reporting errors, but I've repro'd them locally and can confirm they're unrelated to my pull request.

.eslintignore
```diff
@@ -1,4 +1 @@
-spec/cordova/fixtures/**
-spec/plugman/projects/**
-spec/plugman/plugins/**
-spec/cordova/temp/**
+**/init-defaults.js
```
stevengill: Why did you remove the other items from `.eslintignore`?

src/plugman/init-defaults.js
```diff
@@ -55,25 +55,26 @@ function readDeps (test) {
     };
 }
 
-var name = pkg.name || defaults.id || basename;
+/* eslint-disable */
```
stevengill: Do you need this `/* eslint-disable */`

src/plugman/init-defaults.js
```diff
@@ -149,10 +149,11 @@ if (!pkg.author) {
...
         : prompt('author');
 }
-/* eslint-enable indent */
-var license = pkg.license ||
+var license = package.license ||
               defaults.license ||
               config.get('init.license') ||
               config.get('init-license') ||
               'ISC';
+/* eslint-enable */
```
stevengill: is this needed? `/* eslint-enable */`

OminousWater:
Thanks for picking up on those problems -- how embarrassing! The lines removed from .eslintignore was a complete mistake on my part, thank you for spotting that. I've re-instated those lines, and revised the glob patterns at the same time, as the ones that were there didn't seem to be doing what they were intended to, with the latest version of eslint. I also removed the now-redundant eslint directives from init-default.js. Do you know how I can trigger another Continous Integration run on the server to check everything's good?

stevengill:
No worries @OminousWater! That is what reviews are for :)

CI gets trigged with new commits or updated commits. Looks like your PR triggered it. I'll wait and see how they go

stevengill:
merged! Thanks again @OminousWater!

jh3141:
It seems quite a few people are experiencing the problem that this PR solved (cf https://stackoverflow.com/questions/47102566/pkg-is-not-defined-when-using-plugman) - I don't know what the process is, but is there any way of expediting a release with this change included?

PR #602, opened 2017-10-24
CB-12774 : Don't munge scoped plugin IDs anymore. Comment requested.

akdor1154:
In order to get plugins that are under an npm scope (e.g. `@akdor1154/some-plugin`) to install, a previous change from took the approach to consider such a plugin to be called (have the id) `some-plugin` instead of `@akdor1154/some-plugin`. This allowed scoped plugins to install and preserved the assumption that plugins will always be installed in `plugins_dir/[plugin_id]`.

However, it required special parsing logic around `npm` package IDs, and it broke the assumption that the `name` in an npm plugin's `package.json` would correspond to Cordova's idea of a plugin ID. IMO this approach is the source of further complexity which is not required, and is leading to weird bugs and special cases with scoped plugins. (see the linked issue, but there is stuff as basic as "`npm install` no longer works after installing a scoped plugin")

This PR changes approach - such plugins are now considered to have the id `@akdor1154/some-plugin`, in agreement with how npm treats such packages. This allows almost all special cases for scoped packages to be removed (yay). The key difference in behaviour as a result of this, though, is that while plugins are still installed in `plugins_dir/[plugin_id]`, `[plugin_id]` may no longer be a single directory, leading to plugin directories that look like
```
plugins
|- @akdor1154
|   |- some-plugin
|- cordova-some-other-plugin
```

Most of the logic changes in this PR are based around making this change work.

It's largely done how I want it, but I guess maintainers probably have strong opinions over whether this is the right way to go or not. Because of this I've left some commits in marked as TEMP that are for my own workflow (i.e. I am expecting to have to make further changes). Please keep in mind I'll remove these. The only one you should be mindful of is the monkey patch to `cordova-common`; this would need to be raised in a separate PR to cordova-common if this change was approved in principle.

The work was done on a Mac but I'll check it properly on my windows laptop when I get home (path sep issues are expected).

### Platforms affected
All

New end-to-end integration tests including a proper scoped plugin fixture
Unit tests on scoped plugins
Unit tests for plugman metadata

### Checklist

cc: @riknoll

### DIFF
--- a/src/cordova/util.js
+++ b/src/cordova/util.js
@@ -255,11 +255,20 @@ function findPlugins (pluginDir) {
     var plugins = [];
 
     if (fs.existsSync(pluginDir)) {
-        plugins = fs.readdirSync(pluginDir).filter(function (fileName) {
-            var pluginPath = path.join(pluginDir, fileName);
-            var isPlugin = isDirectory(pluginPath) || isSymbolicLink(pluginPath);
-            return fileName !== '.svn' && fileName !== 'CVS' && isPlugin;
-        });
+        plugins = fs.readdirSync(pluginDir)
+            .reduce(function (plugins, pluginOrScope) {
+                if (pluginOrScope[0] === '@') {
+                    plugins.push(...fs.readdirSync(path.join(pluginDir, pluginOrScope)).map(s => path.join(pluginOrScope, s)));
+                } else {
+                    plugins.push(pluginOrScope);
+                }
+                return plugins;
+            }, [])
+            .filter(function (fileName) {
+                var pluginPath = path.join(pluginDir, fileName);
+                var isPlugin = isDirectory(pluginPath) || isSymbolicLink(pluginPath);
+                return fileName !== '.svn' && fileName !== 'CVS' && isPlugin;
+            });
     }
 
     return plugins;
--- a/src/cordova/util.js
+++ b/src/cordova/util.js
@@ -258,7 +258,9 @@ function findPlugins (pluginDir) {
         plugins = fs.readdirSync(pluginDir)
             .reduce(function (plugins, pluginOrScope) {
                 if (pluginOrScope[0] === '@') {
-                    plugins.push(...fs.readdirSync(path.join(pluginDir, pluginOrScope)).map(s => path.join(pluginOrScope, s)));
+                    var scopedPlugins = fs.readdirSync(path.join(pluginDir, pluginOrScope))
+                        .map(scopedPluginDir => path.join(pluginOrScope, scopedPluginDir));
+                    plugins.push.apply(plugins, scopedPlugins);
                 } else {
                     plugins.push(pluginOrScope);
                 }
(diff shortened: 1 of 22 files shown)


src/cordova/util.js
```diff
@@ -255,11 +255,20 @@ function findPlugins (pluginDir) {
...
     if (fs.existsSync(pluginDir)) {
-        plugins = fs.readdirSync(pluginDir).filter(function (fileName) {
-            var pluginPath = path.join(pluginDir, fileName);
-            var isPlugin = isDirectory(pluginPath) || isSymbolicLink(pluginPath);
-            return fileName !== '.svn' && fileName !== 'CVS' && isPlugin;
-        });
+        plugins = fs.readdirSync(pluginDir)
+            .reduce(function (plugins, pluginOrScope) {
+                if (pluginOrScope[0] === '@') {
+                    plugins.push(...fs.readdirSync(path.join(pluginDir, pluginOrScope)).map(s => path.join(pluginOrScope, s)));
```
audreyso: Hi! This line will need to be changed as it will not work with Node 4, which is still currently supported. http://node.green/#ES2015-syntax-spread-------operator-with-generic-iterables--in-calls

akdor1154: yup just started looking at the failed test :) While you're here and kind of related, I noticed most of the Cordova codebase avoids `() => arrow functions`. Is this just historical? I have used full `function () { }` when I've paid attention just to match code style, but is this necessary?

audreyso: Thanks for your question! We are sticking to the current style for now. We need to have a discussion on our mailing list about starting to use es2015 features/ which ones to use and get some consensus.

akdor1154:
ahh Windows..

tobilarscheid:
This actually sounds like a good thing to merge - what is blocking us? :-)

cthos:
I'd very much like to be able to do this, given it's a good way to have private cordova plugins (either through a mirror or npm's privately scoped packages offering) - and adding plugins via private git urls is also not particularly functional.

janpio:
@tobilarscheid This PR has a failing test on Windows:

```
Failures:
1) plugin end-to-end Test 013 : should be able to add and remove scoped npm packages without screwing up everything
  Message:
    Expected Plugin "@testscope/cordova-lib-test-plugin-scoped" is not present in the project. See `cordova plugin list`. to be undefined.
  Stack:
    Error: Expected Plugin "@testscope/cordova-lib-test-plugin-scoped" is not present in the project. See `cordova plugin list`. to be undefined.
        at C:\projects\cordova-lib\integration-tests\plugin.spec.js:329:29
        at _rejected (C:\projects\cordova-lib\node_modules\q\q.js:797:24)
        at C:\projects\cordova-lib\node_modules\q\q.js:823:30
        at Promise.when (C:\projects\cordova-lib\node_modules\q\q.js:1035:31)
```

Also unfortunately the branch is now also conflicted and has to be rebased. Could you take care of that @akdor1154 please? Thanks.

oliversalzburg:
I rebased this locally and did not see the same failure.

raphinesse:
@oliversalzburg Great to hear! OP hasn't reacted to previous mentions, so I'd suggest you open a PR with the rebased changes that references this one. Then we can have another look at it.

raphinesse:
Closing in favor of rebased #680.

PR #608, opened 2017-11-20
CB:13055- remove --nofetch option and all old fetching code

audreyso:
Not ready for review yet. Thanks!

### What does this PR do?
remove --nofetch option and all old fetching code

### Checklist


integration-tests/pkgJson-restore.spec.js
```diff
@@ -516,7 +514,7 @@ describe('update pkg.json to include platforms in config.xml', function () {
         expect(pkgJson.dependencies[browserPlatform]).toBeUndefined();
         expect(pkgJson.dependencies[androidPlatform]).toBeUndefined();
         emptyPlatformList().then(function () {
-            return prepare();
+            return prepare({'fetch': true});
```
audreyso: @stevengill In this file, I've added the fetch flag to most/all of the tests to fix them. This did not work the same way for test 005 and 007. You do need to add in **fetch** here (just like all of the other tests) or else you will get  this failure -->`Expected No platforms added to this project. Please use `cordova platform add <platform>`. to be undefined.`

However, adding the fetch flag here, will lead to other 005 and 007 tests failures as shome below. (FYI they are passing independently.) -->

```cp: no such file or directory: /private/var/folders/13/dkz30nws1dv3dzlh6dvthwzm0000gn/T/e2e-test/platform_test_pkgjson/project/node_modules/cordova-android/bin/templates/project/wrapper.gradle```

File: package.json [hunk 2]
```diff
@@ -28,7 +28,6 @@
     "glob": "7.1.1",
     "init-package-json": "^1.2.0",
     "nopt": "4.0.1",
-    "npm": "^2.10.x",
```
audreyso: @stevengill since we are removing npm here, what is the best way to update the scripts in package.json?

integration-tests/platform.spec.js
```diff
@@ -128,12 +128,11 @@ describe('platform end-to-end', function () {
                 expect(err).toBeUndefined();
             }).fin(done);
     });
-
-    it('Test 002 : should install plugins correctly while adding platform', function (done) {
-
-        cordova.plugin('add', path.join(pluginsDir, 'test'))
+    // was using old fetch code/gitclone... test might not be valid anymore.
+    xit('Test 002 : should install plugins correctly while adding platform', function (done) {
```
stevengill: Did you mean to have this `xit`

audreyso: i think it was was using old fetch code/gitclone... wasn't sure if the test was still valid. I took out the "x" so we can see if it passes/fails.

PR #611, opened 2018-04-27
Dependency cleanup

dpogue:
### What does this PR do?
Remove several dependencies that are no longer used directly by this module.
Update several other dependencies to their latest versions. No more deprecation warnings on install!

All tests passed locally


stevengill:
Tested locally, works!

PR #613, opened 2018-05-18
CB-14140 Switch to using fs-extra in favour of shelljs

dpogue:
Nowhere near ready to review or merge, but opening so I can see how the tests are faring on Travis/Appveyor...

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

PR #619, opened 2018-06-22
Let Jasmine handle async test results

raphinesse:
There was a lot of sub-optimal error handling in the tests.

Instead of treating every promise like the unique little sparkly star that it is, we just return them from the test functions and let Jasmine handle them.

This also fixes tons of negative tests that often looked something like this:

```js
sparkleSparkle('NOT!').then(function () {
    expect('Roses').toBe('Blue');
}).catch(function (err) {
    expect('' + err).toMatch('not so sparkly after all');
});
```

These were normalized to something like this (spot all the differences! :wink:):
```js
sparkleSparkle('NOT!').then(function () {
    fail('Expected promise to be rejected');
}, function (err) {
    expect(err).toEqual(jasmine.any(Error));
    expect(err.message).toMatch('not so sparkly after all');
});
```

This commit has multiple benefits
- Removes 700 LOC w/out removing any testing
- Improves output when async tests fail
- Should make async tests less error prone
- Reduces Q-API usage to prepare for removal of Q (CB-14159)

I really don't expect anyone to review this ~2k LOC change set. Since it only affects the tests and they still pass I would merge this without a review if no one objects.

### DIFF
--- a/integration-tests/pkgJson-restore.spec.js
+++ b/integration-tests/pkgJson-restore.spec.js
@@ -31,8 +31,6 @@ var TIMEOUT = 240 * 1000;
 *   should initially be in pkg.json and/or config.xml.
 */
 
-/* eslint "promise/catch-or-return": "error" */
-
 // Use basePkgJson
 describe('tests platform/spec restore with --save', function () {
     var tmpDir = helpers.tmpDir('platform_test_pkgjson2');
--- a/integration-tests/pkgJson-restore.spec.js
+++ b/integration-tests/pkgJson-restore.spec.js
@@ -31,6 +31,8 @@ var TIMEOUT = 240 * 1000;
 *   should initially be in pkg.json and/or config.xml.
 */
 
+/* eslint "promise/catch-or-return": "error" */
+
 // Use basePkgJson
 describe('tests platform/spec restore with --save', function () {
     var tmpDir = helpers.tmpDir('platform_test_pkgjson2');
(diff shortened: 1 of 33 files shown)


integration-tests/pkgJson-restore.spec.js
```diff
@@ -31,6 +31,8 @@ var TIMEOUT = 240 * 1000;
 *   should initially be in pkg.json and/or config.xml.
 */
 
+/* eslint "promise/catch-or-return": "error" */
```
brodybits: Why put `eslint "promise/catch-or-return": "error"` here? Shouldn't it be in a .eslint config file (doesn't have to be top-level)?

raphinesse: That's an oversight. Thanks, will remove!

brodybits:
+1 from a quick scan 🎉

raphinesse:
All right, gonna merge this...

raphinesse:
@brodybits I haven't actively looked into reasons for CI failures. But if anything, this should help make the tests more robust, yes.

Having skimmed through almost all of the tests now, I have to say there's some horrible code in there and setup/teardown of test dirs is often pretty crude too. So I rather expect that there are still tests that are note stable. #613 will hopefully further improve the situation.

raphinesse:
> _@brodybits wrote_: Nice work [...] wonder if we should apply similar updates to some of the other Cordova repos?

Definitely. I've done that for one or two already (create/fetch I guess). Should be easier for everything other than lib. Maybe an issue with sub-tasks would be in order.

However, I mainly started this to reduce Q-usage. Basically we need to remove usage of Q instance methods in depending packages to be able to return native Promises in the dependencies. Thus, `plugman` and `cordova-cli` would be next on my priority list.

brodybits:
> Should be easier for everything other than lib.

That would definitely be good!

> Maybe an issue with sub-tasks would be in order.

I am starting to agree with this. JIRA issues are really good for logging & tracking changes, but if I label PR discussions for multiple components under the same JIRA issue it becomes cluttered pretty quickly.

> reduce Q-usage.

💯 I was already thinking to raise an issue to reduce or hopefully eliminate Q usage, guess I should leave it in your hands for now.

> _@raphinesse wrote in <https://github.com/apache/cordova-lib/pull/613#issuecomment-399471248>:_ I'm still working on other cordova stuff [...]

Yup, understood. I guess we are all in the same boat:)

raphinesse:
> I was already thinking to raise an issue to reduce or hopefully eliminate Q usage

https://issues.apache.org/jira/browse/CB-14158

PR #621, opened 2018-06-27
CB-14033 Support symbolic dir links on Windows

raphinesse:
### Platforms affected
Windows

### What does this PR do?
This uses junctions whenever creating directory symlinks on Windows since that does not require any special privileges.

Tests on AppVeyor CI

### Checklist

### DIFF
--- a/integration-tests/plugman_fetch.spec.js
+++ b/integration-tests/plugman_fetch.spec.js
@@ -94,7 +94,7 @@ describe('fetch', function () {
         });
         it('Test 003 : should create a symlink if used with `link` param', function () {
             return fetch(test_plugin, temp, { link: true }).then(function () {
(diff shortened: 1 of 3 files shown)


brodybits:
I think it would be idea to test with Cordova CLI, would be ideal to test with plugin which *may* be affected by this change. I can take a better look tomorrow.

raphinesse:
@brodybits Since there is a test that specifically targets the changed code, I'm quite confident it's OK.
However, some real-world testing is always good. Thanks for taking care of it :+1:

raphinesse:
@brodybits Did you test this on Windows yet? Or did I misunderstand your comment? :thinking:

brodybits:
My apologies, no time for me to test it this week, probably not much time next week either. Considering that you tested it in spec and got an approval from someone else I don't think you should let me block it.

I do think it would be good for you to try a version of cordova-cli that does use a version of cordova-lib with this change and verify that it does actually solve the problem at some point.

PR #623, opened 2018-07-03
Clean up messy HooksRunner tests

raphinesse:
These tests tended to cause spurious fails on CI. After having done a complete cleanup, I don't find it surprising at all. I hope they are stable now.

Noteworthy changes include:
- Reset test files before each test
- Don't disguise setup/teardown as tests
- Compare objects instead of JSON strings
- Don't use reserved words as variable names
- Drop some duplicated/obsolete tests
- Fix linting errors
- Remove shelljs usage
- DRY and simplify code
- 200 LOC less

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,6 +23,8 @@ install:
 build: off
 
 test_script:
+  # Workaround for https://github.com/appveyor/ci/issues/2420
+  - set "PATH=%PATH%;C:\Program Files\Git\mingw64\libexec\git-core"
   - node --version
   - npm --version
   - "npm test"
--- a/integration-tests/HooksRunner.spec.js
+++ b/integration-tests/HooksRunner.spec.js
@@ -17,425 +17,267 @@
  under the License.
  **/
 
-/* eslint  no-mixed-spaces-and-tabs : 0 */
-/* eslint  no-tabs : 0 */
+const Q = require('q');
+const path = require('path');
+const fs = require('fs-extra');
+const globby = require('globby');
 
-var cordova = require('../src/cordova/cordova');
-var HooksRunner = require('../src/hooks/HooksRunner');
-var shell = require('shelljs');
-var path = require('path');
-var fs = require('fs');
-var os = require('os');
-var Q = require('q');
-var child_process = require('child_process');
-var helpers = require('../spec/helpers');
-var PluginInfo = require('cordova-common').PluginInfo;
-var superspawn = require('cordova-common').superspawn;
+const HooksRunner = require('../src/hooks/HooksRunner');
+const cordovaUtil = require('../src/cordova/util');
+const cordova = require('../src/cordova/cordova');
+const { tmpDir, testPlatform } = require('../spec/helpers');
+const { PluginInfo, superspawn } = require('cordova-common');
+const { Q_chainmap } = require('../src/util/promise-util');
 
-var platform = os.platform();
-var tmpDir = helpers.tmpDir('hooks_test');
-var project = path.join(tmpDir, 'project');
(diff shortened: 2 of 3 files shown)


brodybits:
- +1 (+100) for removing eslint comments from HooksRunner spec
- +1 (+100) for migrating some `var`s to `const` and getting rid of a bunch of other vars
- +1 (+100) for less JavaScript

A couple questions:
- It looks like test 020 either disappeared or was merged into test 019. If that was your intention I wonder if we should consider renumbering the remaining tests?
- This change introduces fs-extra in the same commit as massive refactoring of an ugly test script. I would rather see fs-extra introduced in a separate commit (less messy - less icky).

Also looks like there are some CI issues:(

raphinesse:
@brodybits thanks for the feedback and bringing the CI problems to my attention.

I was able to fix all but one CI error (Appveyor/Node6). I actually suspect it to be spurious. But even if so, that's the kind of thing that should be prevented by this PR :unamused: 

Regarding your questions:

> It looks like test 020 either disappeared or was merged into test 019.

Yes, 020 was dropped in favor of 019 since it was actual a weaker version of the latter. The test title suggested that 020 should have been using callbacks, which it din't. So my guess is that it once tested functionality that has since been removed, but had been "fixed" instead of being removed. Thus becoming a duplicate.

> I wonder if we should consider renumbering the remaining tests?

Not if I'm to decide. IMHO the numbers only serve as a way to quickly identify tests. As long as there are no duplicates, I would see renumbering as unnecessary diff noise.

> This change introduces fs-extra in the same commit as massive refactoring of an ugly test script. I would rather see fs-extra introduced in a separate commit (less messy - less icky).

If you just want the change to `package.json` in a separate commit, that's fine with me. I would like to avoid anything beyond that. In a nutshell: IMHO it would be more work than it's worth.

brodybits:
> I was able to fix all but one CI error (Appveyor/Node6). I actually suspect it to be spurious.

What a bummer. As a side point I am starting to hope we can deprecate Node 6 considering that it has already gone from Active LTS to maintenance LTS with EOL next April.

> Yes, 020 was dropped in favor of 019 since it was actual a weaker version of the latter.

I would favor a comment explaining this.

> Not if I'm to decide.

Fine with me if we have the explanation comment ^^,

> If you just want the change to `package.json` in a separate commit, that's fine with me.

Hmm ... if I do this I think it would be smoothest for me tor raise this change in a separate PR that we can integrate into master or that you can include in this one. Would interrupt some other work.

I would be happy to give a final review for likely approval if we get these items resolved.

raphinesse:
> I would favor a comment explaining this.

As in mentioning it in the commit? I would not want to put a placeholder comment in the code.

> Hmm ... if I do this I think it would be smoothest for me tor raise this change in a separate PR that we can integrate into master or that you can include in this one. Would interrupt some other work.

It's fine, I can do this. I just wanted to confirm if separating out the `package.json` change was all you had in mind.

dpogue:
> As a side point I am starting to hope we can deprecate Node 6 considering that it has already gone from Active LTS to maintenance LTS with EOL next April

We won't drop node 6 support until next April at the earliest. We can add a deprecation message for users that are running it, indicating that it will not be supported as of April.

brodybits:
> As in mentioning it in the commit?

ok

> It's fine, I can do this.

Thanks, yes. I wonder if the commit should refer to CB-14140 since it would actually be part of that improvement. Leaving it up to you.

raphinesse:
@brodybits thanks for the review. I prepared the commits as discussed

brodybits:
I would say merge it (with rebase) assuming the CI builds pass.

raphinesse:
Exactly my plan 😁

PR #617, opened 2018-06-15
Set the fetch option to true in plugman

dpogue:
This should resolve the issues brought up in [these comments](https://github.com/apache/cordova-lib/commit/474a33bebdf30b73ad5067bcf94fec55027c4155#r29314973)

Should go in to the next patch release

PR #624, opened 2018-07-04
CB-13055 Fold all fetch options to `true`

raphinesse:
Manual constant folding assuming any `fetch` options to be `true`.

Extends #617

### DIFF
--- a/src/plugman/fetch.js
+++ b/src/plugman/fetch.js
@@ -63,22 +63,10 @@ function fetchPlugin (plugin_src, plugins_dir, options) {
         if (result) {
             if (result[1]) { options.git_ref = result[1]; }
             if (result[2]) { options.subdir = result[2]; }
-            // if --fetch was used, throw error for subdirectories
 
+            // throw error for subdirectories
             if (options.subdir && options.subdir !== '.') {
-                events.emit('warn', 'support for subdirectories is deprecated and will be removed in Cordova@7');
-                if (options.fetch) {
-                    return Q.reject(new CordovaError('--fetch does not support subdirectories'));
-                }
-            }
-
-            // Recurse and exit with the new options and truncated URL.
-            var new_dir = plugin_src.substring(0, plugin_src.indexOf('#'));
-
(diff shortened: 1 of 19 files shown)


raphinesse:
Will check out the failures on Travis :unamused:

raphinesse:
Fixed `integration-tests/plugman_uninstall.spec.js`. Should be good now.

raphinesse:
The test file that causes the errors on Windows is nasty. I'll DRY it first.

raphinesse:
OK, I fixed up a few test corner cases in the first commit and added a commit that removes all usage of fetch options in the tests. Could you please take another look @dpogue?

raphinesse:
OK, I'm done waiting for those awfully slow AppVeyor Tests just to see that one of them timed out again. :rage4:

src/plugman/fetch.js
```diff
@@ -63,22 +63,10 @@ function fetchPlugin (plugin_src, plugins_dir, options) {
...
-            }
-
-            // Recurse and exit with the new options and truncated URL.
-            var new_dir = plugin_src.substring(0, plugin_src.indexOf('#'));
-
-            // skip the return if user asked for --fetch
-            // cordova-fetch doesn't need to strip out git-ref
-            if (!options.fetch) {
-                return fetchPlugin(new_dir, plugins_dir, options);
+                return Q.reject(new CordovaError('Cordova does not support subdirectories'));
```
janpio: Ugh, this broke the documented way to install plugin tests: https://github.com/apache/cordova-plugin-test-framework#tldr-try-it

janpio: This is also still documented at https://github.com/apache/cordova-cli/blob/master/doc/readme.md#plugin-spec

dpogue: That `:subdirectory` syntax was specific to the old Cordova fetching code (which has been removed), and is not supported by npm

janpio: https://github.com/apache/cordova-lib/issues/722

PR #622, opened 2018-06-29
CB-14166: (cli) Fixed issue when installing plugins on windows

AlmirKadric:
The npm package name is already wrapped in quotes to prevent special characters from being execute. However the additional win32 code was further injecting quotes around the version causing npm to return a EINVALIDTAGNAME error. The win32 code is redundant and seems to not be required.

### Platforms affected
Windows, CLI

### What does this PR do?
Fix plugin installation on windows platform

### What testing has been done on this change?
Ran it locally and error vanished

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [x] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

PR #627, opened 2018-07-17
CB-14211: Run test suite in PowerShell and cmd.exe

oliversalzburg:
### Platforms affected
None

### What does this PR do?
Runs the test suite in both PowerShell and Command Prompt. Previously only one was checked.

I like to believe it tests itself :D

### Checklist

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,6 +7,10 @@ environment:
   - nodejs_version: "8"
   - nodejs_version: "10"
 
+platform:
+  - CommandPrompt
+  - PowerShell
+
 install:
   - ps: Install-Product node $env:nodejs_version
   - npm install
@@ -27,4 +31,20 @@ test_script:
   - set "PATH=%PATH%;C:\Program Files\Git\mingw64\libexec\git-core"
   - node --version
   - npm --version
-  - "npm test"
+
+for:
+-
+  matrix:
+    only:
+      - platform: CommandPrompt
+
+  test_script:
+  - cmd: "npm test"
+
+-
+  matrix:
+    only:
+      - platform: PowerShell
+
+  test_script:
+  - ps: "npm test"


appveyor.yml
```diff
@@ -7,6 +7,10 @@ environment:
   - nodejs_version: "8"
   - nodejs_version: "10"
 
+platform:
+  - cmd
```
janpio: Is this a name you chose or something that AppVeyor has named?
If it's your choice: Maybe make it `use_cmd` to make this clearer.

oliversalzburg: I think that can be whatever you want. If you prefer a prefix, I can add it, but I thought it's clear enough :D

oliversalzburg: But, what about using `CommandPrompt` and `PowerShell`? Or do you prefer to have a verb in there?

janpio: Does doesn't really matter, my confusion came from assuming this was a "magic string" you could set to trigger some platform - but it seems this is just selecting which part of the script is being executed.

raphinesse:
While I think that it's a good idea to test in different shells, this doubles the amount of jobs, right? Currently, AppVeyor tests take *very* long already. Since the jobs are only executing sequentially and some of our tests are awfully slow, it takes about **one hour** for them to finish. So with these changes I'm expecting a runtime of about **two hours**. :disappointed: 

IMO, we should only increase the number of jobs if it has no drastic impact on the overall build time.

I'm currently working on cleaning up the biggest runtime-offenders, so that we hopefully can reduce our test duration drastically as a next step. However, I suspect these changes are still a few weeks from being ready to land. Right now the tests take about 480s on my machine with the two biggest offenders amounting to 305s. Say I can reduce the overall run time by 50% that would still leave us with a 10 minute run time per job on the CI. Which would still be 30 minutes without this change or 60 minutes with it. Which I would still be inclined to avoid.

I guess what I'm trying to say is: I would very much prefer to find some way to test this with less impact on the CI run time. Like excessive unit tests for `superspawn` for example.

oliversalzburg:
I understand that sentiment completely. The idea just came up after trying to analyze issues relating to shell interaction, which this component does. Because this is prone to error, it should be included in an automatic testing approach IMHO.

A test suite that takes 2h to complete is annoying. Having to run a test suite on your own workstation 20 times because you're trying to track down an issue introduced by shell interaction is even more so.

It's probably overkill to run all 3 NodeJS versions on both shells though, now that I think about it.

raphinesse:
Yes, for a smaller impact, running only one version with both shells would be a start

brodybits:
Interesting failure on Node 10 on PowerShell [[1]](https://ci.appveyor.com/project/ApacheSoftwareFoundation/cordova-lib/build/1.0.1366/job/55c71ss4ex5g0are#L80) / [[2]](https://ci.appveyor.com/project/ApacheSoftwareFoundation/cordova-lib/build/1.0.1366/job/55c71ss4ex5g0are#L199)

Should probably be fixed. See also: [[3]](https://nodesource.com/blog/understanding-the-buffer-deprecation-in-node-js-10/), [[4]](https://medium.com/@jasnell/node-js-buffer-api-changes-3c21f1048f97)

[1] <https://ci.appveyor.com/project/ApacheSoftwareFoundation/cordova-lib/build/1.0.1366/job/55c71ss4ex5g0are#L80>
[2] <https://ci.appveyor.com/project/ApacheSoftwareFoundation/cordova-lib/build/1.0.1366/job/55c71ss4ex5g0are#L199>
[3] <https://nodesource.com/blog/understanding-the-buffer-deprecation-in-node-js-10/>
[4] <https://medium.com/@jasnell/node-js-buffer-api-changes-3c21f1048f97>

oliversalzburg:
@brodybits Nice find.

So right after we decide that it's probably not worth running all NodeJS versions on both shells, we see failures with one specific version in one specific shell.

That somewhat contradicts the decision to scale down the test matrix, doesn't it?

raphinesse:
> So right after we decide that it's probably not worth running all NodeJS versions on both shells, we see failures with one specific version in one specific shell.

It sure looks that way. I *think* that I saw the Buffer deprecation warning while working on some other Cordova stuff before this PR. But I couldn't say if it was the same code that failed. :man_shrugging:

janpio:
So we have CI specific bugs as well? Yay!

oliversalzburg:
@raphinesse Yeah, well observed. I was looking at that output earlier and could not quite figure out what exactly the failure was.

> PowerShell fails with NativeCommandError when a process writes to StdErr. Caused by this line: https://github.com/stevenvachon/nopter/blob/master/appveyor.yml#L37

From: https://help.appveyor.com/discussions/problems/10014-false-build-fail-status

Now that's a bit annoying. The `appveyor.yml` that he links to runs the tests in succession in both shells btw. Maybe that would be an option too.

janpio:
So, how should we continue here? 
Was an issue created for the "Buffer deprecation warning" stuff?

raphinesse:
@janpio the deprecation warning would have been resolved since. For the rest: I still don't want to double test times on AppVeyor

janpio:
So rebasing on `master` should fix the failing test @raphinesse? If so, could you please take care of that @oliversalzburg? Thanks.

I agree with the test run time - we will have to find another solution here somehow :/

raphinesse:
@janpio @oliversalzburg This all started with #622 which challenges the correctness of the escaping done in lib for some shells. Thus, we should start with unit tests for `superspawn` in `cordova-common` that test (or show the absence of) escaping. For example by calling `superspawn` to write some special characters to a file and check the result.

After we have a better understanding of the problems at hand, we could select a few tests to form a smoke test suite for PowerShell that we run on CI in addition to our regular tests. What do you think?

janpio:
I fully agree, I spent the last few weeks making a ruby library crossplatform and had a lot of fun with the escaping and quoting of command line params as well. We should probably create an issue to collect some more information what exactly is going wrong etc.

(Btw: I couldn't replicate the JIRA issue https://github.com/apache/cordova-lib/pull/622 is trying to solve on Windows)

PR #629, opened 2018-07-26
Spec cleanup

raphinesse:
This is part of an ongoing effort to make the tests of `cordova-lib` faster, more reliable and more maintainable.

This PR is mainly about maintainability and a bit about reliability. It focuses on the two suites `pkgJson.spec` and `pkgJson-restore.spec`. These are the two suites with the longest duration (taken together responsible for about 3/5 of the E2E test runtime). Furthermore, their code quality was so awful that I could reduce the LOC by over 30% respective over 50% without dropping any test cases or sacrificing any clarity (IMHO even drastically improving it).

Other highlights include:

- Use new, verbose test reporter for better information, especially in CI
- Unique, randomized directory name generated by `tmpDir` to prevent fs caching from affecting tests
- Proper suite-wide setting of test timeouts without affecting the global default
- Some minor improvements here and there

As I said before, this is part of an ongoing effort. In the next steps I will try to make the tests in the two `pkgJson` suites more focused (they are often testing more/other things than what it says in their description) and faster (will try to mock `cordova-fetch` which is a bit of a PITA for these E2E tests).

I'm also aware that many of the factored-out helpers in these two suites are identical. I'd like to leave them that way for now though and consider reusing them across suites after the steps outlined above.

Also, if you review the changes be sure to ignore white space changes. Diff anchoring to `it`/`describe` might also help. If someone is interested in that, I can provide more granular commits for these. But I thought it would even be worse to have a 100 commit PR :unamused:

### DIFF
--- a/integration-tests/pkgJson-restore.spec.js
+++ b/integration-tests/pkgJson-restore.spec.js
@@ -14,1449 +14,689 @@
     specific language governing permissions and limitations
     under the License.
 */
-var helpers = require('../spec/helpers');
-var path = require('path');
-var shell = require('shelljs');
-var events = require('cordova-common').events;
-var ConfigParser = require('cordova-common').ConfigParser;
-var cordovaPlatform = require('../src/cordova/platform');
-var prepare = require('../src/cordova/prepare');
-var cordova_util = require('../src/cordova/util');
-var cordovaPlugin = require('../src/cordova/plugin');
-// cordova = require('../src/cordova/cordova'),
-var TIMEOUT = 240 * 1000;
+
+const path = require('path');
+const fs = require('fs-extra');
+const prepare = require('../src/cordova/prepare');
+const cordovaPlugin = require('../src/cordova/plugin');
+const cordovaPlatform = require('../src/cordova/platform');
+const { ConfigParser } = require('cordova-common');
+const { listPlatforms, requireNoCache } = require('../src/cordova/util');
+const { tmpDir: getTmpDir, testPlatform, setDefaultTimeout } = require('../spec/helpers');
 
 /** Testing will check if "cordova prepare" is restoring platforms and plugins as expected.
 *   Uses different basePkgJson files depending on testing expecations of what (platforms/plugins/variables)
 *   should initially be in pkg.json and/or config.xml.
 */
+describe('restore', function () {
+    setDefaultTimeout(240 * 1000);
 
-// Use basePkgJson
-describe('tests platform/spec restore with --save', function () {
-    var tmpDir = helpers.tmpDir('platform_test_pkgjson2');
-    var project = path.join(tmpDir, 'project');
-    var results;
-    var listener = function (res) { results = res; };
+    const fixturesPath = path.join(__dirname, '../spec/cordova/fixtures');
+    var tmpDir, project, pkgJsonPath, configXmlPath;
 
-    beforeEach(function () {
-        jasmine.DEFAULT_TIMEOUT_INTERVAL = 300 * 1000;
-        shell.rm('-rf', project);
-        // Copy then move because we need to copy everything, but that means it will copy the whole directory.
-        // Using /* doesn't work because of hidden files.
-        shell.cp('-R', path.join(__dirname, '..', 'spec', 'cordova', 'fixtures', 'basePkgJson'), tmpDir);
-        shell.mv(path.join(tmpDir, 'basePkgJson'), project);
-        process.chdir(project);
+    beforeEach(() => {
+        tmpDir = getTmpDir('pkgJson');
+        project = path.join(tmpDir, 'project');
+        pkgJsonPath = path.join(project, 'package.json');
+        configXmlPath = path.join(project, 'config.xml');
         delete process.env.PWD;
-        events.on('results', listener);
     });
 
-    afterEach(function () {
-        events.removeListener('results', listener);
-        cordova_util.requireNoCache(path.join(process.cwd(), 'package.json'));
-        process.chdir(path.join(__dirname, '..')); // Needed to rm the dir on Windows.
-        shell.rm('-rf', tmpDir);
+    afterEach(() => {
+        process.chdir(__dirname); // Needed to rm the dir on Windows.
+        fs.removeSync(tmpDir);
     });
 
-    // Factoring out some repeated checks.
-    function emptyPlatformList () {
-        return cordovaPlatform('list').then(function () {
-            var installed = results.match(/Installed platforms:\n {2}(.*)/);
-            expect(installed).toBeDefined();
-            expect(installed[1].indexOf(helpers.testPlatform)).toBe(-1);
-        });
-    }
-    /** Test#000 will check that when a platform is added with a spec, it will
-    *   add to pkg.json with a '^' and to config.xml with a '~'. When prepare is run,
-    *   pkg.json will have no change and config.xml (first char) will change from a '~' to a '^'.
-    */
-    it('Test#000 : tests that the spec (~,^) is added and updated as expected in config.xml', function () {
-        var cwd = process.cwd();
-        var pkgJsonPath = path.join(cwd, 'package.json');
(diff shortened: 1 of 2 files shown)


raphinesse:
Any thoughts on this? I'll assume lazy consensus there's no feedback.

The test failure is spurious. If anyone has an idea what could have caused it, I'll be happy to investigate.

brodybits:
Sorry I didn't get a chance to look through this one, looks pretty important from the commit messages. I gotta say that spurious test failures do not surprise me so much.

brodybits:
I tried using `--anchored=describe --anchored=it` options (git 2.18.0 with help from homebrew) to look at 08459accb6496732e5cd43f8f9ea2d08c573e7b7, gotta say it doesn't seem to help. I also tried opendiff, no better. I think I would have to do side-by-side before and after file comparison to check the changes. I think one function at a time would really help.

raphinesse:
Thanks for taking the time @brodybits!

I can take out the first six commits so we can land it in a separate PR.

Regarding the other changes: As I mentioned, I have about a hundred commits in private breaches that originally made up those changes. Sadly they are not really straightforward and producing a nice history, would approximately equal redoing most of the changes from scratch. I guess you understand that I'm not really keen on doing that kind of work.

Last time I checked, I was able to produce diffs that were anchored to the different tests. I will check that again when I find the time. Also notice that I wrote extensive commit messages for these commits, trying to convey what is going on.

How do you wish to proceed with this?

brodybits:
@raphinesse I would like to get the first 6 commits (0917185..a94a5ca) and FIXUP in a8015e6 landed asap, hereby approve these changes. (Not sure if FIXUP in a8015e6 was supposed to be squashed into another commit or not, I am fine either way.) Build started to look a little flakey once I merged #631 (the contributed change itself was green), which I would really rather not revert unless absolutely necessary, started working on followup fix in #632 and tried the wanted spec commits there.

I also hope we can get the build more consistently green at the same time.

raphinesse:
@brodybits  Anchored diff works for me (per commit, of course). Example

```
git show --ignore-space-change --anchored=it f9686c0bc09f
```

brodybits:
What is the motivation to move the tests around in f5ffeec?

raphinesse:
@brodybits It's about getting rid of the last of all the useless describe blocks that were originally in the spec file. PuI put it in a separate commit to not make the diff completely useless.

brodybits:
Can you explain why `events` is removed? I guess it is not needed any longer?

raphinesse:
@brodybits Correct, `events` isn't needed anymore.

Previously it was used as part of an overly complicated function to get the list of installed platforms. This is now done much simpler using `listPlatforms` from `cordova/util`.

raphinesse:
@brodybits since you seem to be interested in the nitty gritty details, I prepared two branches with the detailed history for these refactorings. These are not necessarily up-to-date with the changes in this PR nor are they cleaned up. However, they might give you a better idea of what was done.

I suggest looking at the `pkgJson` branch first, since I did the refactoring second and had a better idea of what was expecting me. I hope this helps.

https://github.com/apache/cordova-lib/compare/master...raphinesse:verbose-pkgJson-spec-cleanup
https://github.com/apache/cordova-lib/compare/master...raphinesse:verbose-restore-spec-cleanup

brodybits:
Thanks @raphinesse, just saw your comments. I actually did my own work to break f9686c0 into smaller parts (along with f5ffeec ("merge more describe blocks"), c606e1f ("better names"), and a couple little things I would like to fix) in: https://github.com/apache/cordova-lib/compare/master...brodybits:bb-cleanup-restore-spec-in-parts-wip1

It would be great if you can move the last commit (4d139e6 "Cleanup pkgJson.spec") into a separate PR that I (or someone else) could maybe review at some more leisure.

I can take a quick look at the branches you prepared, probably not in very deep detail. The changes do look really good in general and I think I will be taking some of the changes "in faith".

raphinesse:
Thanks for all the review effort, @brodybits.

Frankly, I had hoped to merge the whole original PR "in faith". The preparation of this PR by itself already was a lot of work. And getting the tests to a state where I would actually be satisfied with the result will require a lot more work.

As I tried to convey in the original PR description, the two suites I refactored here were written very poorly. And although the refactor made them way more approachable and maintainable, they still are not good test suites in my book for various reasons. So IMO, they will have to undergo further radical changes.

Consequently, I did not want to spend much time waiting for this change to be reviewed in detail and polished to perfection or even formulating this very message, but rather get on with working on the problems I tried to fix here.

In the next few months, my free time will be scarce, and it's unlikely that I will spend it on something I don't really want to do :unhappy:

integration-tests/pkgJson-restore.spec.js
```diff
@@ -14,1449 +14,681 @@
...
+            }).then(function () {
+                // Add PLATFORM_2 to project with --save
+                return cordovaPlatform('add', PLATFORM_2, {save: true});
+            }).then(function () {
+                // Both platforms are installed but only PLATFORM_2 is in pkg.json
+                expect(installedPlatforms()).toEqual([PLATFORM_2, PLATFORM_1]);
+                expect(getPkgJson('cordova.platforms')).toEqual([PLATFORM_2]);
+            }).then(function () {
+                // Remove all platforms without --save.
+                return cordovaPlatform('rm', [PLATFORM_1, PLATFORM_2]);
```
brodybits: Looks like a very minor change in tested behavior from the following:
```js
                // Remove helpers.testPlatform without --save.
                return cordovaPlatform('rm', ['browser']);
```

I would favor avoiding change in tested behavior during this kind of cleanup commit if possible.

integration-tests/pkgJson-restore.spec.js
```diff
@@ -14,1449 +14,681 @@
...
+const { ConfigParser } = require('cordova-common');
+const { listPlatforms, requireNoCache } = require('../src/cordova/util');
+const { tmpDir: getTmpDir, testPlatform, setDefaultTimeout } = require('../spec/helpers');
 
 /** Testing will check if "cordova prepare" is restoring platforms and plugins as expected.
 *   Uses different basePkgJson files depending on testing expecations of what (platforms/plugins/variables)
 *   should initially be in pkg.json and/or config.xml.
 */
+describe('restore', function () {
+    setDefaultTimeout(240 * 1000);
```
brodybits: minor nit: I would rather see this defined as a constant at the beginning, even if it now appears only once.

brodybits:
I would personally be fine with merge "in faith", would be a little happier if you would first ask on mailing list with subject something like "Any objections with spec updates in faith?". I did find a couple minor things already which I just commented in the changes, would be happy to deal with them after the fact.

I was also wondering if we should consider breaking some of the spec scripts into separate files, maybe not so urgent with the cleanup.

I definitely see your point about the quality of the existing tests, definitely grateful for your efforts to clean it up.

integration-tests/pkgJson-restore.spec.js
raphinesse: Yes, I changed it slightly so the test makes actual sense and tests what the description says. You're right that this should go into a separate PR though. I have some similar changes prepared anyway.

integration-tests/pkgJson-restore.spec.js
raphinesse: I actually like this better the way it is. The function name already tells us what the argument is. So it's not really a magic number.

Something like `setDefaultTimeout(4 * ONE_MINUTE_IN_MS)` would be nicer, but hopefully we will be able to reduce this soon anyway.

That being said, I have no strong feelings about this. So I will change it and merge this part of the PR.

integration-tests/pkgJson-restore.spec.js
raphinesse: I only joined the two adjacent calls to `cordovaPlatform('rm', ...);` into one. I wouldn't say that this changes tested behavior. Especially since `restore` is under test here, not `cordovaPlatform`.

PR #630, opened 2018-07-30
Improve plugman/uninstall.js messages

janpio:
Currently you get an error message like this if you try to uninstall a plugin while an unsupported platform (in this case a folder called `.git`) is present:

`(node:7428) UnhandledPromiseRejectionWarning: CordovaError: .git not supported.`

The error message doesn't make it obvious that this is about a "platform". This PR fixes this to be more clear.

While fixing this one case, I also encountered other messages in the same file that could be more precise and include platform/plugin in the output, so I changed those as well.

--- 

Complete output from where I encountered it:

```
λ ionic cordova plugin remove cordova-sqlite-storage
> cordova plugin remove cordova-sqlite-storage --save
(node:7428) UnhandledPromiseRejectionWarning: CordovaError: .git not supported.
    at Function.module.exports.uninstallPlatform (C:\nvm\v8.11.3\node_modules\cordova\node_modules\cordova-lib\src\plugman\uninstall.js:76:25)
    at C:\nvm\v8.11.3\node_modules\cordova\node_modules\cordova-lib\src\cordova\plugin\remove.js:76:50
    at _fulfilled (C:\nvm\v8.11.3\node_modules\cordova\node_modules\cordova-lib\node_modules\q\q.js:787:54)
    at self.promiseDispatch.done (C:\nvm\v8.11.3\node_modules\cordova\node_modules\cordova-lib\node_modules\q\q.js:816:30)
    at Promise.promise.promiseDispatch (C:\nvm\v8.11.3\node_modules\cordova\node_modules\cordova-lib\node_modules\q\q.js:749:13)
    at C:\nvm\v8.11.3\node_modules\cordova\node_modules\cordova-lib\node_modules\q\q.js:509:49
    at flush (C:\nvm\v8.11.3\node_modules\cordova\node_modules\cordova-lib\node_modules\q\q.js:108:17)
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
    at Function.Module.runMain (module.js:695:11)
(node:7428) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:7428) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
(Note that the fact that `.git` should probably not be handled as a platform is tracked in https://issues.apache.org/jira/projects/CB/issues/CB-14231 and not part of this issue/PR)

### DIFF
--- a/integration-tests/plugman_uninstall.spec.js
+++ b/integration-tests/plugman_uninstall.spec.js
@@ -181,7 +181,7 @@ describe('uninstallPlatform', function () {
                 .then(function (result) {
                     fail();
                 }, function err (errMsg) {
-                    expect(errMsg.toString()).toContain('atari not supported.');
+                    expect(errMsg.toString()).toContain('Platform "atari" not supported.');
                 });
         }, 6000);
 
@@ -217,9 +217,9 @@ describe('uninstallPlugin', function () {
                 .then(function (result) {
                     var del = common.spy.getDeleted(emit);
                     expect(del).toEqual([
-                        'Deleted "C"',
-                        'Deleted "D"',
-                        'Deleted "A"'
+                        'Deleted plugin "C"',
+                        'Deleted plugin "D"',
+                        'Deleted plugin "A"'
                     ]);
                 });
         });
@@ -229,7 +229,7 @@ describe('uninstallPlugin', function () {
                 .then(function (result) {
                     fail();
                 }, function err (errMsg) {
-                    expect(errMsg.toString()).toEqual('"C" is required by (A) and cannot be removed (hint: use -f or --force)');
+                    expect(errMsg.toString()).toEqual('Plugin "C" is required by (A) and cannot be removed (hint: use -f or --force)');
                 });
         }, 6000);
 
@@ -237,7 +237,7 @@ describe('uninstallPlugin', function () {
             return uninstall.uninstallPlugin('C', plugins_install_dir, {force: true})
                 .then(function () {
                     var del = common.spy.getDeleted(emit);
-                    expect(del).toEqual(['Deleted "C"']);
+                    expect(del).toEqual(['Deleted plugin "C"']);
                 });
         });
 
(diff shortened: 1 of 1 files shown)


raphinesse:
Looks good to me. You need to update some tests to check for the new messages, though.

brodybits:
If I would do the following commands on a new Cordova project using cordova@8, npm@6.2.0 (node@10.7.0) on my mac:
- `cordova plugin add cordova-sqlite-storage`
- `cordova platform add ios`
- `cordova platform add browser`
- `cordova platform add android`

then remove `cordova-sqlite-storage` using the following command: `cordova plugin rm cordova-sqlite-storage`, I get the following otuput, with no ugly messages:

```sh
Uninstalling cordova-sqlite-storage from android
Android Studio project detected
Uninstalling cordova-sqlite-storage from browser
js-module uninstall called : plugins/cordova-sqlite-storage/www/SQLitePlugin.js
Uninstalling cordova-sqlite-storage from ios
Removing "cordova-sqlite-storage"
Removing plugin cordova-sqlite-storage from config.xml file...
Removing cordova-sqlite-storage from package.json
```

I wonder if the difference is due to Windows vs mac, node/npm version, or how your project is setup?

Should I try the same thing on my Windows machine?

brodybits:
Thanks @janpio for the response. Shouldn't that information be in the description, and shouldn't the JIRA number be in the title?

I gotta say that your use case sounds valid and interesting, not so surprised that CLI would have trouble with it. I think it would be nice to see a more clear description of your use case somewhere in JIRA and/or PR description.

janpio:
> Shouldn't that information be in the description, and shouldn't the JIRA number be in the title?

No, as this PR does nothing to solve the problem (which is described in the JIRA issue I linked to and to which a PR fixing that issue will refer to). This other problem just showed me that the error message here in this file could be better, so I created a PR directly to improve the outputs. (One could create a JIRA issue "cordova-lib/plugin/uninstall message are unclear" but that is just busy work we can skip imho - the PR here contains all the relevant information on what it does and why.)

PR #631, opened 2018-08-01
CB-14243: change dash to underscore for save-exact key reference

amaldonado:
### Platforms affected
This affects how plugins are added for both Android and IOS.

### What does this PR do?
This PR fixes CB-14243. Cordova-lib is ignoring save-exact b/c the wrong key name is being used to reference it in the options object that is propagated from cli.js.

Unit tests have been run and are passing.

### Checklist
✅Reported an issue in the JIRA database
✅Commit message follows the format: "CB-14243: {message}"
✅ Unit Tests pass

### DIFF
--- a/src/cordova/plugin/add.js
+++ b/src/cordova/plugin/add.js
@@ -87,7 +87,7 @@ function add (projectRoot, hooksRunner, opts) {
                         pluginInfoProvider: pluginInfoProvider,
                         variables: opts.cli_variables,
                         is_top_level: true,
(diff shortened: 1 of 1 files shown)


brodybits:
Thanks @amaldonado for the contribution, merged it since it definitely looks right to me. A couple points of feedback:

I think it could have been done a little bit better, raised #632 with the suggestion.

This behavior is evidently not covered by the unit test suite, marking this as a major TODO item in #632.

PR #634, opened 2018-08-14
Cleanup pkgJson-restore.spec

raphinesse:
Only the `pkgJson-restore.spec` part of #629 as requested by @brodybits. See #629 for a full description.

### DIFF
--- a/integration-tests/pkgJson-restore.spec.js
+++ b/integration-tests/pkgJson-restore.spec.js
@@ -367,7 +367,7 @@ describe('restore', function () {
         *   When both files contain the same platforms and cordova prepare is run,
         *   neither file is modified.
         */
-        it('Test#004 : if pkg.json and config.xml have the same platforms, do not modify either file', function () {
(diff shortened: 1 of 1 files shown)


raphinesse:
Thanks for the quick approval @brodybits :+1: 

Now there's some unrelated tests failing again. This really starts to piss me off... :unamused:

brodybits:
> Now there's some unrelated tests failing again.

I spotted that too, looks like something flakey. Definitely not the first time.

raphinesse:
A quick bisect says it was me in 60ba43944927c41b587d179639472253aaf0d47f

brodybits:
> A quick bisect says it was me in 60ba439.

Wonder if we should start to get rid of Q while you are at it?

raphinesse:
> Wonder if we should start to get rid of Q while you are at it?

I have a local branch for that and it's definitely not something to do while we're at something else. There are various blockers.

raphinesse:
Merged as CI failures were unrelated

PR #636, opened 2018-08-14
Fix broken unit tests

raphinesse:
This fixes some unit tests that appear to have been broken by 60ba43944927c41b587d179639472253aaf0d47f and then some that were uncovered during the fixing.

### DIFF
--- a/src/cordova/prepare.js
+++ b/src/cordova/prepare.js
@@ -31,7 +31,9 @@ var path = require('path');
 var config = require('./config');
 var _ = require('underscore');
 
-exports = module.exports = prepare;
+// Add indirection to be able to spy on prepare
+exports = module.exports = (...args) => exports.prepare(...args);
+module.exports.prepare = prepare;
 module.exports.preparePlatforms = preparePlatforms;
 module.exports.restoreMissingPluginsForPlatform = restoreMissingPluginsForPlatform;
 
--- a/integration-tests/.eslintrc.yml
+++ b/integration-tests/.eslintrc.yml
@@ -1,2 +1,4 @@
 env:
     jasmine: true
+rules:
+    prefer-promise-reject-errors: off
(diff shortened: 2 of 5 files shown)


src/cordova/prepare.js
```diff
@@ -31,7 +31,9 @@ var path = require('path');
 var config = require('./config');
 var _ = require('underscore');
 
-exports = module.exports = prepare;
+// Add indirection to be able to spy on prepare
+exports = module.exports = (...args) => exports.prepare(...args);
+module.exports.prepare = prepare;
```
brodybits: I am sorry to say that I really do not like this construct. It exports an API that looks like it may be consumed in multiple ways which I think we do not want to support in the future.

I would like to propose the following compromise solution: change `exports.prepare` to `exports._prepare`

raphinesse: I do not like doing this for testability only, but I've given this a lot of thought and this seems to be the best solution for function-exports. Look at it as an alias. It has the added benefit of allowing you to do
```js
const { prepare, preparePlatforms } = require('.../prepare');
```
which I find quite nice.

brodybits: Hmm I did not see that case before, it does make more sense now. However that kind of aliasing continues to strike me as ugly both to understand and to support in an API. This is of course symptomatic of the quality of the existing design.

I am also not so happy that the commit message does not describe that this kind of aliasing is done and that the comment in the code also does not really describe the aliasing.

I would favor the following kind of approach:
- keep the aliasing, with descriptive comments in both the commit and in the code, along with a TODO comment to improve the design in this area
- raise an issue to rework the code and possibly the API to solve this in a better way

raphinesse:
The reason that this has surfaced, seems to be https://github.com/jasmine/jasmine/issues/1590. I'll pin Jasmine to 3.1 and re-evaluate this PR later.

PR #637, opened 2018-08-14
Make coverage script consider e2e tests too

raphinesse:
Uses `nyc` for code coverage


brodybits:
`coverage` should have been removed from `.gitignore`, as I just proposed in PR #784

PR #641, opened 2018-08-15
Create test projects instead of asserting them

raphinesse:
With this PR, the test fixtures in `pkgJson-restore.spec` are created from a base fixture in the tests that need them.

Apart from better flexibility to change (will need this for performance tuning later) and less fixture files, this approach has the benefit that the test environment is defined in the tests themselves, making them easier to understand.


raphinesse:
Minor formatting changes and added TODO referencing apache/cordova-common#38 and apache/cordova-common#39.

PR #651, opened 2018-08-16
Remove all usage of Q-specific methods on Promise instances

raphinesse:
With this change landed, it should be safe to return native Promise instances from any modules consumed by `cordova-lib`.


raphinesse:
Actually, the *remove All* part of this PR is only true with the changes from #650 which was originally part of this branch. Should we merge this now anyway?

Issue #642, opened 2018-08-15
Flaky tests


raphinesse:
The tests have been unreliable, especially `e2e-tests` and especially on CI.

Since this issue is long-running, with no single point of failure or single fix, it should act as a meta-marker for related issues and PRs and a place to collect and document observed failures for later reference.

**Related:** #619, #623, #626, #628, #629, #633, #634, #635, #636, #639, #640, #641, #665 

### Open items

- [x] [CI-1]: Faulty promise handling in `plugman.uninstall` (#759)
- [x] [CI-2]: integration-tests/platform.spec.js:260:15 (assuming as fixed by switch to fs-extra)
- [x] [CI-3]: Timeout in integration-tests/platform.spec.js:263 (#679)
- [x] [CI-4]: Super strange failure in integration-tests/HooksRunner.spec.js (#653)
- [ ] [CI-5]: Error while running npm in pkgJson test 025
- [x] Re-do fixes from #636 that are unrelated to Jasmine update (#677)
- [ ] Investigate `requireNoCache`. I observed it failing to return fresh instances in combination with symlinks. We should use `fs.readJsonSync` where possible (#644)

[CI-1]: https://travis-ci.org/apache/cordova-lib/jobs/416116565#L1470
[CI-2]: https://ci.appveyor.com/project/ApacheSoftwareFoundation/cordova-lib/build/1.0.1401/job/3i2o6h6qpbyyaxl6#L1845
[CI-3]: https://ci.appveyor.com/project/ApacheSoftwareFoundation/cordova-lib/build/1.0.1402/job/2eokntx4n00891cr#L1926
[CI-4]: https://ci.appveyor.com/project/ApacheSoftwareFoundation/cordova-lib/build/1.0.1415/job/gq0h4096n2qfdo10#L1653
[CI-5]: https://ci.appveyor.com/project/ApacheSoftwareFoundation/cordova-lib/build/1.0.1422/job/92mlwsmwoxtlqac6#L1695

brodybits:
Add #695 (flaky plugman test on Windows)?

raphinesse:
I'm closing this as tests are pretty stable currently. Any further problems that pop up can be handled in separate issues.

PR #655, opened 2018-08-22
Fix shebang extraction in HooksRunner

raphinesse:
This PR fixes and improves shebang extraction in HooksRunner.

This will fix a class of CI failures from #642.


raphinesse:
GH is shuffling the rebased commits again :unamused: The test commit is actually the first one.

raphinesse:
@dpogue Thanks for the review!

PR #657, opened 2018-08-22
Add and apply ESLint rule against unhandled promises

raphinesse:
[promise/catch-or-return](https://github.com/xjamundx/eslint-plugin-promise/blob/master/docs/rules/catch-or-return.md)


raphinesse:
Closing until #650 and my upcoming `cordova.serve` PR are merged. I forgot this was depending on them.

PR #658, opened 2018-08-22
Remove deprecated platform support files

dpogue:
### Platforms affected
Amazon FireOS, Tizen, WebOS, and BlackBerry I guess... nothing we currently support

### What does this PR do?
Removes some outdated files that were missed for removal when support was dropped for some unsupported platforms.

Tests pass (I hope)


raphinesse:
I'm not very familiar with the different platforms, let alone old ones. Could you very briefly explain why the following files are no longer needed:
```
spec/cordova/metadata/browser_parser.spec.js
src/cordova/metadata/browser_parser.js
src/plugman/platforms/browser.js

src/plugman/util/android-project.js
src/plugman/util/plist-helpers.js

src/util/windows/jsprojManager.js
```

dpogue:
`src/cordova/metadata/browser_parser.js` was [moved into Cordova-Browser](https://github.com/apache/cordova-browser/blob/master/bin/template/cordova/browser_parser.js) as part of the Platform API, along with its spec file.

The plugman files, similarly, were incorporated into the platform-specific repos. ([Example](https://github.com/apache/cordova-browser/blob/master/bin/template/cordova/browser_handler.js), [Another example](https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/lib/AndroidProject.js)).

It was intended that these would be deleted from cordova-lib after people had a chance to update to platform versions that included the Platform API, which is why I'm removing them now.

PR #665, opened 2018-08-26
Radically focused restore.spec finally fast and reliable

raphinesse:
- Runtime: 217s -> 2s (on my machine)
- No dependency on the network
- Easier tests

Test names and comments will still have to be edited a bit if we want to merge this, but I first wanted to know what you think about this change.


raphinesse:
This should also resolve the test failures for #622

PR #682, opened 2018-09-06
GH-676 Remove browserify

### LINKED ISSUE #676: Remove `browserify` parameter/command
This was previously discussed that this parameter/command was never fully implemented or used, so it can safely be removed.

Side effect:
This will also remove the dependency to `cordova-js` in `cordova-lib` (if I didn't miss anything).

raphinesse:
Resolves #676


janpio:
Are there follow up changes necessary in `cordova-cli`? (docs, available CLI commands and params...)

Issue #681, opened 2018-09-06
Use native Promises instead of Q


raphinesse:
We have to do this top down in our dependency graph (see below). `cordova-cli` and `cordova-plugman` are already Q-less. So lib is ready to go next.

![Cordova Tooling dependency graph](https://graphviz2.glitch.me/graphviz?graph=digraph+cordovaToolingDependencies+%7B%0A++++node+%5Bfontname+%3D+%22monospace%22%5D%3B%0A%09%0A%09%7Bcordova+lib+fetch+create%7D+-%3E+common%0A%09%7Bplugman+cordova%7D+-%3E+lib%0A%09lib+-%3E+%7Bserve+create+fetch%7D%0A%09create+-%3E+fetch%0A%09%0A%09%23+Not+actually+tooling+but+still+cordova+dependencies%0A%09lib+-%3E+js++%5Bcolor+%3D+grey%5D%0A%09create+-%3E+%22app-hello-world%22+%5Bcolor+%3D+grey%5D%0A%09%22app-hello-world%22++%5Bcolor+%3D+grey%2C+fontcolor+%3D+grey%5D%0A%09js+%5Bcolor+%3D+grey%2C+fontcolor+%3D+grey%5D%0A%7D&layout=dot&format=svg&mode=download)

PR #687, opened 2018-09-08
Test, fix and cleanup `cordova serve`

raphinesse:
This is the last larger refactor that I needed/wanted to get done before tackling #681 and in turn #657.

I tried to keep the commits very focused to make the review easier. I also did manual testing of the server.


raphinesse:
I will go ahead and merge this in the coming days if there will be no further feedback.

PR #688, opened 2018-09-09
CB-14166 distinguish npm.cmd and npm.exe in win32, to check if escape is necessary (workaround solution)

knight9999:
### Platforms affected
all platforms in windows. 
(Especially the nodist is used as managing node versions.)

### What does this PR do?
This PR is based on the PR https://github.com/apache/cordova-lib/pull/622 .

When cordova executes npm command in windows, escaping arguments should be necessary.
However, this necessity depends on the type of npm command according to windows feature.
i.e.
npm.cmd   => escaping is necessary
npm.exe     => escaping is not necessary (escaping brings an error)

As discussions in https://issues.apache.org/jira/projects/CB/issues/CB-14166,
nvm-windows => only npm.cmd is installed and cordova uses npm.cmd
nodist => both npm.cmd and npm.exe are installed and cordova uses npm.exe

Therefore this PR adding function which checks the extension of npm command and skip escaping function if the extension is '.exe'.

And I improved escaping function itself from
```
parsedSpec.version.indexOf(character);
```
to
```
parsedSpec.version.indexOf(character) >= 0;
```
because the former return -1 (logically true) even if the `version` does not contain `character`.
The latter return true if and only if the `version` contains `character`;

In my windows 10, cordova 8.0.0 and Nodist environment (where npm.cmd and npm.exe are installed, cordova uses npm.exe automatically in superspawn.js), I checked by following commands,
```
> cordova create CB-14116-8
> cd CB-14116-8
> codova platform add windows@5.0.0
```
and confirm it works.
Before this PR, the above commands brings following error in Nodist environment.

```
npm ERR! code EINVALIDTAGNAME
npm ERR! Invalid tag name ""1"": Tags may not have any characters that encodeURIComponent encodes.
```

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     "read-chunk": "^3.0.0",
     "semver": "^5.3.0",
     "shebang-command": "^1.2.0",
-    "underscore": "^1.9.0"
+    "underscore": "^1.9.0",
+    "which": "^1.3.1"
   },
   "devDependencies": {
     "codecov": "^3.0.1",
--- a/src/plugman/fetch.js
+++ b/src/plugman/fetch.js
@@ -18,6 +18,7 @@
 */
 
 var fs = require('fs-extra');
+var which = require('which');
 var url = require('url');
 var underscore = require('underscore');
 var semver = require('semver');
@@ -136,10 +137,13 @@ function fetchPlugin (plugin_src, plugins_dir, options) {
                 }
 
                 if (process.platform === 'win32' && parsedSpec.version) {
-                    var windowsShellSpecialCharacters = ['&', '\\', '<', '>', '^', '|'];
-                    specContainsSpecialCharacters = windowsShellSpecialCharacters.some(function (character) {
-                        return parsedSpec.version.indexOf(character);
-                    });
+                    var npmExtension = path.extname(which.sync('npm'));
+                    if (npmExtension && npmExtension.toUpperCase() !== '.EXE') {
+                        var windowsShellSpecialCharacters = ['&', '\\', '<', '>', '^', '|'];
+                        specContainsSpecialCharacters = windowsShellSpecialCharacters.some(function (character) {
+                            return parsedSpec.version.indexOf(character) >= 0;
+                        });
+                    }
                 }
 
                 var fetchPluginSrc = specContainsSpecialCharacters ?


knight9999:
In this PR, I used `which` module is used in order to check `npm` command, because cordova-lib still uses cordova-fetch 1.x.
I feel that It is better to use `isNpmInstalled` function in the cordova-fetch 2.0.

raphinesse:
Thanks for your initiative on this and the great work debugging this issue, @knight9999.

However, I stand by what I said in https://github.com/apache/cordova-lib/pull/622#issuecomment-404836911:
> IMHO this should happen in `superspawn` not in `plugman`. `plugman` should have no obligation to worry whether fetch might use its arguments unescaped in a shell.

Or as @oliversalzburg puts it in [CB-14166]:

> The solution is still to properly escape when running commands on the shell, and to not escape when not running on the shell. Proper abstractions will take care of that for you.
>
> In this scenario, the only problem is that part of the code (`fetch.js`) decides that it already knows how the command is executed in the end, which it does not. The additional escaping introduced in this place is a mistake. It needs to be removed and `superspawn` needs to be taught how to properly escape, or, as I have been saying continuously, it should be replaced with something that works (`cross-spawn`, `execa`).

To sum it up:
- The current code that handles the version escaping is broken and needs to be removed, not patched.
- We need a properly working abstraction for our use case of "I want to run a command, without worrying about the platform I'm on". IMHO, the options to achieve that, ordered by the amount of work involved (ascending) are:
  - Patch `superspawn` to handle this case properly
  - Change `superspawn` to use `cross-spawn` instead of `child_process.spawn`
  - Get rid of `superspawn` in favor of `cross-spawn` or `execa`

Incidentally, these options are also ordered by my approval for them (ascending).


[CB-14166]: https://issues.apache.org/jira/browse/CB-14166?focusedCommentId=16605736&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16605736

knight9999:
@raphinesse Thanks for your valuable comments.
I agree your suggestion to fix this issue basically. 

Here I think this PR is only quick fix. Because improving cordova-common, cordova-fetch
and changing cordova-lib to refer them seem to require rather long time.  Actually the cordova-lib
still refers to cordova-fetch 1.3.0 yet.

I hope that this PR should be removed after fundamental improvements of cordova-common/cordova-fetch which just you pointed out are applied.
Even if this PR is not merged, this suggests the dirty and easy fix of cordova-lib for each user.

If your new PR is coming soon, it is no problem to skip this PR, I think.

raphinesse:
I prepared a proof-of-concept branch at https://github.com/raphinesse/cordova-common/pull/1 for option *ii* from my previous comment. If anyone wants to tackle this, feel free to use it.

janpio:
I am fine with any of those (including merging this PR) - but there maybe should be a test that can somehow confirm this behavior. It will be pretty useful when someone tackles the bigger issues/tasks @raphinesse mentioned.

brodybits:
This workaround solution is included in 8.1.x in PR #697.

This proposal as raised on master is now closed in favor of original solution proposed in #622.

Issue #689, opened 2018-09-11
Limit `requireCordovaModule` to Cordova modules


raphinesse:
## The Context

The context provided to hooks has a method called `requireCordovaModule`.

https://github.com/apache/cordova-lib/blob/1bc9dd050cbc12b046fd7e20628084d79d2f9fe4/src/hooks/Context.js#L50-L72

The apparent purpose of this method is to allow users to require modules of the exact version of Cordova that is running the hook. I guess that makes some sense, given how we currently use the Node module system to provide singletons.

## The Problem

However, this method conceptually leaks all internals of `cordova-lib`, including any of its dependencies, to the hooks context. In fact, the [only documentation on this method](https://cordova.apache.org/docs/en/latest/guide/appdev/hooks/#javascript) even **encourages** users to use it to require Q, which we could break, if we removed Q from our dependencies as part of #681. I can see how this might have seemed like a great idea when promises weren't available in native JS, but that has thankfully changed now.

## The Solution

I suggest we limit the modules allowed to import using `Context.prototype.requireCordovaModule` to a fixed whitelist containing only `cordova-*` dependencies of `cordova-lib` and `cordova-lib` itself.

As a first step, we should only issue a deprecation warning if anything not on the whitelist is required.

As an additional safety measure, we might warn users about using private interfaces when they deep-require something (like `require('cordova-lib/src/hooks/Context')` for example).

What do you think? I'm looking forward to your feedback.

raphinesse:
After #706, I'm beginning to think we should consider rolling out the deprecation step even earlier than the next major.

brodybits:
I would personally favor starting the deprecation now and limiting `requireCordovaModule` to Cordova modules in the next major.

brodybits:
Shouldn't we consider dropping support for `requireCordovaModule('cordova-lib')` or `requireCordovaModule('*/cordova-lib')` altogether?

raphinesse:
> Shouldn't we consider dropping support for `requireCordovaModule('cordova-lib')` or `requireCordovaModule('*/cordova-lib')` altogether?

I'm not sure what you mean by `requireCordovaModule('*/cordova-lib')`. After #709, The argument for `requireCordovaModule` must be either the name of a Cordova package (`cordova-*`) or a path to a module in a Cordova package (`cordova-*/**`).

[...]

PR #707, opened 2018-09-28
Deprecate `requireCordovaModule` for non-Cordova modules

raphinesse:
Addresses the deprecation part of #689.

### DIFF
--- a/spec/hooks/Context.spec.js
+++ b/spec/hooks/Context.spec.js
@@ -1,0 +1,108 @@
+/*!
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


brodybits:
I just pushed a couple updates that I think are needed:
- emit deprecation warning event in case of use of `requireCordovaModule` with non-Cordova module
- fix and update spec for `requireCordovaModule`

I am about to include the deprecation warning event in the hotfix proposed in #708.

brodybits:
@raphinesse go ahead and push your implementation. I was only doing this to help me with the hotfix.

raphinesse:
This version is not node@4 compatible, I'm afraid. But it should be straightforward to make it compatible.

@brodybits If you want to include this in #708, I think I can whip up a commit that makes this node@4 compatible.

brodybits:
FYI I am only using part of 1fdc38d58615e8a61accaf25b2418ba6ccd538a5 in patch PR #708. You can see it in <https://github.com/apache/cordova-lib/pull/708/commits/0b0ae7a39b34bcbe2d6aa80244c38ad3bda609d8>. The other 3 commits so far *are* part of patch #708. Please let me know if you would like me to explain this in more detail.

raphinesse:
@brodybits I noticed it and it's fine. The bug is fixed anyway.

raphinesse:
@brodybits Agreed. _If_ we should fork off another 8.x branch _from master_ we would need to apply something similar to #708 there too.

PR #709, opened 2018-09-28
Breaking changes in hooks/Context

### LINKED ISSUE #689: Limit `requireCordovaModule` to Cordova modules

raphinesse:
This includes a few changes to `hooks/Context` that necessitate a major release:
- Throw error if non-cordova module is required (Resolves #689)
- Remove outdated compatibility layer from 2015
- Lazily load `Context.cordova` to avoid cyclical dependency
- Use `const` for imports

### DIFF
--- a/spec/hooks/Context.spec.js
+++ b/spec/hooks/Context.spec.js
@@ -27,6 +27,27 @@ describe('hooks/Context', () => {
         Context = rewire('../../src/hooks/Context');
     });
 
+    describe('cordova', () => {
+        let context;
+
+        beforeEach(() => {
+            spyOn(Context.prototype, 'requireCordovaModule');
+            context = new Context();
+        });
+
+        it('is only loaded when accessed', () => {
+            expect(context.requireCordovaModule).not.toHaveBeenCalled();
+        });
+
+        it('is set to require("cordova-lib").cordova', () => {
+            const cordova = Symbol('cordova');
+            context.requireCordovaModule.and.returnValue({ cordova });
+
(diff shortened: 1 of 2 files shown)


raphinesse:
Thanks for merging! :hooray:

PR #710, opened 2018-09-29
Drop dependency on Q, use native promises

raphinesse:
- Part of https://github.com/apache/cordova/issues/7
- I took care to make a dedicated commit for every change that wasn't a simple search-and-replace
- Apart from the actual removal of Q this fixes a few problems that surfaced thereafter

### DIFF
--- a/src/cordova/requirements.js
+++ b/src/cordova/requirements.js
@@ -17,10 +17,10 @@
     under the License.
 */
 
-var cordova_util = require('./util');
-var Q = require('q');
-var CordovaError = require('cordova-common').CordovaError;
-var knownPlatforms = require('../platforms/platforms');
+const { object: zipObject } = require('underscore');
+const { preProcessOptions } = require('./util');
+const { CordovaError } = require('cordova-common');
+const knownPlatforms = require('../platforms/platforms');
 
 /**
  * Runs requirements check against platforms specified in 'platfoms' argument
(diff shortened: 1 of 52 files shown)


src/cordova/requirements.js
```diff
@@ -32,21 +32,18 @@ var knownPlatforms = require('../platforms/platforms');
  *   requirements check results for each platform.
  */
 module.exports = function check_reqs (platforms) {
-    return Q().then(function () {
-        var platforms = cordova_util.preProcessOptions(platforms).platforms; // eslint-disable-line no-use-before-define
+    return Promise.resolve().then(() => {
```
dpogue: Is there a reason to wrap this in a promise, given that the inner block is already returning `Promise.all`?

src/cordova/restore-util.js
```diff
@@ -111,7 +110,7 @@ function installPlatformsFromConfigXML (platforms, opts) {
 
         // No platforms to restore from either config.xml or package.json.
         if (comboArray.length <= 0) {
-            return Q('No platforms found in config.xml or package.json. Nothing to restore');
+            return Promise.resolve('No platforms found in config.xml or package.json. Nothing to restore');
```
dpogue: This pattern seems awkward to me, since it feels more like these should be emitted as logging events.
Does this function always resolve with a log message string, or with an array of the added platforms? Currently it does both and it would be better in terms of API to be consistent with the resolved value type.

src/hooks/HooksRunner.js
```diff
@@ -45,7 +44,7 @@ function HooksRunner (projectRoot) {
  */
 HooksRunner.prototype.fire = function fire (hook, opts) {
     if (isHookDisabled(opts, hook)) {
-        return Q('hook ' + hook + ' is disabled.');
+        return Promise.resolve('hook ' + hook + ' is disabled.');
```
dpogue: This method resolves with a string, or throws an error (not rejects), or resolves with the result of running scripts, or rejects with error from running scripts? :grimacing:

src/cordova/requirements.js
raphinesse: It's a poor-man's [`Promise.try`](http://cryto.net/~joepie91/blog/2016/05/11/what-is-promise-try-and-why-does-it-matter/).

In a nutshell: It ensures that this function  __always__ returns a `Promise`. Even if anything we do would cause an error to be thrown, thanks to the wrapping it will be converted to a rejection instead of bubbling up the call stack.

In this case, suppose `preProcessOptions(platforms)` would return `undefined`. Then our accessing the result's `platforms` property would result in an error to be thrown.

dpogue: Ahh, I actually just left a comment on the cordova-fetch PR referencing `Promise.try`. Since we seem to use this pattern frequently, maybe it makes sense to depend on something like https://github.com/tc39/proposal-promise-try (although that isn't published to npm :disappointed:)?

raphinesse: Yeah, sorry for doing a ELI5 here. And I agree with you regarding the lib: I just suggested the exact same thing in the cordova-fetch PR :grin:

raphinesse:
> we can make note of those other cleanup tasks somewhere

Could you take care of that please @dpogue? :pray:

piotr-cz:
Along with this change, docs for [Hooks: Sample Usage](https://cordova.apache.org/docs/en/latest/guide/appdev/hooks/#sample-usage) should be updated

raphinesse:
@piotr-cz Agreed. Actually the whole page is quite dated. I will try to fix that in the next few days.

raphinesse:
@piotr-cz See https://github.com/apache/cordova-docs/pull/958

PR #711, opened 2018-09-30
Fix 2 integration tests failures on macOS

dpogue:
### What does this PR do?
`os.tmpdir()` on macOS returns `/var/folders`, but `/var` is a symlink to `/private/var`.

We dereference symlinks in a bunch of places, which was leading to tests failing when they expected something to get called with a path of `/var/folders/cordova-lib-whatever/...` and they were actually getting called with a path of `/private/var/folders/cordova-lib-whatever/...`.

This change ensures that when we create a temporary directory, we get the real path before we return it to the tests.

Ran tests of macOS and they all passed.


raphinesse:
Could also be solved by using https://github.com/sindresorhus/temp-dir

PR #712, opened 2018-10-01
[WIP] 8.2.0 proposal WIP for discussion

brodybits:
I started working on a proposal to make an 8.2.0 minor release with some changes from the master branch. Purpose is to get some of the changes into users' hands asap, especially some more of the cleanup, deprecation message from GH-707, and eventually migration to fs-extra. This proposed 8.2.0 minor release is meant to come after the patch release to fix dependencies.

This WIP PR is FYI only, I will raise another PR to actually make the 8.2.0 minor release assuming that we actually decide to proceed with it. Some commits are marked "XXX" to mean that they will need rewording, better explanation, or to be moved at some point. Also some GH numbers are deliberately not marked right in order to avoid extra noise in the related GH issues and PRs.

I am specifically reaching out to @raphinesse since he expressed interest in a minor release with the deprecation message.


dpogue:
Now that cordova-lib 9 is released, are we still realistically considering an 8.2.0 release?

brodybits:
Nope

PR #721, opened 2018-11-21
[WIP] Update cordova-common & export common object - under discussion

brodybits:
* update unit tests to use objects from `cordova-lib.js` instead of `require('cordova-common')` whenever possible
* export common object from cordova-common, with quick unit-test verification that it works
* do not export `binname` attribute
* cordova-common@3 update

This means that tools like Cordova CLI would use the exported `common` object instead of explicitly requiring the `cordova-common` dependency. I think this would lead to both easier maintenance and reduced likelihood of violating the [cordova-common singleton rule](https://github.com/apache/cordova-coho/blob/master/docs/tools-release-process.md#cordova-common-singleton-rule).

If this proposal is accepted, I think we should deprecate use of individual `common` member objects from the exports (`events`, `configparser`, `PluginInfo`, and `CordovaError`) and remove them someday in the future.

If this proposal is not accepted for any reason, an alternative solution would be to add an export of the `CordovaLogger` object. Then Cordova CLI could drop the explicit cordova-common dependency.

### DIFF
--- a/cordova-lib.js
+++ b/cordova-lib.js
@@ -26,6 +26,7 @@ exports = module.exports = {
     get binname () {
         return this.cordova.binname;
     },
+    common: common,
     events: common.events,
     configparser: common.ConfigParser,
     PluginInfo: common.PluginInfo,
--- a/cordova-lib.js
+++ b/cordova-lib.js
@@ -20,6 +20,12 @@
 var common = require('cordova-common');
 
 exports = module.exports = {
+    set binname (name) {
+        this.cordova.binname = name;
+    },
+    get binname () {
+        return this.cordova.binname;
+    },
     common: common,
     events: common.events,
     configparser: common.ConfigParser,
--- a/cordova-lib.js
+++ b/cordova-lib.js
@@ -20,12 +20,6 @@
 var common = require('cordova-common');
 
 exports = module.exports = {
-    set binname (name) {
-        this.cordova.binname = name;
-    },
-    get binname () {
-        return this.cordova.binname;
-    },
     common: common,
     events: common.events,
     configparser: common.ConfigParser,
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "node": ">=6.0.0"
   },
   "dependencies": {
-    "cordova-common": "^2.2.0",
+    "cordova-common": "^3.0.0",
     "cordova-create": "^1.1.0",
     "cordova-fetch": "^1.3.0",
     "cordova-serve": "^2.0.0",
--- a/spec/cordova/create.spec.js
+++ b/spec/cordova/create.spec.js
@@ -18,9 +18,13 @@
 */
 
 var fs = require('fs-extra');
+
 var helpers = require('../helpers');
+
 var path = require('path');
-var events = require('cordova-common').events;
+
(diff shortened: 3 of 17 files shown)


brodybits:
Marked [WIP] due to Travis CI test failures and some more unit test updates needed.

brodybits:
Build is now green, looking forward to feedback and possible review.

cordova-lib.js
```diff
@@ -20,12 +20,6 @@
 var common = require('cordova-common');
 
 exports = module.exports = {
-    set binname (name) {
-        this.cordova.binname = name;
-    },
-    get binname () {
-        return this.cordova.binname;
-    },
```
brodybits: `binname` attribute is used by cordova-cli, hope we can remove it in the future - my bad again

dpogue: We can't. This is used by other tools that wrap cordova-lib, like `ionic`, to display their command-line tool name in help and error messages instead of `cordova`

dpogue:
Don't we still have an issue because cordova-lib depends on cordova-fetch, and cordova-fetch pulls in a copy of cordova-common?

I'd much rather find a safe workaround for the cordova-common singleton issues than muddy the exposed API surface of our other modules.

dpogue:
> * cordova-common@3 update

I'm not sure we can do that here until after the [stage 2](https://github.com/apache/cordova/issues/10) releases have been done.

brodybits:
If we would just export the `CordovaLogger` logger object like this, then cordova-cli should not have to import cordova-common on its own.

```diff
--- a/cordova-lib.js
+++ b/cordova-lib.js
@@ -30,6 +30,7 @@ exports = module.exports = {
     configparser: common.ConfigParser,
     PluginInfo: common.PluginInfo,
     CordovaError: common.CordovaError,
+    CordovaLogger: common.CordovaLogger,
     plugman: require('./src/plugman/plugman'),
     cordova: require('./src/cordova/cordova'),
     cordova_platforms: require('./src/platforms/platforms')
```

Do you see any problem with this kind of an idea?

dpogue:
That still doesn't mean that cordova-fetch and cordova-cli are using the same instance of CordovaLogger.

It seems like there are some options for doing real singletons in node >=6 using `global` and `Symbol` (https://derickbailey.com/2016/03/09/creating-a-true-singleton-in-node-js-with-es6-symbols/) but it might be too clever for what it's worth? That would almost need to be another major of cordova-common, and we just released 3.0.0 :confused:

brodybits:
> That still doesn't mean that cordova-fetch and cordova-cli are using the same instance of CordovaLogger.

Agreed, sure less than ideal. Does that make my idea broken somehow?

> there are some options for doing real singletons in node >=6 [...] but it might be too clever for what it's worth?

I would sure favor the "cleverness" of using real singletons over whatever we have to do to follow our own singleton rule. I would think that using real singletons would be better documented on the net, for example.

> That would almost need to be another major of cordova-common, and we just released 3.0.0

Would another minor or major release of cordova-common really be so bad, considering that its major number is still behind the major number of cordova-lib, Cordova platforms, etc.?

brodybits:
Closing in favor of a better solution

PR #741, opened 2018-12-03
Fix crash in `cordova requirements` due to an unbound function

### LINKED ISSUE #740: `cordova requirements android` crashes - `this.isCordova is not a function`
I left a comment here explaining what I believe is the problem: https://github.com/apache/cordova-lib/commit/af1a1b519ce792451875be438b7a795806445899#diff-82ee48eb5f4bcbbd61e7573c641594e3R36

Changing `this.isCordova` to `cordova_util.isCordova` fixes the problem for me.

bluetech:
Fixes #740.

### Platforms affected

All of them, I think.

### What does this PR do?

Functions in utils.js expect to be called with `this` bound to the module. So this call crashed with an error

```
    this.isCordova is not a function
```

This PR fixes the call to go through the proper binding.

It fixed my build. I also added a simple fail-before, pass-after unit test.

### DIFF
--- a/spec/cordova/requirements.spec.js
+++ b/spec/cordova/requirements.spec.js
@@ -1,0 +1,40 @@
+/**
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


brodybits:
Thanks @bluetech!

PR #743, opened 2018-12-05
Updated Platform Config Git URL Paths

erisu:
### Platforms affected
none

### What does this PR do?
- Update each platform config with a valid and active git URL.

none


janpio:
If it was invalid before, is this actually still needed?
(I have the feeling I asked this before somewhere...)

dpogue:
There was (is?) a `--git` flag you could pass when doing `cordova platform add` that would clone the latest from git. I suspect that's gone with the move to using npm for dependencies.

PR #750, opened 2019-03-14
Remove saving platforms/plugins to config.xml

dpogue:
### Platforms affected
All (via tooling)

### Motivation and Context

We've had one major version of cordova-lib that attempted to ensure both config.xml and package.json were kept in sync, and now it's time to push people more strongly towards using package.json.

### Description

We will still attempt to read values from config.xml and update them in package.json, but we will no longer reflect changes back to config.xml.
Whatever's saved in package.json should always have priority over what is read from config.xml.

The next phase will be to [improve the handling of package.json updates](https://github.com/apache/cordova-common/pull/34), and then in the next major to completely remove the code that looks at config.xml for platforms/plugins.

### Testing

Updated spec and E2E tests.

### Checklist

### DIFF
--- a/integration-tests/pkgJson.spec.js
+++ b/integration-tests/pkgJson.spec.js
@@ -105,13 +105,6 @@ describe('pkgJson', function () {
         };
     }
 
-    function stringContaining (substring) {
-        return {
-            asymmetricMatch: s => typeof s === 'string' && s.includes(substring),
-            jasmineToString: _ => `<stringContaining(${substring})>`
-        };
-    }
-
     const customMatchers = {
         toSatisfy: () => ({ compare (version, spec) {
             const pass = semver.satisfies(version, spec);
@@ -155,10 +148,8 @@ describe('pkgJson', function () {
                     // Check that the plugin and spec add was successful to pkg.json.
                     expect(getPkgJson('cordova.plugins')[pluginId]).toBeDefined();
                     expect(getPkgJson('dependencies')[pluginId]).tohaveMinSatisfyingVersion('1.1.2');
-                    // Check that the plugin and spec add was successful to config.xml.
-                    expect(getCfg().getPlugins()).toEqual([
-                        { name: pluginId, spec: specWithMinSatisfyingVersion('1.1.2'), variables: {} }
-                    ]);
+
+                    expect(getCfg().getPluginIdList()).toEqual([]);
                 }).then(function () {
                     // And now remove it with --save.
                     return cordova.plugin('rm', pluginId, { save: true });
@@ -267,12 +258,6 @@ describe('pkgJson', function () {
             return cordova.platform('add', PLATFORM, { save: true })
                 .then(function () {
                     expect(getPkgJson('cordova.platforms')).toEqual([PLATFORM]);
-
-                    // Config.xml and ios/cordova/version check.
-                    expect(getCfg().getEngines()).toEqual([{
-                        name: PLATFORM,
-                        spec: specSatisfiedBy(platformVersion(PLATFORM))
-                    }]);
                 }).then(function () {
                     return cordova.plugin('add', PLUGIN, { save: true });
                 }).then(function () {
@@ -281,7 +266,6 @@ describe('pkgJson', function () {
 
                     // Check that installed version satisfies the dependency spec
                     const version = pluginVersion(PLUGIN);
-                    expect(version).toSatisfy(getCfg().getPlugin(PLUGIN).spec);
                     expect(version).toSatisfy(getPkgJson(`dependencies.${PLUGIN}`));
                 });
         });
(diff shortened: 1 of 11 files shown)


src/cordova/platform/remove.js
```diff
@@ -54,9 +54,11 @@ function remove (hooksRunner, projectRoot, targets, opts) {
                     var platformName = target.split('@')[0];
                     var xml = cordova_util.projectConfig(projectRoot);
                     var cfg = new ConfigParser(xml);
-                    events.emit('log', 'Removing platform ' + target + ' from config.xml file...');
-                    cfg.removeEngine(platformName);
-                    cfg.write();
+                    if (cfg.getEngines && cfg.getEngines().some(function (e) { return e.name === platformName; })) {
```
raphinesse:
```suggestion
                    if (cfg.getEngines().some(e => e.name === platformName)) {
```

integration-tests/pkgJson.spec.js
```diff
@@ -550,37 +505,4 @@ describe('pkgJson', function () {
             });
         });
     });
-
-    // No pkg.json included in test file.
-    describe('local path is added to config.xml without pkg.json', function () {
-        beforeEach(() => useProject('basePkgJson13'));
```
raphinesse: Did you check if the fixture is still used now that this has been deleted? What is the new behavior without a package.json?

dpogue: I did not really look into the tests beyond trying to make them pass. 😓 

My hope is to revisit these when we update the package.json handling.

raphinesse: Well `basePkgJson13` refers to a project fixture that might be unused now. I can't check it now though.

jpike88:
Should at least put a warning in the CLI! This is a huge change and was throwing me off for an hour while I was figuring out why plugins mysteriously were disappearing from config.xml.

dpogue:
@jpike88 Nothing should be disappearing from config.xml, we're not removing anything from the file. We're just not adding to it anymore, and instead consolidating everything in package.json.

jpike88:
I think a warning that's detecting existing mentions of plugins/engines in config.xml should be thrown, because tools like cordova-check-plugins rely on a rm/add process through Cordova and it can be disorienting to see things just disappearing from config.xml. Would make more sense to encourage users to clear it out, or just do a full clear out with a message saying why.

tfrijsewijk:
Help me out here.. I started a new cordova project, added some plugins, and now I want to build my project using PhoneGap Build. But PGB wants config.xml..

What to do? How will PGB know which plugins to restore if it doesn't know what to do with package.json?

janpio:
That sounds like a problem with Phonegap Build @tfrijsewijk, not with `cordova-lib`. Please talk to the Phonegap Build people on how to solve this problem.

tfrijsewijk:
You are right! But have you ever been to the Adobe Community forums? :-)

This is a pretty breaking change (hence the major increment). I came here through [Google](https://www.google.com/search?q=phonegap+package.json+config.xml) -> [Stackoverflow (1st hit)](https://stackoverflow.com/questions/46080642/cordova-7-config-xml-or-package-json) -> Second answer and I can imagine I'm not the only one.

So - pretty please - can you give me a hint on how to get this going for PhoneGap Build? I mean, PhoneGap Build uses cordova right? Is this is a matter of patience and PGB will support plugins from package.json ?

tfrijsewijk:
You are right! But have you ever been to the Adobe Community forums? :-)

This is a pretty breaking change (hence the major increment). I came here through [Google](https://www.google.com/search?q=phonegap+package.json+config.xml) -> [Stackoverflow (1st hit)](https://stackoverflow.com/questions/46080642/cordova-7-config-xml-or-package-json) -> Second answer and I can imagine I'm not the only one.

So - pretty please - can you give me a hint on how to get this going for PhoneGap Build? I mean, PhoneGap Build uses cordova right? Is this is a matter of patience and PGB will support plugins from package.json ?

janpio:
I literally have no idea, as I have never used Phonegap Build and don't know how it works. I think they use a forked version of Cordova behind the scenes, but usually lag behind. From rereading your comments here, I would try to manually add the plugins to `config.xml` (which is still valid for Cordova CLI, just not the default and main way) and see if this fixes stuff for PGB.

(If you want to keep discussing this, better open a new issue here instead of commenting on a merged pull request. We usually don't like support requests like this, but abusing an old PR is even worse and more confusing :monkey:)

ryanmtaylor:
I'd love to see the fact that config.xml plugins don't really affect anything (and hence aren't being added) documented literally anywhere – eg. https://cordova.apache.org/docs/en/latest/config_ref/#plugin

This is the only place I could find really any documentation of this fact.

janpio:
(Same for the fact that plugins come from `package.json` probably)

janpio:
Best open a new issue for this @ryanmtaylor - otherwise it will get lost here in this closed PR. `cordova-lib` is probably an ok location, as the change was done here.

raphinesse:
I'd argue that PRs and issues that aim to improve the documentation should be created in cordova-docs.

ryanmtaylor:
@janpio @raphinesse thank you created the feature request here: https://github.com/apache/cordova-docs/issues/1004

pjc2007:
As per mention above (linked issue), I had to manually add the `cordova-plugin-file` plugin into `config.xml` for it to work. This stumped me for a very long time (as I never thought to try this)

PR #752, opened 2019-03-22
Fix restoring plugins from package.json

dpogue:
### Platforms affected
all

### Motivation and Context

Some of the code removed in GH-750 had side effects on the rest of the plugin restoration procedure. Namely, despite wanting to migrate to package.json, we actually relied on config.xml for retrieving the plugin spec and variables. When we stopped syncing changes back to config.xml, those plugins stopped getting restored.

### Description

I've refactored this significantly to reduce complexity and make it easier to read and understand.

We look at package.json for plugins.
Then we look at config.xml. If a plugin exists in config.xml but not in package.json, we copy it to package.json.
Then we build up a list of the name/version/variables for each plugin, and pass them to the plugin installer in series.

### Testing

All tests pass.
It appears to work properly, having been tested in several of my own apps now, but I don't trust that the unit/e2e tests actually validate it, so I encourage further testing.

### Checklist

### DIFF
--- a/spec/cordova/restore-util.spec.js
+++ b/spec/cordova/restore-util.spec.js
@@ -129,7 +129,8 @@ describe('cordova/restore-util', () => {
             return o;
         }, {});
         if (Object.keys(specs).length > 0) {
-            expect(pkgJson.dependencies).toEqual(jasmine.objectContaining(specs));
+            let specs = Object.assign({}, pkgJson.dependencies, pkgJson.devDependencies);
+            expect(specs).toEqual(jasmine.objectContaining(specs));
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/cordova/util.js
```diff
@@ -151,12 +152,24 @@ function isCordova (dir) {
...
     if (!projectRoot) {
         throw new CordovaError('Current working directory is not a Cordova-based project.');
     }
+
+    return projectRoot;
+}
+
+// Cd to project root dir and return its path. Throw CordovaError if not in a Corodva project.
+function cdProjectRoot () {
+    const projectRoot = this.getProjectRoot();
```
raphinesse: I can't do a proper review right now, but I think the usage of `this` here does hurt more than it helps. I know this follows the style of the rest of `util` but I'd avoid it in new code.

spec/cordova/restore-util.spec.js
```diff
@@ -390,7 +356,7 @@ describe('cordova/restore-util', () => {
                 expectPluginsInPkgJson([{
                     name: 'cordova-plugin-camera',
                     spec: '^2.3.0',
-                    variables: { variable_1: 'value_1', variable_2: 'value_2', variable_3: 'value_3' }
+                    variables: { variable_1: 'value_1', variable_3: 'value_3' }
```
raphinesse: Did you check whether the changed tests duplicate any of the old ones after the change?

src/cordova/util.js
dpogue: I agree, but removing it causes a bunch of tests to fail (I suspect due to improper stubbing)

raphinesse: Bummer. I already had a feeling that this could be the reason. I agree that it's better not having to touch the tests then.

raphinesse:
Just FYI: I'm currently doing a thorough review of this PR.

raphinesse:
I've added some mostly cosmetic changes to the rewritten code. I think they are slight improvements. Feel free to disagree :smile:

The changes to the tests from this PR and from GH-750 turned out to be suboptimal to say the least. I've rewritten the tests for `restoreUtil.installPluginsFromConfigXML` to actually match the new logic.

We should really take a closer look at the other tests and test files touched by GH-750 and make sure that they still make sense with the new logic. Simply modifying bits and pieces until the tests stop failing just isn't enough. :sweat_smile: 

@dpogue If we agree on the changes in this PR, feel free to squash all my commits with yours.

PR #755, opened 2019-03-25
Script path must not be wrapped in quotes

### LINKED ISSUE #754: Version check broken on non-Windows-OS
# Bug Report

## Problem
Originally reported at https://github.com/apache/cordova/issues/52

No errors emitted during `cordova prepare`

Cordova throws an ENOENT error, most likely due to an invalid escaping strategy on a path.

## Information
```
Installing "cordova-plugin-camera" for android
Running command: /builds/fairmanager/mobile2/platforms/android/cordova/version 
Command finished with error code 0: /builds/fairmanager/mobile2/platforms/android/cordova/version 
Running command: "/builds/fairmanager/mobile2/platforms/android/cordova/version" 
Command finished with error code ENOENT: "/builds/fairmanager/mobile2/platforms/android/cordova/version" 
```

I assume that this is caused by unconditionally wrapping the path in quotes here: https://github.com/apache/cordova-lib/blob/master/src/plugman/install.js#L168

I assume this is a leftover from when the shell was used to check for the path existence. It has no benefit here in any scenario that I can imagine.

### Command or Code
`cordova prepare --verbose`

### Environment, Platform, Device
Linux

### Version information
Cordova@9.0.0

## Checklist
- [X] I searched for existing GitHub issues
- [X] I updated all Cordova tooling to most recent version
- [X] I included all the necessary information above

oliversalzburg:
### Platforms affected
All

### Motivation and Context
Fixes #754
Fixes apache/cordova#52

### Description
Paths must not be wrapped in quotes before passing them to `fs`. When invoking commands on the shell, escaping should be handled by the shell library.

### Testing

None. Opening PR here for CI.

### Checklist

- [X] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [X] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [X] I've updated the documentation if necessary

### DIFF
--- a/src/plugman/install.js
+++ b/src/plugman/install.js
@@ -165,7 +165,7 @@ function callEngineScripts (engines, project_dir) {
     return Promise.all(
         engines.map(function (engine) {
             // CB-5192; on Windows scriptSrc doesn't have file extension so we shouldn't check whether the script exists
-            var scriptPath = engine.scriptSrc ? '"' + engine.scriptSrc + '"' : null;
+            var scriptPath = engine.scriptSrc || null;
             if (scriptPath && (isWindows || fs.existsSync(engine.scriptSrc))) {
 
                 if (!isWindows) { // not required on Windows


brodybits:
I have to say that I am not super familiar with this code, but the change looks right to me. From some digging I found that the quotes were originally added in ae2ce7ac6325d1c6558365b4be82e5a8ff0983b9 back in 2014, with the following commit message:

> Refactoring of install & uninstall tests

brodybits:
I would like to get one more review before we merge this one. You may want to send a request to <dev@cordova.apache.org>.

oliversalzburg:
@purplecabbage We're on line 168, `engine.scriptSrc` exists and points to a path. Thus, `scriptPath` will contain the original path wrapped in quotes.
On line 169, `scriptPath` is truthy, `isWindows` is `false`, but `fs.existsSync` is still evaluated. This receives the quoted path and throws.

purplecabbage:
Okay,  I was missing the call in line: 174 ... my mistake.

This must be an issue in the way superspawn.spawn calls vs child_process.exec since this code has been functioning correctly for the last 5 years.

brodybits:
I think the explanation is that in apache/cordova-common#50 we changed our "superspawn" implementation to use [`cross-spawn`](npmjs.com/package/cross-spawn). I think the general purpose was to improve maintainability.

When looking at the code around the proposed change, I am wondering if we should remove any more Windows-specific code now?

raphinesse:
Sorry for breaking stuff 🙄

I think we need an integration test that actually fails before this change.

oliversalzburg:
@brodybits In that file I only see the `chmod` attempt and that should still not be attempted on Windows probably. Windows doesn't have the same concept of the executable flag.

@raphinesse Sounds difficult, as exceptions thrown inside that block don't seem to surface or cause the build to end. We would first need proper error handling for exceptions thrown inside that block, then fail the build for them, then add a test, then merge this fix. I would hate to see a delay like this for this fix. But I can also see that this is currently not an ideal situation.

Or am I missing an easier approach?

raphinesse:
@oliversalzburg The existing tests quite often spy on `events` to assert that expected events are fired. That should be applicable to this case too. Couldn't we also just assert that we get the actual version of a platform instead of `null`? (I can't check myself in more detail since I'm on mobile)

I agree with you that the current error-ignoring strategy is quite odd, however. But that's definitely out of scope.

If the behavior turns out to be too difficult to test after all, we can of course ditch the test.

oliversalzburg:
The problem I'm having here is that this is a non-Windows-specific issue and I work on Windows. On Windows it is not even an issue if you pass the quoted path to `superspawn.spawn`, so I can't even repro the error case locally.

raphinesse:
Understood. I _might_ be able to take a look at it tonight or tomorrow. Will post any update here.

benmod:
Hello, in which version of cordova-lib is this patch available? I am using 9.0.2 and getting the same error.

raphinesse:
Are you sure you are using the right package @benmod? The latest release of `cordova-lib` is 9.0.1 and it contains this fix.

PR #758, opened 2019-03-29
Refactor plugin_fetch.spec

raphinesse:
### Motivation and Context
The logic in this test file was not very transparent and it relied on outdated fixtures.

### Description
I refactored the test to be more readable, robust and self contained. Finally I removed the reliance on the outdated fixtures and moved it to `spec` since it is actually a unit test.

### Testing
`npm t`


raphinesse:
Merged, as this really isn't worth a review.

Issue #757, opened 2019-03-27
Use up-to-date fixtures in tests


raphinesse:
Some of our automated tests use project fixtures. A lot of those are outdated. For example, some of those tests use project fixtures that have outdated `cordova-android` versions with bundled dependencies installed. Running our tests against these outdated fixtures does not reflect real-world usage.

Instead, a project as created by the currently running version of Cordova should be used.

CC @erisu

PR #764, opened 2019-04-13
Expressive pkg json.spec

raphinesse:
This factors out some test helpers common to `spec/cordova/restore-util.spec.js` and `integration-tests/pkgJson.spec.js` to the end of using them to create test projects with certain properties in `pkgJson.spec` instead of asserting the properties of various fixture variants. Finally, the unused fixtures have been deleted.


spec/project-test-helpers.js
```diff
@@ -0,0 +1,90 @@
...
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+const path = require('path');
+const fs = require('fs-extra');
+const { ConfigParser } = require('cordova-common');
+
+const fixturesPath = path.join(__dirname, 'cordova/fixtures');
```
erisu:
```suggestion
const fixturesPath = path.join(__dirname, 'cordova', 'fixtures');
```

raphinesse: Actually, I usually change the for you suggested to the one I used. I like it better because I find it more legible and it is shorter. The result is the same in either case:
```js
require('path').win32.join('foo', 'bar/baz')
// => 'foo\\bar\\baz'
```

raphinesse:
Related to #757

PR #765, opened 2019-04-13
Remove compatibility code for plugins added by cordova@<5.4.0

raphinesse:
This remove s code that is commented with:
> CB-9987 We need to reinstall the plugins for the platform it they were added by cordova@<5.4.0

I'm unsure as to the kind of release this should go in. Since we probably officially removed support for cordova@<5.4.0, this should be non-breaking. If you want to be very strict, this would be breaking. Opinions?


raphinesse:
Oops, forgot the tests

raphinesse:
I'll play it safe and mark this as a breaking change.

PR #767, opened 2019-04-13
Simpler and better cordova/util.getPlatformApiFunction

raphinesse:
### Motivation and Context
The logic in `cordova/util.getPlatformApiFunction` was a tad to complex and had a few problems because of that.

During testing I frequently had the situation that an error occurred during loading of a platform API module, mostly because of a missing dependency. This function caught these errors and insted threw an error claiming `'Your ' + platform + ' platform does not have Api.js'` which was plainly wrong and confusing.

On top of that came a dead code branch and IMHO unnecessary checks. Enough reason for a rewrite

AFAICT this method is not publicly exposed, so this should be a minor change at most.

### Description
- The happy path is unchanged as are the situations in which errors are thrown
- The second argument is now unused and could be removed at the call sites in a following or this PR
- The other notable changes all pertain to logged messages:
  - No more warning that file should be called `Api.js`
  - No more special error message if platform is a custom one (`!platforms[platform]`)
  - Improved error message if module doesn't implement the expected interface

### Testing
The existing tests pass.

### DIFF
--- a/spec/cordova/platforms/platforms.spec.js
+++ b/spec/cordova/platforms/platforms.spec.js
@@ -62,7 +62,7 @@ describe('platforms/platforms', () => {
             expect(platformApi).toBeDefined();
             expect(platformApi.platform).toEqual('windows');
             expect(events.emit.calls.count()).toEqual(1);
-            expect(events.emit.calls.argsFor(0)[1]).toMatch('PlatformApi successfully found');
+            expect(events.emit.calls.argsFor(0)[1]).toMatch('Platform API successfully found in:');
             expect(util.convertToRealPathSafe.calls.count()).toEqual(1);
             expect(util.isCordova.calls.count()).toEqual(0);
             expect(util.requireNoCache.calls.count()).toEqual(1);
--- a/spec/cordova/platforms/platforms.spec.js
+++ b/spec/cordova/platforms/platforms.spec.js
@@ -62,7 +62,7 @@ describe('platforms/platforms', () => {
             expect(platformApi).toBeDefined();
             expect(platformApi.platform).toEqual('windows');
             expect(events.emit.calls.count()).toEqual(1);
-            expect(events.emit.calls.argsFor(0)[1]).toEqual('PlatformApi successfully found for platform windows');
+            expect(events.emit.calls.argsFor(0)[1]).toMatch('PlatformApi successfully found');
             expect(util.convertToRealPathSafe.calls.count()).toEqual(1);
             expect(util.isCordova.calls.count()).toEqual(0);
             expect(util.requireNoCache.calls.count()).toEqual(1);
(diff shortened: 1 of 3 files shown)


src/cordova/util.js
```diff
@@ -351,43 +351,23 @@ function isSymbolicLink (dir) {
...
-        }
+        PlatformApi = exports.requireNoCache(dir);
     } catch (err) {
-        // Emit the err, someone might care ...
-        events.emit('warn', 'Unable to load PlatformApi from platform. ' + err);
-        if (!platforms[platform]) {
-            events.emit('error', 'The platform "' + platform + '" does not appear to be a valid cordova platform. It is missing API.js. ' + platform + ' not supported.');
-        }
+        // Module not found or threw error during loading
+        err.message = `Unable to load Platform API from ${dir}:\n` + err.message;
```
erisu:
```suggestion
        err.message = `Unable to load Platform API from ${dir}:\n${err.message}`;
```
Minor stuff, not necessary and wont block.
- Since you also converted this string to an interpolate strings, `err.message` could also be added and not concatenated.

src/cordova/util.js
```diff
@@ -351,43 +351,23 @@ function isSymbolicLink (dir) {
...
+        throw err;
     }
 
-    if (!PlatformApi) {
-        throw new Error('Your ' + platform + ' platform does not have Api.js');
+    // Module doesn't implement the expected interface
+    if (!PlatformApi || !PlatformApi.createPlatform) {
+        throw new Error(`The package at ${dir} does not appear to implement the Cordova Platform API.`);
```
erisu:
```suggestion
        throw new Error(`The package at "${dir}" does not appear to implement the Cordova Platform API.`);
```
I cant remember but I thought we typically use quotes around pathings. Again nothing important.

src/cordova/util.js
```diff
@@ -351,43 +351,23 @@ function isSymbolicLink (dir) {
...
     }
 
+    events.emit('verbose', 'PlatformApi successfully found in ' + dir);
```
erisu:
```suggestion
    events.emit('verbose', 'Platform API successfully found in: ' + dir);
```
??

spec/cordova/platforms/platforms.spec.js
```diff
@@ -62,7 +62,7 @@ describe('platforms/platforms', () => {
             expect(platformApi).toBeDefined();
             expect(platformApi.platform).toEqual('windows');
             expect(events.emit.calls.count()).toEqual(1);
-            expect(events.emit.calls.argsFor(0)[1]).toEqual('PlatformApi successfully found for platform windows');
+            expect(events.emit.calls.argsFor(0)[1]).toMatch('PlatformApi successfully found');
```
erisu:
```suggestion
            expect(events.emit.calls.argsFor(0)[1]).toMatch('Platform API successfully found in:');
```
If you accept the emit change below, I went ahead and updated this as well..

spec/cordova/util.spec.js
```diff
@@ -220,7 +220,7 @@ describe('util module', function () {
                 var specPlugDir = __dirname.replace('spec-cordova', 'spec-plugman');
                 util.getPlatformApiFunction((path.join(specPlugDir, 'fixtures', 'projects', 'platformApi', 'platforms', 'windows', 'cordova', 'Api.js')), 'windows');
                 expect(events.emit.calls.count()).toBe(1);
-                expect(events.emit.calls.argsFor(0)[1]).toBe('PlatformApi successfully found for platform windows');
+                expect(events.emit.calls.argsFor(0)[1]).toMatch('PlatformApi successfully found');
```
erisu:
```suggestion
                expect(events.emit.calls.argsFor(0)[1]).toMatch('Platform API successfully found in:');
```

PR #768, opened 2019-04-13
Proper async code in plugman.createpackagejson

raphinesse:
### Motivation and Context

The existing code returned an instantly resolved promise despite the fact that it was doing async operations.

### Description

Promisified the async function and properly returned the resulting promise.

### Testing

Manually tested this using the plugman CLI. Works like it should.

### DIFF
--- a/src/plugman/createpackagejson.js
+++ b/src/plugman/createpackagejson.js
@@ -21,7 +21,8 @@ var fs = require('fs-extra');
 var path = require('path');
 var PluginInfo = require('cordova-common').PluginInfo;
 var events = require('cordova-common').events;
-var init = require('init-package-json');
+const pify = require('pify');
+const initPkgJson = pify(require('init-package-json'));
 
 // returns a promise
 function createPackageJson (plugin_path) {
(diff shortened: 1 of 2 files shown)


src/plugman/createpackagejson.js
```diff
@@ -39,18 +40,17 @@ function createPackageJson (plugin_path) {
         platforms: pluginInfo.getPlatformsArray()
     };
 
-    fs.writeFile(path.join(__dirname, 'defaults.json'), JSON.stringify(defaults), 'utf8', function (err) {
-        if (err) throw err;
-        events.emit('verbose', 'defaults.json created from plugin.xml');
-        var initFile = require.resolve('./init-defaults');
-        var dir = process.cwd();
+    return fs.writeFile(path.join(__dirname, 'defaults.json'), JSON.stringify(defaults), 'utf8')
```
dpogue: `fs.writeFile` needs to be wrapped to return a promise, doesn't it?

raphinesse: It's the one from `fs-extra`. That's already promisified.

PR #772, opened 2019-04-13
Replace underscore with modern JS

### LINKED ISSUE #643: Get rid of underscore, maybe replace it with lodash
Open for opinions and alternatives.

raphinesse:
Because we can. The only logical difference should be that `Object.assign` takes _own_ enumerable props, while `_.extend` enumerable.

Resolves #643

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,8 +31,7 @@
     "md5-file": "^4.0.0",
     "read-chunk": "^3.1.0",
     "semver": "^5.6.0",
-    "shebang-command": "^1.2.0",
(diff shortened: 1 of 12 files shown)


erisu:
Nice! 👍

As part of the work I been doing on the draft branched for all tooling repos, I been looking to deprecate the `underscore` dependency.

I forgot which repo exactly I was looking at, at the time, but good to see and know that this PR already existed.

When I started the draft branches I forgot to check all PRs :D.

PR #774, opened 2019-04-15
Make src/plugman/init-defaults.js lintable

raphinesse:
### Motivation and Context

`src/plugman/init-defaults.js` was not lintable since it was using the reserved word `package`.

### Description

Use the fact that the PromZard context is also provided as `this` to assign the `package` variable to `pkg` instead.

### Testing

Manual testing with Plugman CLI.


raphinesse:
@dpogue, faster than the speed of CI :grinning: 
Thanks!

PR #775, opened 2019-04-15
Cleanup plugman.createPackageJson

raphinesse:
### Description
This PR removes dead code from `plugman/createpackagejson` and `plugman/init-defaults` and simplifies the passing of arguments between the two.  There should be absolutely no change in behavior.

For further details please see the individual commits.

### Testing
Manual testing using Plugman CLI

### DIFF
--- a/src/plugman/init-defaults.js
+++ b/src/plugman/init-defaults.js
@@ -17,11 +17,7 @@
     under the License.
  **/
 
-/* global config */
-/* global dirname */
-/* global basename */
-/* global yes */
-/* global prompt */
+/* global config, dirname, basename, yes, prompt */
 // PromZard file that is used by createpackagejson and init-package-json module
 
 // The PromZard context is also provided via this. Make use of this to avoid
--- a/src/plugman/init-defaults.js
+++ b/src/plugman/init-defaults.js
@@ -17,6 +17,7 @@
     under the License.
  **/
 
+/* global config */
 /* global dirname */
 /* global basename */
 /* global yes */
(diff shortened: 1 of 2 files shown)


src/plugman/init-defaults.js
```diff
@@ -47,7 +42,7 @@ function readDeps (test) {
                     if (er) return next();
                     try { p = JSON.parse(p); } catch (e) { return next(); }
                     if (!p.version) return next();
-                    deps[d] = config.get('save-exact') ? p.version : config.get('save-prefix') + p.version;
+                    deps[d] = undefined + p.version;
```
dpogue: This seems wrong... especially if it's a string, it looks like it would turn into `"undefined1.0.0"` instead of `"^1.0.0"` or `"~1.0.0"`

raphinesse: You are right. This is exactly what happens. And it's exactly what happened before the changes in this PR. As I wrote above, this does not change existing behavior in any way.

I could include a fix so that it prepends '^' instead. I actually had this in a previous version of this PR. But my current plan is to follow this PR up with one that removes the dependency feature altogether. It's completely useless. Go ahead, create a plugin with `plugman create`, copy over `node_modules` from say `cordova-lib` and then run `plugman createpackagejson` in the plugin directory (can't run it outside, since `readDeps` is broken that way too). Then see for yourself if you think that feature is worth fixing :sweat_smile: 

Frankly, I think this is so broken, removing it could be considered a bug fix.

dpogue: :scream: 

> Frankly, I think this is so broken, removing it could be considered a bug fix.

Sounds about right

raphinesse: So did you give it a try? 😅

PR #783, opened 2019-06-11
nyc@14 update in devDependencies

brodybits:
### Platforms affected

All

### Motivation and Context

This update resolves the `npm audit` warning at this time.

### Description

see title: nyc@14 update in `devDependencies`

### Testing

- [x] `npm run cover` succeeds locally on Node.js versions 6, 8, and 10
- [x] verified that Travis CI succeeds
- [x] verified that AppVeyor CI succeeds _- on Node.js versions 6 & 8, TBD still queued on Node.js 10_

### Checklist

- [x] I've run the tests to see all new and existing tests pass
- ~~I added automated test coverage as appropriate for this change~~
- ~~Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)~~
- ~~If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))~~
- ~~I've updated the documentation if necessary~~

PR #785, opened 2019-06-13
Revert "nyc@14 update in devDependencies"

brodybits:
Reverts apache/cordova-lib#783 - I think we should revert PR #783 since it fails on AppVeyor on Node.js version 10.

I think it would be cleanest to revert the change, which is a single line in `devDependencies`, then investigate *why* we get the single test failure on AppVeyor on Node.js 10.

An alternative could be to disable the single e2e test on Windows until we get a chance to investigate the issue with the test case. I would probably not favor this alternative.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "eslint-plugin-standard": "^4.0.0",
     "jasmine": "^3.3.1",
     "jasmine-spec-reporter": "^4.2.1",
-    "nyc": "^14.1.1",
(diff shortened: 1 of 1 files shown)


brodybits:
Reverting PR #783 does not seem to resolve the failure on AppVeyor Node.js version 10, closing now. I think we need to disable the failing e2e test on Windows until we can resolve this failure.

PR #786, opened 2019-06-13
Quick workaround for e2e failure on AppVeyor CI

brodybits:
This quick workaround is proposed in order to give us a green build as quickly as possible.

The build failure appeared when we reviewed and merged PR #783 but did *not* go away when I tried reverting it in PR #785.

### DIFF
--- a/integration-tests/pkgJson.spec.js
+++ b/integration-tests/pkgJson.spec.js
@@ -255,6 +255,9 @@ describe('pkgJson', function () {
 
         // Test#025: has a pkg.json. Checks if local path is added to pkg.json for platform and plugin add.
         it('Test#025 : if you add a platform/plugin with local path, pkg.json gets updated', function () {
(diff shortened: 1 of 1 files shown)


janpio:
Is there an issue for that test failure? Do we understand why this is happening?

brodybits:
Closing for now. I would like to try fs-extra@8 update before continuing with this proposal (<https://github.com/apache/cordova-lib/issues/787#issuecomment-501877936>).

This failure is now tracked in #787. At this point I do not understand why, do not have much time to investigate right now.

brodybits:
As I said in <https://github.com/apache/cordova-lib/issues/787#issuecomment-501943751>, fs-extra@8 does not seem to resolve the build issue. I am reopening this proposal to fix the build asap. A quick review would be much appreciated.

brodybits:
Can I get a quick review? I think this is really needed to fix the build, already raised #787 to track the test that fails on AppVeyor CI.

PR #799, opened 2019-10-10
chore: drop node 6 and 8 support

raphinesse:
Dropping node 6 and 8 support will allow us to not worry about dubious failures that only happen in these versions and to use new features like `async/await`.


raphinesse:
@brodybits My plan exactly :+1:

PR #804, opened 2019-10-19
Fix cordova/emulate.spec

raphinesse:
In some tests of `cordova/emulate.spec` it did call `cordova.run` or `cordova.compile` instead of `cordova.emulate`. This PR fixes this. In one case the expectations had to be adapted.

Additionally, this removes filtering of some outdated platform that does not exist anymore.

### DIFF
--- a/spec/cordova/emulate.spec.js
+++ b/spec/cordova/emulate.spec.js
@@ -21,7 +21,7 @@ var platforms = require('../../src/platforms/platforms');
 var HooksRunner = require('../../src/hooks/HooksRunner');
 var util = require('../../src/cordova/util');
 
-var supported_platforms = Object.keys(platforms).filter(function (p) { return p !== 'www'; });
+var supported_platforms = Object.keys(platforms);
 
 describe('emulate command', function () {
     var is_cordova;
--- a/spec/cordova/emulate.spec.js
+++ b/spec/cordova/emulate.spec.js
@@ -46,7 +46,7 @@ describe('emulate command', function () {
     describe('failure', function () {
         it('Test 001 : should not run inside a Cordova-based project with no added platforms by calling util.listPlatforms', function () {
             list_platforms.and.returnValue([]);
-            return cordova.compile()
+            return cordova.emulate()
                 .then(function () {
                     fail('Expected promise to be rejected');
                 }, function (err) {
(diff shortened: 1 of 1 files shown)


spec/cordova/emulate.spec.js
```diff
@@ -99,11 +99,15 @@ describe('emulate command', function () {
                 platformApi.build = originalBuildSpy;
             });
             it('Test 006 : should leave parameters unchanged', function () {
-                return cordova.run({ platforms: ['blackberry10'], options: { password: '1q1q' } })
+                const baseOptions = { password: '1q1q', device: false, emulator: true };
+                const expectedRunOptions = Object.assign({ nobuild: true }, baseOptions);
+                const expectedBuildOptions = Object.assign({ couldBeModified: 'insideBuild' }, baseOptions);
+
+                return cordova.emulate({ platforms: ['blackberry10'], options: { password: '1q1q' } })
```
erisu: Do we still want to keep the deprecated  `blackberry10` reference or should we change it to a non-deprecated platform?

spec/cordova/emulate.spec.js
```diff
@@ -99,11 +99,15 @@ describe('emulate command', function () {
...
                     .then(function () {
-                        expect(prepare_spy).toHaveBeenCalledWith({ platforms: [ 'blackberry10' ], options: { password: '1q1q', 'couldBeModified': 'insideBuild' }, verbose: false });
-                        expect(platformApi.build).toHaveBeenCalledWith({ password: '1q1q', 'couldBeModified': 'insideBuild' });
-                        expect(platformApi.run).toHaveBeenCalledWith({ password: '1q1q', nobuild: true });
+                        expect(prepare_spy).toHaveBeenCalledWith({ platforms: [ 'blackberry10' ], options: expectedBuildOptions, verbose: false });
```
erisu: Same here

spec/cordova/emulate.spec.js
raphinesse: I don't think it matters at all for these tests. If we were to change it, I'd use a platform that does not exist at all. A lot of tests use `atari` as a platform for that purpose.

erisu: That would work :D

Was just a bit confusing cause I knew it was a valid platform at one point.

Anyways 👍

PR #809, opened 2019-10-29
chore: improve npm ignore list

erisu:
### Motivation and Context

Decrease production package size

### Description

**BEFORE**
npm notice package size:  105.7 kB
npm notice unpacked size: 487.6 kB
npm notice total files:   86

**AFTER**
npm notice package size:  88.3 kB
npm notice unpacked size: 394.7 kB
npm notice total files:   66

### Testing

- `npm t`

### Checklist

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,6 @@
-spec-cordova
-spec-plugman
+.*
+appveyor.yml
 spec
+integration-tests
 coverage
-src/plugman/defaults.json
 temp


timbru31 (dismissed):
We should probably add the two .eslint, .travis, .ratignore, too. IMO not required for the end user

.npmignore
```diff
@@ -1,6 +1,8 @@
+.*
+appveyor.yml
+spec
 spec-cordova
 spec-plugman
-spec
+integration-tests
 coverage
-src/plugman/defaults.json
```
raphinesse: That file is generated during testing IIRC. Can you confirm? Is it excluded anyway due to being ignored by git?

.npmignore
```diff
@@ -1,6 +1,8 @@
+.*
+appveyor.yml
+spec
 spec-cordova
 spec-plugman
```
raphinesse: I believe those two are ancient history and can be removed. No blocker though.

raphinesse:
> We should probably add the two .eslint, .travis, .ratignore, too. IMO not required for the end user

Note the first line. It ignores all dotfiles.

timbru31:
> > We should probably add the two .eslint, .travis, .ratignore, too. IMO not required for the end user
> 
> @timbru31 Note the first line. It ignores all dotfiles.

Oh indeed, I've missed that in the diff.

.npmignore
erisu: It will not be excluded because it is in the `.gitignore` file in this setup. 

From npm docs
```
Use a .npmignore file to keep stuff out of your package. If there’s no .npmignore file, but there is a .gitignore file, then npm will ignore the stuff matched by the .gitignore file. If you want to include something that is excluded by your .gitignore file, you can create an empty .npmignore file to override it. Like git, npm looks for .npmignore and .gitignore files in all subdirectories of your package, not only the root directory.
```

So from how I am reading this, once the `.npmignore` file is created to exclude items that are not excluded by `.gitignore`, npm will use only the `.npmignore` file.

.npmignore
erisu: I wasn't going risk it but I couldn't find anything on it, nor have I seen it. I think I will got ahead and remove it.

.npmignore
raphinesse: Alright, I checked and `defaults.json` is not created anymore since #775. Thus, we should not need it in any of the ignore files anymore.

raphinesse (approved):

.npmignore
erisu:
```
spec-cordova
spec-plugman
```

has been dropped.

PR #680, opened 2018-09-06
CB-12774: Support scoped plugins

oliversalzburg:
### Platforms affected
All

### What does this PR do?
See #602 for details.

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [x] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [x] Added automated test coverage as appropriate for this change.

PR #818, opened 2019-11-02
Extend and improve plugin tests in preparation of supporting scoped plugins

raphinesse:
### Motivation and Context

This is a revised version of all passing tests that were included in #602 or #680 respectively. I pulled them out of the original PR to reduce its volume and because they are valid without the other changes of the PR.

### Description

- Extend tests for addHelper.installPluginsForNewPlatform
- Simplify and extend unit tests for cordova/util.findPlugins
- Add passing unit tests for plugman/util/metadata
- Test plugin/add.getFetchVersion w/ scoped plugins
- Less strict path comparison in plugman_fetch.spec

### DIFF
--- a/spec/cordova/plugin/add.spec.js
+++ b/spec/cordova/plugin/add.spec.js
@@ -381,7 +381,7 @@ describe('cordova/plugin/add', function () {
                     });
             });
             it('should retrieve installed plugins and installed platforms version and feed that information into determinePluginVersionToFetch', function () {
-                plugin_util.getInstalledPlugins.and.returnValue([{ 'id': 'cordova-plugin-camera', 'version': '2.0.0' }]);
+                plugin_util.getInstalledPlugins.and.returnValue([{ 'id': '@cordova/plugin-thinger', 'version': '2.0.0' }]);
                 cordova_util.getInstalledPlatformsWithVersions.and.returnValue(Promise.resolve({ 'android': '6.0.0' }));
                 pluginInfo.engines = {};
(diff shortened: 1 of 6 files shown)


spec/cordova/plugin/add.spec.js
```diff
@@ -381,15 +381,15 @@ describe('cordova/plugin/add', function () {
                     });
             });
             it('should retrieve installed plugins and installed platforms version and feed that information into determinePluginVersionToFetch', function () {
-                plugin_util.getInstalledPlugins.and.returnValue([{ 'id': 'cordova-plugin-camera', 'version': '2.0.0' }]);
+                plugin_util.getInstalledPlugins.and.returnValue([{ 'id': '@cordova/plugin-thinger', 'version': '2.0.0' }]);
```
erisu: Do we need to continue to test a plugin that is not scoped?

raphinesse: I also thought about that. I actually had it that way before. I didn't particularly like it, because it basically duplicated the test. I thought if a scoped plugin would work, then a normal one should work too. But that implication is not necessarily true of course 🤷‍♂

breautek: I agree with @raphinesse here, while in theory if scope works, then non-scoped plugins should work as well, but they are two different formats and I don't think it hurts testing both formats for peace of mind.

raphinesse: I took another look at the tests and what was being tested and I don't really see the value of duplicating the test here.

PR #821, opened 2019-11-10
Support scoped plugins

raphinesse:
### Motivation and Context

See #602 for details. Closes #680.

### Description

This contains the revised changes from #680 minus the changes from #818 and apache/cordova-common#94.

### Testing

Only unit testing so far

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "node": ">=10.0.0"
   },
   "dependencies": {
-    "cordova-common": "^3.1.0",
+    "cordova-common": "4.0.0-nightly.2019.11.8.40cb6cee",
     "cordova-create": "^2.0.0",
     "cordova-fetch": "^2.0.0",
     "cordova-serve": "^3.0.0",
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -17,7 +17,7 @@
     "node": ">=10.0.0"
   },
   "dependencies": {
-    "cordova-common": "^3.1.0",
+    "cordova-common": "4.0.0-nightly.2019.11.8.40cb6cee",
```
erisu: If I am correct, you can set the version to a tag (`nightly`) so it will always fetch the latest nightly. This means we wouldn't need to always update the `package.json` as we continue to add changes to common.

```suggestion
    "cordova-common": "nightly",
```

The releaser would need to remember to change this to an official released version before release.

raphinesse: I actually deliberately chose to use a fixed nightly to keep our CI tests stable. I think it's actually good having to manually bump the dependency version to get new features. That explicitly shows the dependence on a new feature.

PR #833, opened 2020-05-09
breaking: upgrade cordova dependencies for next major

erisu:

- Update Cordova dependencies to start using the next major releases.
- Use dependencies that fixed to the issues related to the combination of Node 12x+ and Rewire.
- Use the next major's patch nightlies (e.g: cordova-common) which contains bug fixes. This is to continue validating the fixes in the dependencies itself and identify if there are other issues to be resolved before pushing the next patch release for those dependencies.

### Tasks
- [x] Update Cordova related dependencies
- [x] Fixed Unit Tests
- [ ] Fixed End-to-End Tests

### Testing

- `npm t`

### Checklist

- [x] I updated automated test coverage as appropriate for this change

### DIFF
--- a/src/platforms/platformsConfig.json
+++ b/src/platforms/platformsConfig.json
@@ -2,7 +2,7 @@
     "ios": {
         "hostos": ["darwin"],
         "url": "https://github.com/apache/cordova-ios.git",
-        "version": "^5.0.0",
+        "version": "^6.0.0-nightly.2020.5.11.e65d685c",
         "deprecated": false
     },
     "osx": {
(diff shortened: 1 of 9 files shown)


src/platforms/platformsConfig.json
```diff
@@ -2,7 +2,7 @@
     "ios": {
         "hostos": ["darwin"],
         "url": "https://github.com/apache/cordova-ios.git",
-        "version": "^5.0.0",
+        "version": "^6.0.0-nightly.2020.5.11.e65d685c",
```
dpogue, line 5: This should just be `^6.0.0`, right? Not nightly?

erisu: Ah yes, that can be 6 now.

PR #856, opened 2020-09-24
fix(cordova/util): version detection for legacy platforms

raphinesse:
### Motivation and Context

The new mechanism introduced in #815 relied on all platforms exporting their version in their version script. This is not the case for the browser platform though. Thus we need to revert to spawning an external process for legacy platforms unfortunately.

### Description

For platforms that do not expose `Api.version`, synchronously call the version script instead.

### Testing

I manually tested the new version with the current browser platform and `cordova info` output (which was broken for `browser` because of the bug fixed here)

### DIFF
--- a/spec/cordova/util.spec.js
+++ b/spec/cordova/util.spec.js
@@ -107,6 +107,19 @@ describe('util module', function () {
                 .then(versions => expect(versions[PLATFORM]).toBe('9.0.0'));
         });
     });
+    describe('getPlatformVersion method', () => {
+        it('should get the version from a legacy platform', () => {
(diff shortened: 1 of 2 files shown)


timbru31 (approved):

breautek (approved):

raphinesse:
Thanks for the reviews. I pushed a new version it
- includes a test to make sure it works on Windows as well
- spawns `node version` instead of `version` so as to not have to rely on `version.bat` on Windows

I hope the approvals still stand with these changes :)

breautek (approved):

PR #860, opened 2020-10-28
feat(cordova/util): support requiring platform API from node_modules

dpogue:
### Platforms affected
all

### Motivation and Context

We are looking at the possibility of having most of the platform tooling code run out of `node_modules` instead of being copied into the `platforms` folder. This allows loading the Platform API from either `platforms` or `node_modules`.

### Description

Try loading the Platform API from the `platforms` folder, but failing that, require the main file from the platform package.json in `node_modules`.

### Testing

- Added unit test
- Passes existing tests

### Checklist

### DIFF
--- a/src/cordova/util.js
+++ b/src/cordova/util.js
@@ -304,6 +304,8 @@ function isDirectory (dir) {
 function getPlatformApiFunction (dir, platform) {
     let PlatformApi;
     try {
+        // First try to load the platform API from the platform project
+        // This is necessary to support older platform API versions
         PlatformApi = exports.requireNoCache(dir);
     } catch (err) {
         try {
@@ -312,6 +314,7 @@ function getPlatformApiFunction (dir, platform) {
                 cdvPlatform = `cordova-${platform}`;
             }
 
+            // Load the platform API directly from node_modules
             PlatformApi = exports.requireNoCache(cdvPlatform);
         } catch (err) {
             // Module not found or threw error during loading
--- a/src/cordova/util.js
+++ b/src/cordova/util.js
@@ -307,15 +307,19 @@ function getPlatformApiFunction (dir, platform) {
         // First try to load the platform API from the platform project
         // This is necessary to support older platform API versions
         PlatformApi = exports.requireNoCache(dir);
-    } catch (err) {
+    } catch (loadFromDirError) {
+        events.emit('verbose', `Unable to load Platform API from ${dir}:`);
+        events.emit('verbose', CordovaError.fullStack(loadFromDirError));
+
+        const cdvPlatform = platform.replace(/^(?:cordova-)?/, 'cordova-');
         try {
             // Load the platform API directly from node_modules
-            const cdvPlatform = platform.replace(/^(?:cordova-)?/, 'cordova-');
             PlatformApi = require(cdvPlatform);
-        } catch (err) {
-            // Module not found or threw error during loading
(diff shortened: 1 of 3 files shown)


src/cordova/util.js
```diff
@@ -301,14 +301,23 @@ function isDirectory (dir) {
 
 // Returns the API of the platform contained in `dir`.
 // Potential errors : module isn't found, can't load or doesn't implement the expected interface.
-function getPlatformApiFunction (dir) {
+function getPlatformApiFunction (dir, platform) {
```
dpogue, line 304: Apparently we were already passing in the platform name as a second parameter, but not using it 🤷🏼‍♂️

src/cordova/util.js
```diff
@@ -301,14 +301,23 @@ function isDirectory (dir) {
...
     let PlatformApi;
     try {
         PlatformApi = exports.requireNoCache(dir);
```
raphinesse, line 307:
```suggestion
        // First try to load the platform API from the platform project
        // This is necessary to support older platform API versions
        PlatformApi = exports.requireNoCache(dir);
```

src/cordova/util.js
```diff
@@ -301,14 +301,23 @@ function isDirectory (dir) {
...
-        // Module not found or threw error during loading
-        err.message = `Unable to load Platform API from ${dir}:\n${err.message}`;
-        throw err;
+        try {
+            let cdvPlatform = platform;
+            if (!platform.startsWith('cordova-')) {
+                cdvPlatform = `cordova-${platform}`;
+            }
+
+            PlatformApi = exports.requireNoCache(cdvPlatform);
```
raphinesse, line 315:
```suggestion
            // Load the platform API directly from node_modules
            PlatformApi = exports.requireNoCache(cdvPlatform);
```

Do we need `requireNoCache` here? I guess not. But I'm not entirely sure.

src/cordova/util.js
```diff
@@ -301,14 +301,23 @@ function isDirectory (dir) {
...
-function getPlatformApiFunction (dir) {
+function getPlatformApiFunction (dir, platform) {
     let PlatformApi;
     try {
         PlatformApi = exports.requireNoCache(dir);
     } catch (err) {
-        // Module not found or threw error during loading
-        err.message = `Unable to load Platform API from ${dir}:\n${err.message}`;
-        throw err;
+        try {
```
raphinesse, line 309: Maybe only retry if `err.code === 'MODULE_NOT_FOUND'`? Not sure what would be better here.

raphinesse (approved):
LGTM. Left a few comments for good measure anyway :wink:

src/cordova/util.js
```diff
@@ -301,14 +301,26 @@ function isDirectory (dir) {
...
+            let cdvPlatform = platform;
+            if (!platform.startsWith('cordova-')) {
+                cdvPlatform = `cordova-${platform}`;
+            }
```
raphinesse, line 315: Different way to do the same thing:
```suggestion
            const cdvPlatform = platform.replace(/^(?:cordova-)?/, 'cordova-');
```

raphinesse (approved):
LGTM

raphinesse (commented):
Well, I think we should add a unit test before we land this

raphinesse (approved):

erisu (approved):
LGTM

PR #870, opened 2021-06-06
test(plugin.spec): fix version change in test fixture

raphinesse:
Fixes a test that failed due to an updated dependency ([example test log](https://travis-ci.com/github/apache/cordova-lib/jobs/511151470#L913))


PieterVanPoyer (approved):
Seems to be very good in my opinion.

PR #874, opened 2021-06-17
fix: Platforms restored from both dev and normal dependencies.

### LINKED ISSUE #866: Wrong platform version is detected when dependencies and devDependencies are both present
# Bug Report

## Problem
When dependencies and devDependencies are both present, running `cordova prepare` installs the wrong platform version.

`Using cordova-fetch for cordova-electron@2.0.0` is printed

`Using cordova-fetch for cordova-electron@1.0.0` is printed

### Environment, Platform, Device
the relevant part of my package.json:

```json
{
  "devDependencies": {
    "cordova-electron": "2.0.0"
  },
  "dependencies": {
    "cordova-plugin-googleplus": "8.5.1"
  },
  "cordova": {
    "plugins": {
      "cordova-plugin-googleplus": {}
    },
    "platforms": [
      "electron"
    ]
  }
}
```

### Solution

I managed to find the problem in this line:
https://github.com/apache/cordova-lib/blob/master/src/cordova/platform/addHelper.js#L102

If "dependencies" exists, the code skips the devDependencies.

### Version information
cordova-lib@10.0.0

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

aidanas:
Closes #866

"[Wrong platform version is detected when dependencies and devDependencies are both present](https://github.com/apache/cordova-lib/issues/866)"

Relates to:
"[Save platforms and plugins to devDependencies](https://github.com/apache/cordova-fetch/issues/64)"

### Platforms affected
All

### Motivation and Context

The issue was discovered using a project which has both dev and normal dependencies where platform dependency was specified under `devDependencies`.

Steps to recreate:
1. Have a project with both dev and normal dependencies in `package.json`.
2. `cordova platform add electron@2.0.0`
3. `rm -rf ./node_modules ./plugins ./platforms`
4. `cordova prepare electron`
5. Should be using 2.0.0 from `package.json` but will be using 1.0.0 opinned version. 

### Description

This is a small fix to address an issue where Cordova CLI restores platforms using its pinned version rather than version present in `package.json`. Such a case exists if `package.json` has normal and dev dependencies. With the logic currently at `master` it checks if `dependencies` key exists within `package.json` and attempts to find platform version there if not found  it never bothers to check `devDependencies`.

With the fix it will check normal dependencies and if not found will check dev dependencies.

### Testing

Using the same steps as to reproduce and confirming that correct version is being used.

### Checklist

- [x] I added automated test coverage as appropriate for this change (courtesy of @lirichard)

### DIFF
--- a/spec/cordova/platform/addHelper.spec.js
+++ b/spec/cordova/platform/addHelper.spec.js
@@ -247,6 +247,21 @@ describe('cordova/platform/addHelper', function () {
                     });
                 });
 
+                it('should use pkgJson version devDependencies, if dependencies are nonempty but do not include the platform', function () {
+                    package_json_mock.dependencies.lorem = {}; // Add some item to dependencies so it's defined but nonempty
+                    package_json_mock.cordova = { platforms: ['ios'] };
+                    package_json_mock.devDependencies.ios = {};
+                    cordova_util.requireNoCache.and.returnValue(package_json_mock);
+                    fs.existsSync.and.callFake(function (filePath) {
+                        return path.basename(filePath) === 'package.json';
+                    });
+                    fs.readFileSync.and.returnValue('{}');
+                    return platform_addHelper('add', hooks_mock, projectRoot, ['ios'], { save: true, restoring: true }).then(function () {
+                        expect(platform_addHelper.getVersionFromConfigFile).not.toHaveBeenCalled();
+                        expect(fs.writeFileSync).toHaveBeenCalled();
+                    });
+                });
+
                 it('should only write the package.json file if it was modified', function () {
                     package_json_mock.cordova = { platforms: ['ios'] };
                     cordova_util.requireNoCache.and.returnValue(package_json_mock);
--- a/src/cordova/platform/addHelper.js
+++ b/src/cordova/platform/addHelper.js
@@ -99,7 +99,9 @@ function addHelper (cmd, hooksRunner, projectRoot, targets, opts) {
                         } else if (pkgJson.dependencies[platform]) {
                             spec = pkgJson.dependencies[platform];
                         }
-                    } else if (spec === undefined && pkgJson && pkgJson.devDependencies && cmd === 'add') {
+                    }
+
+                    if (spec === undefined && pkgJson && pkgJson.devDependencies && cmd === 'add') {
                         if (pkgJson.devDependencies['cordova-' + platform]) {
                             spec = pkgJson.devDependencies['cordova-' + platform];
                         } else if (pkgJson.devDependencies[platform]) {


erisu:
Can you explain why you have the platform package defined in both `dependencies` and `devDependencies`?

Have it defined in both is not considered as a valid npm package...

For example, if you run the following commands

```shell
mkdir testProject
npm init -y
npm i cordova-android
npm i -D cordova-android
```

You will see that there is only one entry for `cordova-android`. The npm process would just move the package from the `dependencies` list to `devDependencies`.

That is why cordova-lib is using an `if else` statement. It checks for one and if it doesn't exist then it should be in the other.

I think the solution is to delete one of them, and preferably keep it in `devDependencies` as the package is not something bundled with the app.

lirichard:
It's not that the platform package is in both `dependencies` and `devDependencies`—it's that the platform package is in `devDependencies` and some other unrelated package(s) are in `dependencies`.

For example you can replicate by creating this test project:

```
cordova create testProject
cd testProject
npm i lodash@4.17.21
cordova platform add android@10.1.0
rm -r platforms
cordova prepare android
```

The package.json ends up with these `dependencies` and `devDependencies` (I've omitted everything else in the package.json for brevity):

```json
// package.json
{
    "dependencies": {
        "lodash": "^4.17.21"
    },
    "devDependencies": {
        "cordova-android": "^10.1.0",
        "cordova-plugin-whitelist": "^1.3.4"
    }
}
```

However, the last `cordova prepare android` outputs `Using cordova-fetch for cordova-android@^9.0.0` instead of the expected `10.1.0` from `devDependencies`.

Here, when you run `cordova prepare android`, the current code tries to look inside `dependencies` for `cordova-android` ([`pkgJson.dependencies` is truthy](https://github.com/apache/cordova-lib/blob/dd872f044211592b11c17d23d56cd17192e4b51f/src/cordova/platform/addHelper.js#L96)). Naturally, it doesn't find `cordova-android`. It then _skips_ the `devDependencies` ([because of the `else if`](https://github.com/apache/cordova-lib/blob/dd872f044211592b11c17d23d56cd17192e4b51f/src/cordova/platform/addHelper.js#L102)), which is the bug we're addressing here.

The `else if` is a little misleading because the code does a sequential process to find the `spec` value. See:
1. `if (spec === undefined ...)` on [addHelper.js#L96](https://github.com/apache/cordova-lib/blob/dd872f044211592b11c17d23d56cd17192e4b51f/src/cordova/platform/addHelper.js#L96) to look inside `package.json` `dependencies`; then
2. `if (... spec === undefined ...)` on [addHelper.js#L110](https://github.com/apache/cordova-lib/blob/dd872f044211592b11c17d23d56cd17192e4b51f/src/cordova/platform/addHelper.js#L110) to look inside `config.xml`; then finally
3. `if (!spec ...)` on [addHelper.js#L116](https://github.com/apache/cordova-lib/blob/dd872f044211592b11c17d23d56cd17192e4b51f/src/cordova/platform/addHelper.js#L116]) to fall back to the pinned platform version.

erisu:
I sure the issue is actually `installPlatformsFromConfigXML` from the `restore-util.js`. Not the addHelper file.

This method is for:

> // Install platforms looking at config.xml and package.json (if there is one).

When you run `cordova prepare` this is the method that is suppose to find the pacakge info and restore it.

First it it checks config.xml then package.json but it seems to not be filtering the pacakge.json info to find the spec info.

lirichard:
Just following the code visually, I believe this is the order of events:
1. [restore-util.js#L137](https://github.com/apache/cordova-lib/blob/ef1868884164a0838da66bd97503a0a1903e0bcd/src/cordova/restore-util.js#L137) ends up invoking `cordovaPlatform('add', ...)`
2. Which takes us into [`platform/index.js`](https://github.com/apache/cordova-lib/blob/dd872f044211592b11c17d23d56cd17192e4b51f/src/cordova/platform/index.js#L48)
3. Which invokes a local function `add` ([platform/index.js#L70](https://github.com/apache/cordova-lib/blob/dd872f044211592b11c17d23d56cd17192e4b51f/src/cordova/platform/index.js#L70))
4. Which calls `addHelper` ([platform/index.js#L27](https://github.com/apache/cordova-lib/blob/dd872f044211592b11c17d23d56cd17192e4b51f/src/cordova/platform/index.js#L27))

I confirmed that, if I make aidanas's code change within a live Cordova project in the local `node_modules/cordova-lib/src/cordova/platform/addHelper.js`, then `cordova prepare` pulls the correct version `^10.1.0`.

For example using my `testProject` reproduction from before, modified to use `npx` so it uses the local `cordova` and `cordova-lib` instead of global:

```
npx cordova create testProject
cd testProject
npm i lodash@4.17.21
npx cordova platform add android@10.1.0
rm -r platforms

npm i -D cordova

# At this point, before continuing, change "else if" to "if" in ./node_modules/cordova-lib/src/cordova/platform/addHelper.js#L102
# You could do it manually in a text editor - I'm using PowerShell for the sake of reproducing
$addHelperJs = Get-Content ".\node_modules\cordova-lib\src\cordova\platform\addHelper.js" -Raw;
$addHelperJs = $addHelperJs.Replace("else if (spec === undefined && pkgJson && pkgJson.devDependencies && cmd === 'add')", "if (spec === undefined && pkgJson && pkgJson.devDependencies && cmd === 'add')");
$addHelperJs | Set-Content ".\node_modules\cordova-lib\src\cordova\platform\addHelper.js"

npx cordova prepare android
```

```
Discovered platform "android". Adding it to the project
Using cordova-fetch for cordova-android@^10.1.0
Adding android project...
Creating Cordova project for the Android platform:
        Path: platforms\android
        Package: io.cordova.hellocordova
        Name: HelloCordova
        Activity: MainActivity
        Android target: android-30
Subproject Path: CordovaLib
Subproject Path: app
Android project created with cordova-android@10.1.0
Installing "cordova-plugin-whitelist" for android
Plugin doesn't support this project's cordova-android version. cordova-android: 10.1.0, failed version requirement: >=4.0.0 <10.0.0
Skipping 'cordova-plugin-whitelist' for android
```

lirichard:
I created a minimal reproduction repo: https://github.com/lirichard/cordova-platform-devdependencies-reproduction

breautek (approved):
I took an extensive look at this issue and I basically agree with both @erisu and @lirichard here...

There are two issues at play here.

The root issue is what Erisu is bringing attention to. There is a flaw that prevents the restore util from finding the spec.

https://github.com/apache/cordova-lib/blob/dd872f044211592b11c17d23d56cd17192e4b51f/src/cordova/restore-util.js#L99-L112

Here we are comparing the platforms list, which is a list of non-prefixed cordova platforms, and using those values to find the spec from `dependencies` or `devDependencies`, which will have prefixed cordova platforms. Therefore, this will always result in an undefined spec.

This will then pass the `undefined` spec into `addHelper`, which leads into issue 2.

Issue 2 is a flaw in the if / else if condition for determining a spec. Virtually any project will have a truthy `dependencies` entry and thus it will always enter the first `if` condition and never enter the `else if`.

This PR addresses this flaw by separating the if conditions into two independent if conditions, where it checks `devDependencies` last, which makes the behaviour consistent with other parts of the codebase, where `devDependencies` are taken priority over `dependencies`.

This could be simplified by merging the two if conditions into one if condition, similar to:

```javascript
if (spec === undefined && pkgJson && cmd === 'add') {
    let dependencies = Object.assign({}, pkgJson.dependencies || {}, pkgJson.devDependencies || {});
    // spec = check for cordova-platform first then fallback to platform
}
```

But I think this PR is fine as is for now as it is definitely still an improvement compared to what we currently have. I'll create another bug ticket to track first issue that I described earlier.

Also, we cannot guarantee when a release will be made after this becomes merged, but there is a workaround available.

If you have `<engine name="<the-platform>" spec="<desired-spec>" />` in your `config.xml`, the projects will be restored correctly due to the last condition check before it defaults to the default version pins via:

https://github.com/apache/cordova-lib/blob/dd872f044211592b11c17d23d56cd17192e4b51f/src/cordova/platform/addHelper.js#L110-L113

raphinesse (approved):

PR #880, opened 2021-09-11
fix(restore-util): consider long and short names when restoring platforms from package.json

### LINKED ISSUE #878: restore-util doesn't properly find platform spec info
# Bug Report

## Problem

Code:

https://github.com/apache/cordova-lib/blob/dd872f044211592b11c17d23d56cd17192e4b51f/src/cordova/restore-util.js#L99-L112

This should be able to find the spec version

spec will always result in `undefined`

## Information

This is because we are comparing `pkgJson.cordova.platforms` which is an array of non-prefixed cordova platforms, and then we are comparing those values against `dependencies` or `devDependencies` which is always going to be a `cordova-` prefixed platform. This compatibility will always result in an undefined spec.

This issue is described more in depth in [this thread](https://github.com/apache/cordova-lib/pull/874#pullrequestreview-751944535).

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [ x] I included all the necessary information above

breautek:
### Platforms affected

All

### Motivation and Context

Fixes #878 

### Description

Try platform name prefixed with `cordova-` for platform / spec mapping, before falling back to plain platform name.

### Testing

Added a unit test to ensure that platforms are restored without config.xml engine entry, and from devDependency.
All existing unit tests passes.

Manually tested in a [repo project](https://github.com/lirichard/cordova-platform-devdependencies-reproduction).

### Checklist

### DIFF
--- a/src/cordova/restore-util.js
+++ b/src/cordova/restore-util.js
@@ -108,7 +108,7 @@ function installPlatformsFromConfigXML (platforms, opts) {
 
     const platformInfo = platformIDs.map(plID => ({
         name: plID,
-        spec: specs[plID]
+        spec: specs[`cordova-${plID}`]
     }));
 
     let platformName = '';
--- a/src/cordova/restore-util.js
+++ b/src/cordova/restore-util.js
@@ -108,7 +108,7 @@ function installPlatformsFromConfigXML (platforms, opts) {
 
     const platformInfo = platformIDs.map(plID => ({
         name: plID,
-        spec: specs[`cordova-${plID}`]
+        spec: specs[`cordova-${plID}`] || specs[plID]
     }));
 
     let platformName = '';
--- a/spec/cordova/restore-util.spec.js
+++ b/spec/cordova/restore-util.spec.js
@@ -212,6 +212,22 @@ describe('cordova/restore-util', () => {
             });
         });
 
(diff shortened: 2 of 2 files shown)


breautek:
> Not sure if this is a safe assumption though.

Do we support custom platforms...? If so, then this PR would probably break custom platform restorations...

erisu:
I believe we always supported custom platforms, even though I am not sure if any exist.

src/cordova/restore-util.js
```diff
@@ -108,7 +108,7 @@ function installPlatformsFromConfigXML (platforms, opts) {
 
     const platformInfo = platformIDs.map(plID => ({
         name: plID,
-        spec: specs[plID]
+        spec: specs[`cordova-${plID}`]
```
raphinesse, line 111: The following should approximate the logic we use in `addHelper.js` to determine the platform spec:

```suggestion
        spec: specs[`cordova-${plID}`] || specs[plID]
```

raphinesse (approved):

breautek:
Thanks Raphinesse for taking the initiative of adding the required change in this PR!

raphinesse:
My pleasure!

PR #882, opened 2021-09-13
refactor(addHelper): more concise package.json spec lookup

raphinesse:
Improves code and tests related to the code fixed in #874.


erisu (approved):
LGTM

PR #888, opened 2021-11-02
dep(dev)!: bump @cordova/eslint-config@^4.0.0 w/ fix

erisu:

- Use latest Cordova eslint config.
- Applied Fix

Requires https://github.com/apache/cordova-lib/pull/887

### Testing

- `npm t`

### Checklist

### DIFF
--- a/spec/cordova/plugin/remove.spec.js
+++ b/spec/cordova/plugin/remove.spec.js
@@ -17,22 +17,22 @@
     under the License.
 */
 
-var rewire = require('rewire');
-var remove = rewire('../../../src/cordova/plugin/remove');
-var cordova_util = require('../../../src/cordova/util');
-var metadata = require('../../../src/plugman/util/metadata');
-var events = require('cordova-common').events;
-var plugman = require('../../../src/plugman/plugman');
-var fs = require('fs-extra');
-var plugin_util = require('../../../src/cordova/plugin/util');
+const rewire = require('rewire');
+const remove = rewire('../../../src/cordova/plugin/remove');
+const cordova_util = require('../../../src/cordova/util');
+const metadata = require('../../../src/plugman/util/metadata');
+const events = require('cordova-common').events;
+const plugman = require('../../../src/plugman/plugman');
+const fs = require('fs-extra');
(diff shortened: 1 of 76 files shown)


spec/cordova/plugin/remove.spec.js
```diff
@@ -32,8 +32,7 @@ describe('cordova/plugin/remove', function () {
     const cfg_parser_mock = function () {};
     let package_json_mock;
     const plugin_info_provider_mock = function () {};
-    let plugin_info;
-    package_json_mock = jasmine.createSpyObj('package json mock', ['cordova', 'dependencies']);
+    const plugin_info = jasmine.createSpyObj('package json mock', ['cordova', 'dependencies']);
```
raphinesse: Should be `package_json_mock = ...`, right?

spec/plugman/util/metadata.spec.js
```diff
@@ -32,6 +32,7 @@ const fsMock = {
 };
 
 // expect fsMock to only operate on fetchJsonPath
+// eslint-disable-next-line
 Object.entries(fsMock).map(([key, fn]) => {
```
raphinesse, line 36: Would `forEach` work without a warning?
```js
Object.entries(fsMock).forEach(([key, fn]) => {
```

raphinesse (changes requested):
See comments. LGTM otherwise

spec/cordova/plugin/remove.spec.js
erisu: This was already fix

spec/plugman/util/metadata.spec.js
erisu: I just tested with `forEach` and it worked OK.

The warning was about how:
> Array.prototype.map() expects a return value from arrow function

raphinesse (approved):

spec/plugman/util/metadata.spec.js
breautek: Yah, `map` is for creating a new list out of another list... so if the goal was simply iterate then `forEach` is definitely what should have been used.

PR #578, opened 2017-07-18
CB-12361 : added unit-tests for getPlatformDetailsFromDir

audreyso:
### What does this PR do?

added unit-tests for getPlatformDetailsFromDir

### Checklist
- [X] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [X] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [X] Added automated test coverage as appropriate for this change.

PR #913, opened 2023-03-28
[cordova-cli #578] plugin/remove.js: don't stomp opts.cli_variables in removePluginFromPlatform()

sebastian-onlea:
### Platforms affected
CLI - adding and removing plugins with differing required variables per platform (e.g. [cordova-plugin-googleplus](https://github.com/TheAnkurPanchani/cordova-plugin-googleplus) )

### Motivation and Context
resolves apache/cordova-cli#578 :
`cordova plugin rm plugin-package --variable "IOS_ONLY_REQUIRED_VARIABLE=variable-value"` will fail if the project includes e.g. both android and ios platforms.

### Description
In `removePluginFromPlatform()`, at [cordova/plugin/remove.js:105](https://github.com/apache/cordova-lib/blob/7f8b2d011dc05749cf6075fe483b5e1bb08f1a07/src/cordova/plugin/remove.js#L105), `opts.cli_variables` is destructively reassigned to the result of `mergeVariables()`. As mergeVariables() selects only keys specified for a specific platform, this can result in CLI-specified values being lost before the platform that requires them is processed.

In the change, instead call `uninstallPlatform()` with a copy of `opts` that includes the platform-specific `mergeVariables()` result. 

### Testing
- Ran existing tests (`npm run test`)
- verified the change fixes my local use case by `npm link`ing

### Checklist

  > I'm not familiar enough with how to create a reproduction case in the test library for this project.
  > No doc update should be required

### DIFF
--- a/src/cordova/plugin/remove.js
+++ b/src/cordova/plugin/remove.js
@@ -92,7 +92,6 @@ function remove (projectRoot, targets, hooksRunner, opts) {
 
     function removePluginFromPlatform (target, platform) {
         let platformRoot;
-
         return Promise.resolve().then(function () {
             platformRoot = path.join(projectRoot, 'platforms', platform);
             const directory = path.join(pluginPath, target);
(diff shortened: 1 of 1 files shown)


breautek:
Thank you for the PR.

We generally do not make back ports releases. If applicable, the PR should target the `master` branch so that when merged it will be included in our next release. I'm not sure if GitHub allows you to re-target a PR, it may need to be closed and recreated.

dpogue (approved):

PR #917, opened 2023-04-20
feat!(run): call platform api to list targets

erisu:
### Platforms affected

n/a

### Motivation and Context

Prepare for the removal of calling binaries in the platform packages. 
Use the platform API to handle listing out of targets.

### Description

This PR will call `platformApi.listTargets` when the platfrom api contains the method.
To also support platforms that has not been updated, it will fallback and call the old binrary.

### Testing

`npm t`

### Checklist


breautek (approved):
lgtm

PR #918, opened 2023-05-18
fix: platform add with tarball

erisu:
### Platforms affected

n/a

### Motivation and Context

Fix installing platforms with tarball.

`cordova platform add cordova platform add ../cordova-android-12.0.0.tgz` will append `@latest` to the end and causes install failures.

### Description

Only append `@latest` to only platforms that are known.

### Testing

`npm t`

### Checklist

### DIFF
--- a/spec/cordova/platform/addHelper.spec.js
+++ b/spec/cordova/platform/addHelper.spec.js
@@ -167,12 +167,6 @@ describe('cordova/platform/addHelper', function () {
                 });
             });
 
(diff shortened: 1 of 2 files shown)


NiklasMerz (approved):
I did not test it but sounds like a good fix.

dpogue (approved):
If it's easy, would be good to have a unit test for this so that it doesn't accidentally get broken again

dpogue (approved):

breautek (approved):

PR #949, opened 2025-07-19
chore!: bump node requirement & npm dependencies

kumo01GitHub:
### Platforms affected

N/A

### Motivation and Context

node < 20 are already EOL.

### Description

- Upgrade node to 20.17.0
- Bump node modules:
  - cordova-common
  - init-package-json
  - semver
  - write-file-atomic
  - @cordova/eslint-config
  - jasmine
  - nyc
  - rewire
- Update packages by npm update

NOTE:
- Below modules could not be updated because it switched commonjs to esm from next version:
  - detect-indent 
  - detect-newline
  - exec
  - globby
- `nyc` should be replaced with `c8` but separate from this PR.
- Latest version of `init-package-json` requires `node@^20.17.0 || >= 22.9.0` so that I bump node to 20.17.0. It might be better to choose same requirement or `20.5.0`.

### Testing

- npm t

### Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node-version: [16.x, 18.x, 20.x, 22.x, 24.x]
+        node-version: [20.x, 22.x, 24.x]
         os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,11 +18,11 @@
         "elementtree": "^0.1.7",
         "execa": "^5.1.1",
         "globby": "^11.1.0",
-        "init-package-json": "^6.0.0",
+        "init-package-json": "^8.2.1",
         "md5-file": "^5.0.0",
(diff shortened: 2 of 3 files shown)


kumo01GitHub:
I tried to bump Cordova-android and update "getInstalledPlatformsWithVersions method" test case in `util.spec.js` but below test suites finish with error and I couldn't fix immediately.:
1) plugman/uninstall uninstallPlatform success Test 014 : should uninstall dependent plugins
2) plugman/uninstall end Test 013 : end

I assumed that these errors are not problem on bumping node requirement and dependencies. That's why I didn't bump Cordova-android.

erisu (approved):
LGTM

kumo01GitHub:
I think we should stay 5.8.0.:
- Jasmine is released about monthly and Cordova modules are not released as more frequently as Jasmin. Jasmine might be released again before Cordova modules including others are released.
- The main purpose of this PR is bumping node because of EOL. Jasmine bumped node in 5.9.0 but still supports node@18.20.5+. If they drop 18.x, I think it's worth to update Jasmine. So that for now, I think better to prioritize using same version in other repos of Cordova.

PR #953, opened 2025-07-22
feat!: remove support for create package json

erisu:
### Affects

`cordova-plugman`

Remove legacy Plugman CLI-specific plugin creation logic.

* This logic is specific to **Plugman CLI**, which hasn't been updated or released since 2019.
* The functionality was not used by any other Apache Cordova tools.
* There appears to be a lack of documentation or support around this feature. 
  * The create plugin guide also removed references of `plugman` since.

Given the lack of documentation and that Plugman CLI is become outdate and buggy, I would be more inclined to **fully remove** this feature and not migrate it to `cordova-plugman`.

But as for this repo, I think this logic should not exist here.

As for `cordova-plugman`, others even suggest creating a dedicated plugin template with clear documentation would be more beneficial than the current process.

Third-party tools maybe also provide a more modern flow for creating plugins.

### Testing

- `npm t`

### Checklist


erisu:
> is this repo need to be separated from Cordova cli?

No, Cordova-CLI still needs `cordova-lib` to function.

Removing this feature does not effect to Cordova-CLI. Cordova-CLI has never used this specific plugman feature.

kumo01GitHub:
Or it might be my misunderstanding if you will remove function to create plugin from Plugman CLI and `cordova-lib` still supports Plugman CLI.

dpogue:
There are other non-ASF projects that depend on `cordova-lib` to implement their own CLI functionality, so IMO it makes sense for `cordova-lib` to still exist and for `cordova-cli` to be a wrapper over top of it

kumo01GitHub:
I understood. This repo will be remained for third-party modules.

PR #986, opened 2026-04-25
fix: handle prerelease platform versions in engine requirement checks

GitToTheHub:
### Platforms affected

All

### Motivation and Context

- This PR fixes plugin engine requirement checks for prerelease platform versions such as 8.0.2-dev.0.
- Added tests
- Generated-By: GPT-5.3-Codex

### Description

When adding a plugin, Cordova validates engine constraints in plugin metadata. For platform constraints, prerelease versions are compared directly with semver.satisfies(). Since semver excludes prerelease versions from normal ranges by default, versions like 8.0.2-dev.0 may not satisfy >=7.0.0, even though they should be treated as the corresponding release line.

This can trigger warnings such as:

```bash
Unmet project requirements for latest version of cordova-plugin-google-maps-sdk:
    cordova-ios (8.0.2-dev.0 in project, >=7.0.0 required)
Current project does not satisfy the engine requirements specified by any version of cordova-plugin-google-maps-sdk. Fetching latest version of plugin anyway (may be incompatible)
```

### Root cause
The cordova requirement path already normalizes prerelease cordova versions before comparison, but platform versions were not normalized the same way.

### Fix
Normalize prerelease platform versions before semver engine checks, mirroring existing behavior for the cordova version check.

### Checklist

### DIFF
--- a/spec/cordova/plugin/add.spec.js
+++ b/spec/cordova/plugin/add.spec.js
@@ -449,6 +449,12 @@ describe('cordova/plugin/add', function () {
             it('should return an array with failed platform requirements ', function () {
                 expect(add.getFailedRequirements({ 'cordova-android': '>=6.0.0' }, {}, { android: '5.5.0' }, '7.0.0')).toEqual([{ dependency: 'cordova-android', installed: '5.5.0', required: '>=6.0.0' }]);
             });
+            it('should treat prerelease platform versions as the matching release version', function () {
+                expect(add.getFailedRequirements({ 'cordova-ios': '>=7.0.0' }, {}, { ios: '8.0.2-dev.0' }, '12.0.0').length).toBe(0);
+            });
+            it('should return failed platform requirements for prerelease versions below minimum', function () {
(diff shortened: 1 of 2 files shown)


dpogue:
I thought this issue had already been fixed by https://github.com/apache/cordova-lib/pull/935 but it looks like this is a slightly different spot.

Rather than modifying the version with `inc`, we probably want to just pass the same flags to `satisfies` that we did in #935 so our version handling is consistent.

GitToTheHub:
Good point I changed it to just use `satisfies` with the options `{ loose: true, includePrerelease: true }`

dpogue (approved):

GitToTheHub:
I see there are two other conditions above without using `{ loose: true, includePrerelease: true }`:

```js
            if (pluginMap[trimmedReq] && !semver.satisfies(pluginMap[trimmedReq], reqs[req])) {
                badInstalledVersion = pluginMap[req];
            } else if (trimmedReq === 'cordova' && !semver.satisfies(version, reqs[req])) {
                badInstalledVersion = cordovaVersion;
```

Should the options `{ loose: true, includePrerelease: true }` added there too?

PR #989, opened 2026-06-17
chore(deps): Update testing/coverage deps

dpogue:
### Platforms affected
All

### Description

Resolves some npm audit warnings by updating to the latest versions of jasmine and c8. Overrides c8's yargs dependency version to make it work on Node 26.

### Testing

Tests pass locally

### Checklist


package.json
```diff
@@ -23,13 +23,18 @@
   },
   "devDependencies": {
     "@cordova/eslint-config": "^6.0.0",
-    "c8": "^10.1.3",
+    "c8": "^11.0.0",
     "cordova-android": "^13.0.0",
     "elementtree": "^0.1.7",
-    "jasmine": "^5.10.0",
+    "jasmine": "^6.3.0",
     "jasmine-spec-reporter": "^7.0.0",
```
erisu, line 30: I believe this dependency was only used in the `helper.js` file that was removed in this PR, so it might be OK to remove it. Needs confirmation.

dpogue: I might need to restore the helper stuff, because Windows tests were timing out in my fork

erisu (approved):
LGTM

PR #933, opened 2024-10-17
fix(ios): Prevent mix build phases

MaximBelov:
### Platforms affected
Any

### Motivation and Context

Different installation orders for currently installed plugins

iOS
https://github.com/MaximBelov/reproduce-cordova-apply-plugins-sort/blob/20ab95bca8aef992aa5b1982b38e5e6d750580a8/plugins/ios.json

Android
https://github.com/MaximBelov/reproduce-cordova-apply-plugins-sort/blob/20ab95bca8aef992aa5b1982b38e5e6d750580a8/plugins/android.json

Plugins installation order should be the same for any platform and have the source of faithful from package.json

Why is this important?
ios: Prevent mix build phases

The plugin `cordova-plugin-firebasex` runs a hook that adds a new Build Phase "Crashlytics", which should be the last one
https://github.com/dpa99c/cordova-plugin-firebasex/pull/897

### Description

How to reproduce

https://github.com/MaximBelov/reproduce-cordova-apply-plugins-sort

### Testing

How to reproduce 
Repository: https://github.com/MaximBelov/reproduce-cordova-apply-plugins-sort

**Version without patch**

```sh
git checkout master
npm run build
git diff  --no-index plugins/ios.json plugins/android.json

```
Result:  
image
__________

**Version with patch**

```sh

git checkout patch
npm run build
git diff  --no-index plugins/ios.json plugins/android.json

```
Result:
image

### Checklist

- [x] I've run the tests to see all new and existing tests pass
- [ ] I added automated test coverage as appropriate for this change
- [x] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [ ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [ ] I've updated the documentation if necessary

PR #1011, opened 2026-09-10
fix: run plugin hooks in installation order

### LINKED ISSUE #1010: Plugin hook execution order is unspecified (fast-glob readdir order), while install order is deterministic since #933
### Bug Report

**Problem**

Plugin hook execution order is unspecified, so a plugin whose hook depends on another plugin's hook having already run works or fails depending on the filesystem. Nothing in the project surfaces the order, and the failure is silent: the hook reports success and its work is discarded a moment later by the other plugin.

`getAllPluginsHookScriptFiles` in `src/hooks/scriptsFinder.js` enumerates plugins like this:

```js
const plugins = (new PluginInfoProvider()).getAllWithinSearchPath(path.join(opts.projectRoot, 'plugins'));
plugins.forEach(function (pluginInfo) { /* collect that plugin's hook scripts */ });
```

and `getAllWithinSearchPath` (cordova-common, `PluginInfo/PluginInfoProvider.js`) resolves to:

```js
const pluginXmlPaths = fastGlob.sync('{,@*/}*/plugin.xml', { fs, cwd: absPath, onlyFiles: true, absolute: true })
```

`fast-glob` does not sort, so this is `readdir` order for `plugins/`.

**What that looks like in practice**

Measured on a project with 57 installed plugins, macOS/APFS, cordova 13.0.0:

- the glob order is **not** the sorted order overall
- but the five plugins that mattered came back sorted anyway: `cordova-plugin-firebasex-analytics`, `...-core`, `...-crashlytics`, `...-messaging`, `...-performance`

That combination is the awkward part. `cordova-plugin-firebasex-core`'s `after_prepare` **copies** `GoogleService-Info.plist` from the project root over the platform copy, and its sibling modules write configuration keys into that same file from their own `after_prepare` hooks. On this machine `...-analytics` ran before `...-core`, so all six keys it wrote were reported as written and the file ended up with none of them, while `...-crashlytics` and `...-performance` — landing after `core` — kept theirs. On a filesystem that returns those directories in a different order the same project would behave differently, and the plugin author's `after_prepare` would appear to work.

So the plugin bug is the plugin's to fix (it moved to `before_compile`), but the reason it was a coin flip rather than a decision is here.

**Why this looks like an oversight rather than a design choice**

Installation order was made deterministic in #933, from `package.json` as the source of truth:

```js
// If package.json includes the plugins, we use that for proper sort order
const pkgJson = readPackageJsonIfExists(projectRoot);
if (pkgJson?.cordova?.plugins) {
    const pkgPluginIDs = Object.keys(pkgJson.cordova.plugins);
    plugins = plugins.sort((a, b) => pkgPluginIDs.indexOf(a) - pkgPluginIDs.indexOf(b));
}
```

Hook enumeration was not brought along, so the two now disagree: plugins are installed in a defined order and their hooks run in whatever order the filesystem hands back.

**What I would expect**

Either of these would remove the coin flip, and the first keeps the two paths consistent:

1. `getAllPluginsHookScriptFiles` orders plugins the same way `platform/addHelper.js` does — `package.json`'s `cordova.plugins` first, falling back to a plain sort when it is absent.
2. Failing that, document the order as undefined, so plugin authors know that a hook which depends on another plugin's hook needs a phase that cannot race (`before_compile` after `after_prepare`, for instance) rather than an ordering assumption.

Happy to open a PR for option 1 if that is the direction you would take.

**Environment**

- cordova 13.0.0, cordova-lib 13.0.0, cordova-common as bundled
- macOS, APFS
- reproduced with real plugins rather than a fixture; the ordering itself is visible from any project with two plugins whose hooks touch one file

MaximBelov:
### Platforms affected

Any.

### Motivation and Context

Closes #1010.

Plugin hook order is unspecified. `getAllPluginsHookScriptFiles` collects hooks by walking `PluginInfoProvider.getAllWithinSearchPath`, which globs the plugins directory with `fast-glob` and does not sort, so the order is whatever the filesystem returned for `plugins/`.

That is not merely untidy. Where two plugins' hooks touch the same file it decides which one wins, and it can decide differently on another machine, so a plugin can be correct on the author's laptop and wrong in CI with nothing reporting it.

The case that surfaced it, in a real project: one plugin copies a configuration file from the project root into the platform from its own `after_prepare`, and sibling plugins write keys into that same file from theirs. On the machine it was found on, one sibling ran before the copy and had all six keys it wrote discarded a moment later; two others ran after and kept theirs. All three hooks reported success. Measured there: the overall glob order for 57 installed plugins was **not** the sorted order, while the five directories that mattered happened to come back sorted — which is precisely why the behaviour looked like a rule and was not one.

Installation order was already made deterministic from `package.json` in #933. Hook enumeration was not brought along, so today the two disagree: plugins are installed in a defined order and their hooks run in an undefined one.

### Description

`getAllPluginsHookScriptFiles` now orders plugins by their position in `package.json`'s `cordova.plugins`, using the same comparison as `cordova/platform/addHelper.js`, so hooks run in installation order.

Deliberately conservative:

- no `package.json`, or no `cordova.plugins` in it — behaviour is unchanged, the search-path order is kept
- a plugin missing from `cordova.plugins` sorts before the listed ones, exactly as it does during installation (`indexOf` returning `-1`, same as `addHelper`)
- `Array.prototype.sort` is stable, so plugins that compare equal keep the order the search path yielded
- the input array is copied rather than sorted in place, since `getAllWithinSearchPath` caches and returns its arrays

I did not extract `addHelper`'s local `readPackageJsonIfExists` into a shared helper, to keep the diff to the behaviour under discussion. Happy to do that in this PR if you would prefer the two to share one function.

### Testing

`src/hooks/scriptsFinder.js` had no spec, so this adds `spec/hooks/scriptsFinder.spec.js` with seven cases: ordering from `package.json`; ordering that is deliberately not alphabetical, so a sort-by-name implementation would fail it; both fallbacks; a plugin absent from `package.json`; that the given array is not mutated; and that the collected scripts come out in that order rather than only the plugin list.

- `npm run lint` — clean
- `npm run test:unit` — 327 of 328 specs pass, one pre-existing `xit` pending, no failures

### Checklist

- [ ] Commit is prefixed with `(platform)` if this change only applies to one platform — n/a, this is platform-independent
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct keyword to close issues using keywords)
- [ ] I've updated the documentation if necessary — n/a, no documented behaviour changes; the order was previously undocumented and is now the same as installation order


erisu:
@MaximBelov 

I apologies if I'm wrong, but was any AI tooling used to help generate or assistant in part of the commit, in this PR?

If any tooling was used, would it be possible to update the commit messages to include the `Generated-By:` tag as per the [ASF Generative Tooling Guidance](https://www.apache.org/legal/generative-tooling.html).

For example, if "Claude Opus 5" was used, the commit message would contain: `Generated-by: Claude Opus 5`

MaximBelov:
No need to apologise, you're right. Claude Opus 5 was used.

I've amended the commit to carry `Generated-by: Claude Opus 5` (no code change, message only). Thanks for the pointer to the guidance, I hadn't come across the tag before, and I'll add it to my other open PRs too.

